PR #158, opened 2014-05-04
Zefram: Better mesecon-enabled doors

Jeija:
This patch was sent in by Zefram via eMail. I think the concept he describes
sounds pretty genius, it is a big change though and breaks compatibility.
Also, I suggest we need some kind of different textures for these doors;
maybe some, that also allow distinguishing between signalling and operated doors.
I'd like to hear your opinions on his idea and what you think we should do with the
compatibility issue. Would you say that the current doors should be the operated
doors and only signalling doors should be added?
This is Zeframs description of the change:

The mesecons_compatibility doors erred in making steel doors, which
are meant to be locked, openable by anyone using a mesecon signal.
They also didn't handle mirror-paired doors, and nastily duplicated
lots of the standard door code rather than using it and adding to it.
Replace mesecons_compatibility with a new system, in which standard
doors are left alone and new types of door are added that have mesecon
behaviour.  The new door types are each available in both wood and steel,
using the standard door textures.

The mesecon-operated doors open and close according to the mesecon
signal they receive: open when the signal is on and closed when off.
Unlike the old mesecons_compatibility doors, which only accepted the
signal to the bottom half, these accept the signal to either half of
the door.  A convenient kind of control therefore is a wall-mounted
button just above the doorway: the signal flows diagonally down to the
top half of the door.  The door cannot be operated manually.

The mesecon-signalling doors are opened and closed manually, and generate
a mesecon signal indicating whether they're open, on when open and off
when closed.  Thus opening the door can trigger automatic activity.
Pairing a mesecon-signalling door with a mesecon-operated door results
in a door pair where right-clicking on one door operates both.

By making use of the pairing behaviour built into the standard doors mod,
which is inherited by the mesecon doors, and placing doors from sideways
angles, it is possible to effectively get mesecon doors with the opposite
signal sense.  For example, a mesecon-signalling door that sends an on
signal when closed, turning the signal off when opened.

(Nevertheless, this won't be merged until a new stable mintetest version
that contains the override_item feature is released)


Jeija:
Closing this, since the code can no longer be merged and because the double-door feature can be accomplsihed by placing vertical wires two blocks below the doors. Some of the features here would be nice though, but require a complete re-implementation.

PR #200, opened 2015-01-11
Improve object detector

ShadowNinja:
* Add per-node radius.
  * Send player names over digilines.  Player names are send on every tick when the detector is on, and once when the detector shuts off.  This could be optimized to only send when the player list changes.

### DIFF
--- a/mesecons_detector/init.lua
+++ b/mesecons_detector/init.lua
@@ -2,7 +2,6 @@ local GET_COMMAND = "GET"
 
 -- Object detector
 -- Detects players in a certain radius
--- The radius can be specified in mesecons/settings.lua
 
 local object_detector_make_formspec = function (pos)
 	local meta = minetest.get_meta(pos)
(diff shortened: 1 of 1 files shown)


Jeija:
Looks good, but haven't really got to testing it yet.
Just one little thing: Instead of hardcoding the maximum radius of 32 nodes, could you use
```Lua
mesecon.setting("object_detector_max_radius", 32)
```

That way, people that want a radius of more than 32 nodes can change the setting
`mesecon.object_detector_max_radius` in their minetest.conf

ShadowNinja:
@Jeija Good idea, will do.

I'd really prefer to do this by only sending a signal when a player enters or leaves the area, with only information on the player that moved, it seems cleaner and easier for LuaCs to use.  That would kind of be a hack over the current detection system, but maybe it's worth it.  The primary use case I had for this was to detect if exactly one player was around (the victor of a match) so some extra state information would be helpfull too. I could actually do both, since the value doesn't have to be especially small (it should only be copied once, just before running the code).

Any comments on what you think would be the best API?

Jeija:
I'm not really in favor of the idea of only sending a signal when a player enters or leaves an area. As long as minetest itself doesn't provide that kind of API itself, an Lua code would be quite hacky. You could also only check for players in certain intervals, to not add too much server load. That way, players could move into and out of areas without the detector ever noticing them. Also, I think there are more uses cases for the way the player detector currently behaves with your commits applied. For some of the more rare cases were you would need to know when players enter / leave areas, it wouldn't be too hard to keep track of that youself in a LuaController. That also allows to set the scanning frequency yourself in code. Additionally, while your proposal certainly seems clearer, afaik that isn't the way typical real-world low-level bus systems (which digilines tries to imitate) work like (just think of I²C, SPI where you constantly have to poll for sensor data / events. It just kind of violates the whole master-slave concept.
If your proposal is the more popular opinion amongst other contributors / digilines users as well though, I will sure accept it.

Jeija:
Closing due to inactivity and because detector code has been changed, so this would need to be rewritten anyways.

PR #201, opened 2015-01-12
Make movestones behave

est31:
Movestones react very badly to lag and do not at all respect the mvps stopper methods. This fixes both the lag issue and the stopper problem.

### DIFF
--- a/mesecons_movestones/init.lua
+++ b/mesecons_movestones/init.lua
@@ -116,18 +116,26 @@ function mesecon.register_movestone_entity(is_sticky)
 			if not direction then -- no mesecon power
 				--push only solid nodes
 				local name = minetest.get_node(pos).name
+				local push_success = true
 				if  name ~= "air" and name ~= "ignore"
 				and ((not minetest.registered_nodes[name])
 				or minetest.registered_nodes[name].liquidtype == "none") then
-					mesecon.mvps_push(pos, self.lastdir, maxpush)
-					if is_sticky then
+					push_success = mesecon.mvps_push(pos, self.lastdir, maxpush)
+					if is_sticky and push_success then
 						mesecon.mvps_pull_all(pos, self.lastdir)
 					end
 				end
 				local nn = {name=node_name}
-				minetest.add_node(pos, nn)
-				self.object:remove()
-				mesecon.on_placenode(pos, nn)
+				if push_success then
+					minetest.add_node(pos, nn)
+					self.object:remove()
+					mesecon.on_placenode(pos, nn)
+				else
+					local prevpos = mesecon.addPosRule(pos, vector.multiply(self.lastdir, -1))
+					minetest.add_node(prevpos, nn)
+					self.object:remove()
+					-- no call to on_placenode, as we don't want to create an endless loop.
+				end
 				return
 			end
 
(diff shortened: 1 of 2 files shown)


HybridDog:
some time ago l used a movestone for farming and it pushed nodes without limit and without nodeupdate

est31:
bump.

MT-Modder:
Shouldn't movestones move in one direction and be based off of facedir? Right now they move sorta randomly and they definitely don't respect facedir.

est31:
Movestones move based on mesecons placed next to them. It has nothing to do with facedir.

MT-Modder:
Thank you very much. I never really used them before. I started using them to make a pull request.

Jeija:
Thank you very much for your contribution, @est31, and sorry for answering that late. I don't have that much time to spend on mesecons these days, so I rather tend to process the small and simple changes first.

The code looks all right in general, here are some things that I'd prefer to have changed before merging this though:
* Instead of having a function `mesecon.get_movestone_entity(is_sticky)` make a function `mesecon.register_movestone_entity(name, is_sticky)` or something similar that doesn't only return the definition of an entity, but also registers it directly
* You should either explain the functions `addmaxone` and `vecaddmaxone` and why they are required as comments in the code and/or find a way to use something else to achieve what you want. They look a little weird to me, maybe you could try to achieve the same thing with the ternary operator and/or builtin math functions?

est31:
Made the `addmaxone` more explanatory. Yes, it should be called `addlessthanonepointfive`, but thats too long :). Therefore I named the method after its central `if diff => 1 then diff =1 end [...] return ret + diff` part. Also, when you round the result, it is a "real" maximum move by 1. `addmaxonewhenyouroundtheresult` was also too long for me.
Also changed it to use `mesecon.register_movestone_entity`.

Jeija:
While testing your pull request, I found several issues with movestones, e.g. when you let two movestones push stacks in opposite directions, when a sticky movestone advances to the end of the mesecon power line etc. These problems are not caused by your code, est31, but by the concept of movestones as entities itself.
Movestones have been in mesecons since the very first release, even before pistons were present. I wanted to completely remove them since quite some time, but it didn't seem right to remove these features, even if not many people used them.
I think the best compromise between removing them entirely and keeping them in the way they are is to change them so that they become nodes instead of entities. They won't look that good anymore, but they'll become a lot more useful, since they are way more precise as nodes. For example, movestones could now be used to move binary-encoded data in the form of blocks over reading devices, they can now be precisely moved, block per block.

[You can vote](http://strawpoll.me/3636390) on what to do with movestones:
* Make movestones to nodes
* Keep movestones as entities
* Remove movestones altogether

C1ffisme:
Make movestones nodes! I use movestones!

est31:
It seems that movestones need more fixing. But does the move to nodes make it easier? I'm not sure. You could make them check everything every second or so, but then you just slowed the problem down, not solve it. If it is really easier to fix them when they are nodes, then make them nodes, but otherwise I'd keep them as entities.

Jeija:
The poll showed support for the idea of making movestones nodes, I therefore merged my movestones_nodes branch.

PR #202, opened 2015-01-12
Silicon production from normal and desert sand

dora71:
Ability to produce silicon either from "normal" sand or from desert sand.


Jeija:
Makes sense to me. Thanks!

PR #207, opened 2015-01-20
fix texture names for doors

MT-Modder:
This fixes error messages caused by texture name changes recently.


MT-Modder:
I don't know if this will work because I deleted the repository and I'm new to pull requests.

Issue #204, opened 2015-01-18
door textures not found


MT-Modder:
Due to this commit: https://github.com/minetest/minetest_game/commit/61579827875f42a08797229895a2fc2ad8bc3355

HybridDog:
Why does mesecons need the door textures?

MT-Modder:
Compatibility for some reason :).
Look at the mesecons_compatibility mod for more info.

HybridDog:
https://github.com/Jeija/minetest-mod-mesecons/blob/master/mesecons_compatibility/init.lua#L174
Why isn't minetest.registered_nodes["doors:door_wood"] used?

MT-Modder:
Good point! I think you would still have to redefine the on_rightclick function. And you would have to use minetest.registered_nodes["doors:door_wood_b_1"] and minetest.registered_nodes["doors:door_wood_b_2"].

HybridDog:
l think the doors mod should store input of doors.register_door(name, input) into a table called doors.registered_doors with name as index.

MT-Modder:
Yeah that would be the easiest solution. I know pilzadam is going to write the doors mod so I'll suggest it to him.

MT-Modder:
Unfortunately after two hours of testing, the way it's implemented now is the best way :frowning: If you can get it to work go for it but I've tried everything I can. The only way to fix the current issue is to add a "s" to the textures.

MT-Modder:
I submitted a pull request for this issue. Here: https://github.com/Jeija/minetest-mod-mesecons/pull/207

PR #210, opened 2015-01-20
Fix doors texture names

MT-Modder:
Fix doors texture names due to recent changes to minetest_game.
Fixes https://github.com/Jeija/minetest-mod-mesecons/issues/204

Sorry for all the crazyness :) with the pull requests.

### DIFF
--- a/mesecons_compatibility/init.lua
+++ b/mesecons_compatibility/init.lua
@@ -29,9 +29,9 @@ end
 
 function doors:register_door(name, def)
 	def.groups.not_in_creative_inventory = 1
-	
+
 	local box = {{-0.5, -0.5, -0.5,   0.5, 0.5, -0.5+1.5/16}}
-	
+
 	if not def.node_box_bottom then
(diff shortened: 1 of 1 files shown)


Jeija:
Looks good to me! Thanks.
I just don't like it when mintest_game changes things for no real good reason and without maintaining compatibility. That way, people have to run the latest minetest all the time to have mesecons doors work for them...
Some transition phase might have been useful at least.

HybridDog:
There's a way to support both versions: You could use minetest.registered_nodes.

MT-Modder:
@HybridDog minetest.override_item would be better, but I haven't tried it. I tried your method but couldn't get it to work. Pos was nil for the on_rightclick function.

HybridDog:
You only need to get the "tiles" of the original doors with minetest.registered_nodes to support both versions.

MT-Modder:
Wow I feel silly :) I was looking at it the exact opposite way.

PR #215, opened 2015-02-03
Added more sounds to noteblocks

C1ffisme:
I added another octave to noteblocks, and made some sounds for pranking your friends. They'll run in terror when they hear a fire or tnt sound near their house!


Jeija:
Could you please document your changes on the http://mesecons.net website by changing this file over here: https://github.com/Jeija/mesecons.net/blob/master/items/noteblock.php ?

C1ffisme:
I'll see what I can do...

C1ffisme:
Should I make a pull request to the github repo?

Jeija:
Yes, that would be great.

C1ffisme:
Right, here I go.

C1ffisme:
[Done.](https://github.com/Jeija/mesecons.net/pull/7)

Jeija:
Hmm, okay, I just realized one thing that I forgot to check before:
In order to make the TNT sound, you would need to have TNT below the noteblock. If, however, mesecons adds support for activating TNT with mesecon signals in the future, that would set off all the noteblocks of people using this feature...
Maybe you could use some other node for this? Like a coal block or something like that?

C1ffisme:
This isn't always true. If you build something where the mesecon is __Above__ the noteblock, then it is safe. But since this may cause frustration in the future, I will set it to gunpowder, since gunpowder is not activated by mesecons.

VanessaE:
I would go with Jeija's suggestion here, as gunpowder is just a flat item while coal block is..well..a block.  Besides, some games might choose in the future, to trigger gunpowder with a mesecons signal also.

C1ffisme:
Right, okay, changing.

C1ffisme:
Is this pull request still active? It's had the explosion sound changed to a coal block now. Do you want to merge?

Jeija:
I just merged it manually.

Issue #203, opened 2015-01-14
Normal Chest and Protection


tenplus1:
Not so much an issue as a request to add default chest, and protection redo blocks to the end of the mesecons/mesecons_mvps/init.lua file :)

mesecon:register_mvps_stopper("default:chest")
mesecon:register_mvps_stopper("protector:protect")
mesecon:register_mvps_stopper("protector:protect2")

cheapie:
I can understand the protector stuff, but why do you want the chest to stop it?

MT-Modder:
@cheapie Metadata probably is the issue here.

cheapie:
I've never had an issue moving full chests with pistons. In fact, as a server admin, I've occasionally had to use pistons to move nodes owned by other people that contained items.

MT-Modder:
You're right. Maybe it also has to do with protection. Bookshelves should be added too if so.

Jeija:
Merged @MT-Modder's changes in #216
I know there are advantages and disadvantages to this, so if lots of people prefer reverting to the behaviour before the merge, I'll be fine with it.

HybridDog:
l liked the possibility to move chests.

cheapie:
Hmm... maybe pistons/movestones should be able to move this stuff, but have ownership and obey protection?

PR #216, opened 2015-02-05
Register chests and bookshelves as stoppers

MT-Modder:
Solves part of #203. I'm not sure if Jeija wants to add support for protection mod so I left it out.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -232,4 +232,6 @@ function mesecon.mvps_move_objects(pos, dir, nodestack)
 end
 
 mesecon.register_mvps_stopper("default:chest_locked")
+mesecon.register_mvps_stopper("default:chest")
+mesecon.register_mvps_stopper("default:bookshelf")
 mesecon.register_mvps_stopper("default:furnace")


Jeija:
This does kind of make sense, you don't want all your stuff pushed around or getting lost. Unfortunately, it also disables some nice stuff like "delivery wagons" using movestones and chests. I will merge this for now, but possibly revert it if people start to complain.

You're right about me not liking to add support for the protection mod. Mesecons does expose the interfaces to the protection mod to add support for the features described by @tenplus1 in #203. One of mesecons strenghts is, that is does not depend on any other mod or have any relations to mods apart from the default minetest_game. And that is something I want to keep this way, considering mesecons was one of the very early mods and survived the creation and death of many other mods that depended on mesecons or that mesecons could have depended on.

MT-Modder:
Can't a protection mod just register these things when it depends on mesecons?

Jeija:
Yes, it could. It doesn't even need to depend on mesecons, it just has to check if the `mesecon` global is defined and perform actions accordingly. That is what I meant by

> Mesecons does expose the interfaces to the protection mod to add support for the features described by @tenplus1 in #203.

MT-Modder:
:) lol

HybridDog:
>[…] if people start to complain.

l want to complain about an old piston problem. Some months ago l dug an enabled sticky piston and got a usual piston. ls this problem solved yet?

Jeija:
No, @HybridDog, it wasn't until you reported it. These kinds of bugs are the things we developers don't usually spot. Maybe it is just me, but my primary attention goes to the functionality of the nodes, not their drops etc. Thanks for reporting the issue! It is very important for me to get this kind of feedback, because these hidden bugs don't discover themselves.
My latest commit should have fixed the issue.

HybridDog:
>These kinds of bugs are the things we developers don't usually spot.

l know. https://github.com/Zeg9/minetest-teleporters/commit/7953ad7325884c72f644354941312727ea5fe151#commitcomment-4452799

>My latest commit should have fixed the issue.

thanks

VanessaE:
Although this has since been merged, I think it would be wise to revisit it, on the grounds that pushing a "full" chest/bookshelf/etc around on the ground really isn't harmful. Now, there is the question of pushing an item into an area that the placer of the item doesn't own (e.g. A's chest ends up in B's area), but that's something that can be tested for and explicitly blocked if it isn't already.

(The same case be made for minetest_game's screwdriver, since in neither case are you removing the node from the map)

HybridDog:
l think the piston and movestone should have their placer in their meta. So protection support could be possible.

cheapie:
This commit broke my auto-packer machine...

MT-Modder:
I see two options here: either add a setting to enable/disable these stoppers or totally remove all stoppers. I'll submit a pull when jeija decides what to do.

cheapie:
I'm personally of the opinion that only "owned" items should be stoppers - locked chests, protection blocks, steel doors, etc.

MT-Modder:
Hmmm...good point. I think I'll submit a pull request with your suggestion.

cheapie:
OK, sounds good.

MT-Modder:
Oops too many references! Sry for the spam.

HybridDog:
>Oops too many references! Sry for the spam.

github did it, not you

est31:
The main reason behind this was that you could mesecons to move objects like furnaces out of protected areas. Inside the area, only area owners could take out items, outside everybody. Also, protection blocks for TenPlus1's protection mod could be moved. So there is no need for locked chests and doors to be moved, but furnaces and protection blocks should be stoppers.

MT-Modder:
@est31  I'm gonna work on adding protection support to solve this issue.

PR #217, opened 2015-02-05
Add more wires to the craftable group

MT-Modder:


MT-Modder:
Bump

Jeija:
Personally, I'm not in favor of this. It doesn't make sense to me that you could craft the exact same thing using different recipes. That way the same craft ouput could consume different amounts of resources.

MT-Modder:
Then insulated crossover and t-junction should be removed from this group. Should I submit a pull request for it?

Jeija:
@MT-Modder: That would be great, I'd merge such a pull request.

MT-Modder:
Cool I'll get right on it.

MT-Modder:
I decided to just remove the wires from the group so mod creators can use the group if they want to.
Here's the pull: #221

PR #219, opened 2015-02-12
fix support of the old version of the doors mod

HybridDog:
l'm using an older minetest_game and so l don't have trap-, obisidan and glass doors.


MT-Modder:
Definitely needed.

Jeija:
@HybridDog, your code would not have worked since
```Lua
if not minetest.registered_nodes[name] then
    return
end
```
always returns, even if the door was registered, since the door basename is only registered as an item. Therefore I made my own commit.

HybridDog:
ok

PR #222, opened 2015-02-14
Register only locked nodes as stoppers

MT-Modder:
This was suggested here:
https://github.com/Jeija/minetest-mod-mesecons/pull/216


Jeija:
Since multiple people had issues with the current way of handling chests etc. as mvps_stoppers, I think it is the correct decision to rollback that change. Protecting the steel door makes sense to me.

MT-Modder:
Thanks.

PR #227, opened 2015-03-25
Override the /mods command that it doesn't display mesecons_* mods to reduce spam

HybridDog:


Jeija:
I don't like this change, people should be able to see wheter a specific component of mesecons is installed or not, e.g. some people might want to remove movestones from their server. Also, this adds unnecessary bloat to mesecons for maintenance. Whenever the default /mods command changes, mesecons would have to update. And obviously not every mod can do that.
A real fix would be to only display modpacks in the upstream /mods command, a fix just for mesecons doesn't really make sense. Thanks for your effort, but I'm not going to merge this.

HybridDog:
ok, l also noticed that the mesecon.setting can't work right for bools
as far as l know
minetest.setting_getbool("mesecon."..setting) or default
is always true if default is true

PR #229, opened 2015-03-26
Allowed loops in LuaController, fixed certain strings

gamemanj:
Currently, the LuaController doesn't allow using loops, despite there being code to handle a infinite loop.
This is because LuaJIT doesn't call debug hooks on empty loops.
So, this commit adds code into loops, to ensure the debug hooks are called.
Furthermore, this fixes the bug where strings containing " while " or " do " or similar would cause a error, by adding a lexer to identify strings and comments, then ignore them.
Finally, it fixes a potential server-crasher involving pcall and xpcall.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -385,14 +385,13 @@ local function lexLua(code)
         local inEscape=true
         local rstr=""
         for i=1,str:len() do
-            local c=str:sub(1,1)
+            local c=str:sub(i,i)
+            rstr=rstr..c
             if inEscape then
-                rstr=rstr..c
                 inEscape=false
             else
-                if c==quoteType then return rstr..quoteType end
+                if c==quoteType then return rstr end
                 if c=="\\" then inEscape=true end
-                rstr=rstr..c
             end
         end
         return nil --unfinished string
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -1,8 +1,8 @@
---		______
---	   |
---	   |
---	   |		__	   ___  _   __		 _  _
--- |   | |	   |  | |\ |  |  |_| |  | |  |  |_ |_|
+--        ______
+--       |
+--       |
+--       |        __       ___  _   __         _  _
+-- |   | |       |  | |\ |  |  |_| |  | |  |  |_ |_|
 -- |___| |______ |__| | \|  |  | \ |__| |_ |_ |_ |\
 -- |
 -- |
@@ -55,7 +55,7 @@ local function update_real_port_states(pos, rule_name, new_state)
 		L[i] = n % 2
 		n = math.floor(n / 2)
 	end
-	--				   (0,-1) (-1,0)	  (1,0) (0,1)
+	--                   (0,-1) (-1,0)      (1,0) (0,1)
 	local pos_to_side = {  4,	 1,   nil,   3,	2 }
 	if rule_name.x == nil then
 		for _, rname in ipairs(rule_name) do
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -1,8 +1,8 @@
---        ______
---       |
---       |
---       |        __       ___  _   __         _  _
--- |   | |       |  | |\ |  |  |_| |  | |  |  |_ |_|
+--		______
+--	   |
+--	   |
+--	   |		__	   ___  _   __		 _  _
+-- |   | |	   |  | |\ |  |  |_| |  | |  |  |_ |_|
 -- |___| |______ |__| | \|  |  | \ |__| |_ |_ |_ |\
 -- |
 -- |
@@ -55,8 +55,8 @@ local function update_real_port_states(pos, rule_name, new_state)
 		L[i] = n % 2
 		n = math.floor(n / 2)
 	end
-	--                   (0,-1) (-1,0)      (1,0) (0,1)
-	local pos_to_side = {  4,     1,   nil,   3,    2 }
+	--				   (0,-1) (-1,0)	  (1,0) (0,1)
+	local pos_to_side = {  4,	 1,   nil,   3,	2 }
 	if rule_name.x == nil then
 		for _, rname in ipairs(rule_name) do
 			local port = pos_to_side[rname.x + (2 * rname.z) + 3]
@@ -331,16 +331,16 @@ local function create_environment(pos, mem, event)
 		env[name] = _G[name]
 	end
 
-    env.pcall=function(...)
-        local pcr={pcall(...)}
-        if not pcr[1] then
-            if pcr[2]~=timeout_error then
-                error(pcr[2])
-            end
-        end
-        return unpack(pcr)
-    end
-    
+	env.pcall=function(...)
+		local pcr={pcall(...)}
+		if not pcr[1] then
+			if pcr[2] ~= timeout_error then
+				error(pcr[2])
+			end
+		end
+		return unpack(pcr)
+	end
+	
 	return env
 end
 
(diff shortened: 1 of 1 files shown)


mesecons_luacontroller/init.lua
```diff
@@ -396,9 +502,9 @@ local function run(pos, event)
 	-- Load code & mem from meta
 	local mem  = load_memory(meta)
 	local code = meta:get_string("code")
-
-	local err = code_prohibited(code)
-	if err then return err end
+    local err
```
ShadowNinja: This indentation is broken.

mesecons_luacontroller/init.lua
```diff
@@ -396,9 +502,9 @@ local function run(pos, event)
...
+	code,err = code_prohibited(code)
```
ShadowNinja: Space after comma.

ShadowNinja:
This has some style issues, see http://dev.minetest.net/Lua_code_style_guidelines.

Implementing a lexer is very dangerous, as it's very easy to make a mistake and allow dangerous code.

Can you clarify the pcall issue? (send me an email or other PM since it's sensitive)

mesecons_luacontroller/init.lua
```diff
@@ -330,29 +332,133 @@ local function create_environment(pos, mem, event)
...
+    --(dummy) goto
+    
+    local lexed,err=lexLua(code)
+    local rcode=""
+    if err then
+        return nil,"Couldn't lex code:"..err
+    end
+    for _,v in ipairs(lexed) do
+        --remove useless stuff since we're going over it anyway.
+        if v[1]=="whitespace" then v[2]="\r\n" end
```
ShadowNinja: The `\r` is unnecessary.  In fact ` ` should do.

HybridDog:
>This has some style issues (most notable using four spaces instead of tabs), see http://dev.minetest.net/Lua_code_style_guidelines.

This is mesecons, not minetest_game

gamemanj:
It's been *ages* since I thought about it, I have no idea what it was. I'll PM you with what I think I was probably thinking at the time based on what I changed, though.

gamemanj:
How on earth did I mess up so much of the formatting...?
Thanks for pointing it out, I'll try and make it consistent.

gamemanj:
Aaand now I quickly need to un "fix" the formatting in the ASCII-art at the top...

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
 	if not jit then
-		return false
+		return code
+	end
+	--Find the following constructs, and place dummies where (dummy) is:
+	--while ... do (dummy)
+	--for ... do (dummy)
+	--repeat (dummy)
+	
+	-- This only exists in Lua 5.2
```
ShadowNinja: `Lua 5.2+ / LuaJIT` to be precise (but, minor thing).

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
 		env[name] = _G[name]
 	end
 
+	env.pcall=function(...)
+		local pcr={pcall(...)}
+		if not pcr[1] then
+			if pcr[2] ~= timeout_error then
```
ShadowNinja: Shouldn't this be `if pcr[2] == timeout_error`?  Also, it can be combined with the previous conditional to reduce indentation (`if not pcr[1] and pcr[2] <op> timeout_error then`).

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+local function lexLua(code)
+	local lexElements = {}
+	--Find keywords, whitespace, strings, and then everything else is "cleanup"
+	
+	--Keywords and numbers.
+	function lexElements.keyword(str)
+		return str:match("^[%w_]+")
+	end
+	function lexElements.whitespace(str)
+		return str:match("^[\r \t]+")
```
ShadowNinja: I think `\n` also counts as whitespace.  ;-)
Also, if you missed this one I'd check what other characters you might be missing.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+	end
+	--Unimplemented stuff goes here.
+	function lexElements.cleanup(str)
+		return str:match("^.")
+	end
+	function lexElements.string(str)
+		--Now parse a string.
+		local quoteType = str:sub(1,1)
+		if quoteType ~= "\"" and quoteType ~= "\'" then return nil end
```
ShadowNinja: There are also `[[`/`]]` strings, and the `[=[`/`]=]`, etc, variations.  These don't process escapes though.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+				if c == quoteType then return rstr end
+				if c == "\\" then inEscape = true end
+			end
+		end
+		return nil --unfinished string
+	end
+	function lexElements.blockcomment(str)
+		local a = str:match("^%-%-%[%[")
+		if not a then return nil end
+		local s = str:find("%-%-%]%]")
```
ShadowNinja: `str:find("--]]", 1, true)` would be faster and more readable.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+				inEscape = false
+			else
+				if c == quoteType then return rstr end
+				if c == "\\" then inEscape = true end
+			end
+		end
+		return nil --unfinished string
+	end
+	function lexElements.blockcomment(str)
+		local a = str:match("^%-%-%[%[")
```
ShadowNinja: `str:sub(1, 4) == "--[["` should be much faster.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+		if not a then return nil end
+		local s = str:find("%-%-%]%]")
+		if not s then return nil end
+		return str:sub(1,s+3)
+	end
+	function lexElements.linecomment(str)
+		return str:match("^%-%-[^\r\n]+")
```
ShadowNinja: Line comments can be empty, this should be `"^%-%-[^\r\n]*"`.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+	end
+	-- Note: EOL is not reliable for linecounting purposes, but keeps the code better formatted, if anything
+	function lexElements.eol()
+		return str:match("^[\n]+")
```
ShadowNinja: `\r\n` and `\n\r` are also used for EOL.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+	end
+	local lexElementsOrder = {"keyword", "eol", "whitespace",
+		"string", "blockcomment", "linecomment", "cleanup"}
+	local lexResults = {}
+	while code:len() > 0 do
+		--Because break doesn't exist.
+		local function parseElem()
```
ShadowNinja: I'd move this function out to the file scope to reduce indentation and possibly a new function instantiation on each loop.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
+			for _,v in ipairs(lexElementsOrder) do
+				local t, e=lexElements[v](code)
+				if t then
```
ShadowNinja: I'd find this much easier to read with less single-character variable names (`v`, `t`, `e`).  Perhaps `lexEl`, `match`, `error`?  The third one doesn't seem to be used though.

ShadowNinja:
Thanks, I understand the issue now.

Shouldn't `xpcall` be supported through a wrapper too though (the callback would be wrapped so that it would just be ignored if there was a timeout)?

Your code isn't completely cleaned up yet, there are still a number of places without spaces around binary operators (`env.pcall=function`, `for i=1, str:len()`), and before comment text (`--Find ...`).  There's also some "smooshing" (`blockcomment` -> `block_comment`).

Your lexer seems to have some issues (eg, the following code should loop infinitely)
```Lua
--[[]]
-- Lexer thinks the block comment hasn't ended yet
repeat until false
```
It also seems to do some odd things, like think `123` and `1var` are keywords, ignore anything that's unrecognized, and consider EOL characters as seperate from whitespace characters.

I'm O.K. with the inclusion of the `pcall` fix, but I think the lexer is just too dangerous.

Here's an alternative idea to fix the keywords-in-comments issue: Just insert the dummy in constructs that look like loops without looking for comments.  This will add dummys in comments, but that doesn't really matter since only the Lua interpreter will see the modified code, and it will ignore anything in comments.

mesecons_luacontroller/init.lua
```diff
@@ -329,29 +331,139 @@ local function create_environment(pos, mem, event)
...
 	end
-	local prohibited = {"while", "for", "do", "repeat", "until", "goto"}
-	code = " "..code.." "
-	for _, p in ipairs(prohibited) do
-		if string.find(code, "[^%w_]"..p.."[^%w_]") then
-			return "Prohibited command: "..p
+	for _, v in ipairs(lexed) do
+		--remove/reduce useless stuff since we're going over it anyway.
+		if v[1] == "whitespace" then v[2] = " " end
+		if v[1] == "eol" then v[2] = "\r\n" end -- Windows users may like the \r
```
ShadowNinja: The transformed code is only seen by the Lua interpreter, so this doesn't matter.

gamemanj:
I hadn't noticed the issues in the lexer - They're definitely dangerous, now I see them.
Seeing as that's the case, should I refactor the PR to remove the code, or should this be closed and the pcall/xpcall stuff integrated by you?
I have no idea why xpcall wasn't supported, I'm guessing I noticed the problem and immediately panicked.

ShadowNinja:
I've added your `pcall` fix and another related fix on my [luac-fix-pcall](https://github.com/ShadowNinja/minetest-mod-mesecons/commits/luac-fix-pcall) branch.  I don't have write access to this repo though, so We'll have to wait for @Jeija.

mesecons_luacontroller/init.lua
gamemanj: (Note to self: How on earth did I end up writing this mess, exactly????)
I generally don't combine conditions. Absolutely no idea on the "should equals" thing, but it definitely seems like this whole PR was a complete mess and I somehow didn't realize. I just retested as a "just in case", and for some reason, though it *does* stop the pcall attack, it *also* stops pcalls from working in general.
Further looking at it suggests that Lua's stacktrace interferes with my attempt at detecting the timeout, so for some reason I reversed the check and since that made pcalls not crash the system, I just left it that way. It seems I have a tendency to be a complete idiot at times. Oh well, could've been *worse* (it could've done absolutely nothing to help), but still.
I'm closing the PR on the basis of all the flaws, feel free to build your own solution to the pcall thing.

gamemanj:
How did I mess this up so badly... *facepalm*

PR #233, opened 2015-05-11
Don't refer to LIGHT_MAX global; use default.

technomancy:
The LIGHT_MAX global is created in the legacy.lua file in the default
mod; it states that it's there for backwards compatibility, but it would
be better to reference the proper value inside the default table.


Jeija:
Thanks!

PR #230, opened 2015-04-02
use the on_punch given by default to a node for switches and fix whitespaces

HybridDog:
http://dev.minetest.net/minetest.register_node#on_punch
https://github.com/Jeija/minetest-mod-mesecons/pull/230/files?w=1

PR #239, opened 2015-10-01
Use right mouse button instead of left to press button and change state of switch and lever.

beyondlimits:
The left mouse button is used to press button and change state of switch and lever while the same button is used to dig nodes, which makes bad composition especially when we want to pickup levers and button without firing them. However, the right button isn't used for anything. Why not to use it then?

### DIFF
--- a/mesecons_button/init.lua
+++ b/mesecons_button/init.lua
@@ -40,7 +40,7 @@ minetest.register_node("mesecons_button:button_off", {
 	},
 	groups = {dig_immediate=2, mesecon_needs_receiver = 1},
 	description = "Button",
-	on_punch = function (pos, node)
+	on_rightclick = function (pos, node)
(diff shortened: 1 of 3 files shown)


Jeija:
This was initially decided against since minecraft also used the left mouse button for levers and buttons, so the idea was to make migration for minecraft users to minetest easier. But now that they have changed that, I agree with your opinion.

HybridDog:
> But now that they have changed that, I agree with your opinion.

you don't need to make this mod similar to the other game

Jeija:
Well, in the beginning the idea behind mesecons was to make switching to minetest / starting with minetest (while most of your friends use minecraft) possible for those people that cared about using redstone. Obviously, mesecons has grown to become more than that. Even if minecraft hadn't changed their controls, this change would have made sense, now that they also use the right mouse button, that is even more the case.

I'm sorry, I must have completely missed #230. The punching switches part is propably not neccesary anymore, but I will remove the whitespace.

PR #241, opened 2015-11-21
Fix some LuaController issues

ShadowNinja:
* Possible lockups ("security" issue but not too severe since it's just DoS).
  * Fixed some error messages
  * Fixed a global lookup.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -344,7 +344,7 @@ local function create_environment(pos, mem, event)
 	function env.pcall(...)
 		local res = {pcall(...)}
 		if not res[1] and is_timeout(res[2]) then
-			error(timeout_error)
+			error(timeout_error, 0)
 		end
 		return unpack(res)
 	end
@@ -360,7 +360,7 @@ local function create_environment(pos, mem, event)
 		err_handler = err_handler_wrapper(err_handler)
 		local res = {xpcall(f, err_handler, ...)}
 		if not res[1] and is_timeout(res[2]) then
-			error(timeout_error)
+			error(timeout_error, 0)
 		end
 		return unpack(res)
 	end
@@ -371,7 +371,7 @@ end
 
 local function timeout()
 	debug.sethook()  -- Clear hook
-	error(timeout_error)
+	error(timeout_error, 2)
 end
 
 
@@ -403,7 +403,7 @@ local function create_sandbox(code, env)
 		debug.sethook(timeout, "", 10000)
 		local ok, ret = pcall(f, ...)
 		debug.sethook()  -- Clear hook
-		if not ok then error(ret) end
+		if not ok then error(ret, 0) end
 		return ret
 	end
 end
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -370,12 +370,12 @@ end
 
 
 local function code_prohibited(code)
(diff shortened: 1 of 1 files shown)


ShadowNinja:
*bump* @Jeija?

Jeija:
Could you please post (or better email me if it isn't trivial) a short code snippet on how to cause the DoS with the luacontroller? Just disallowing the `function` keyword will severely impact backwards compatibility and certainly break lots of mesecons creations, so I really don't want to do that if it can be avoided (or at least add a setting for it).

The code snippet
```Lua
function evil()
  evil()
end
evil()
```
that is mentioned in #253 does not cause any DoS for me, neither in LuaJIT nor in bundled Lua.

In other words, can you tell me what is wrong about bd1766e448b149f7245c3d2db18af7ecc984f7ca (the instruction counter patch)?

The global lookup fix and removing xpcall and pcall (I think it is ok to remove them entirely since I don't think anyone seriously uses them in luacontroller code. Especially after removing functions, what is the point of them??) are things I agree with and would certainly be willing to merge.

ShadowNinja:
@Jeija:
This will never finish with LuaJIT 2.0.4:
```Lua
debug.sethook(function() error("Timeout") end, 10)
function f() return f() end f()
```
This is possible because of Lua's tail call optimization.  If you don't return the result of `f` it won't be optimized and you'll soon get a stack overflow, which `pcall` would handle.

I'm not sure if pcall is necessary, but it may be useful, so I'l just fix it.

pgimeno:
Agreed with @ShadowNinja. This freezes LuaJIT for some time when put in a LuaController:

```lua
local function tail_recursive_loop(i)
  if i == 10000000000 then print("finished") return end
  return tail_recursive_loop(i+1)
end

tail_recursive(1)
```
During that time, no other mods execute and exiting the game will make it be stuck in the "Shutting down" screen. When the loop finishes, everything returns to normal. This can be used as a DoS.

Detecting a tail-recursive function is much more difficult than the simple lexical analysis done in #254; it basically requires analyzing the full grammar of the Lua code, basically replicating what Lua does. The easy way out is just to disable `function`.

Jeija:
This might be a dumb question: Wouldn't it just be possible to count function calls when using LuaJIT and then abort after some too many function calls like I implemented [here on another branch](https://github.com/Jeija/minetest-mod-mesecons/commit/5e82cec1dd70385479dc1b0d301a0bb54a5fa608)?

At least the code snippets you posted cause a timeout error with my commit, without having to completely remove the `function` keyword.

I originally didn't realize the importance of this issue (DoS), otherwise I would have certainly responded earlier, so sorry for the delay. It would be great if you could post the "evil" code snippet right from the beginning next time, that would make it much easier to understand.

pgimeno:
After a bit of research I see that others are doing something similar. http://lua-users.org/lists/lua-l/2011-04/msg00092.html

I've also found another problem which I will report in a separate issue.

Jeija:
Okay, I'll merge 08b14e3af0384bf23de3fa976ae94e212819218e into master then. Even if it may not fix all of the issues, at it shouldn't make things worse. I'll leave this issue open for now, you can close it if you think my commit solved the problem. If there is something I missed with 08b14e3af0384bf23de3fa976ae94e212819218e (and there might very well be), please report it!

Thank you for your contributions!

Jeija:
This should now be resolved by 6ba73babcbbce4d63d66158bc5879b7881e79060, closing.

PR #244, opened 2015-12-27
Fix sound volume level of noteblock.

sofar:
Playing stereo sounds positionally in OpenAL causes it to play
the sounds unattenuated - same volume for all distances. This
shouldn't happen, and makes noteblocks unneccesarily loud from
very far away.

Convert all noteblock samples to mono, 22kHz 64kbit.


Jeija:
Thank you! This certainly does sound better.

PR #245, opened 2015-12-30
Update noteblock

HybridDog:
Shorter code, there was unnecessary stuff in the node definition etc.

https://github.com/Jeija/minetest-mod-mesecons/pull/245/files?w=1


Jeija:
I'd love to merge this cleanup, but it just isn't backwards compatible. It used to be possible to double the (piano) sound frequency by placing a steelblock underneath the noteblock:
```Lua
if block_below_name == "default:steelblock" then
	soundname=soundname.."2" -- Go up an octave.
end
```
That's no longer possible. Also, the `node.param2` to sound mapping should remain the same. Backwards compatibility of noteblocks is important compared to backwards compatibility of other nodes since people may have put a lot of effort in using noteblocks to make sounds or even short songs.
Also the sound `["default:steelblock"] = "mesecons_noteblock_hihat2"` does not exist. I think you might have misunderstood what that code code that appends `2` does (which is understandable since the code is not very clean).
Thank you for your effort! If you can fix the issues I mentioned, I will sure merge this.

HybridDog:
fixed, sorry l misunderstood code

C1ffisme:
@Jeija did my pull request add that? I remember making something like that before.

Jeija:
Yes, @C1ffisme, you added this feature in https://github.com/Jeija/minetest-mod-mesecons/commit/11cf727bfbe71874a5f5c77aa83ae4145754f98c. But I like it that way since you can easily transpose the sound by one octave.

@HybridDog: There was more issue: In your param2 <--> sound table, you start with 0 <--> C. If I had to write noteblocks from scratch now, I would have also started that way. Unfortunately, the current code starts with 0 <--> C#, so your changes make every sound 1 half tone lower. I'd say it is better to change that for backwards compatibility. This is why I changed the order in table; otherwise I merged everything the way you wrote it.

C1ffisme:
Actually, it is impossible for the sound "mesecons_noteblock_hihat2" to play, since that sound would require a block other than the steelblock to be under it.

If this could really happen, we could always just go into Audacity and edit the sounds real quick.

HybridDog:
>[…] we could always just go into Audacity and edit the sounds real quick.

My audacity often crashes when l press stop quickly after pressing play
and l can't install a newer version because the builds for my 14.04 are failed ones:
https://launchpad.net/~audacity-team/+archive/ubuntu/daily/+packages

C1ffisme:
Hmm... Mine still works, I could try and do this if I get to it.

And you still have to consider...

> Actually, it is impossible for the sound "mesecons_noteblock_hihat2" to play, since that sound would require a block other than the steelblock to be under it.

HybridDog:
> […] consider...[…]

sorry, l was a bit offtopic

Issue #220, opened 2015-02-12
Doors problems


C1ffisme:
I think I found a problem. I have the latest versions of mesecons and minetest (mostly), and the doors won't open when powered by mesecons. At first I thought that my steel doors wouldn't work because of protection, and then I found that wooden doors wouldn't work either. I want both doors to open again at a mesecon signal. Can this be fixed?

cheapie:
Do some thing like this: 

D
D
?
|

where D is the door (2m tall), ? is any node, and | is a vertical mesecon.

C1ffisme:
@cheapie It didn't work.

MT-Modder:
I think I figured out what the problem is. Place the door before placing the mesecon wires. The usual setup still works. The wires don't update after the door is placed. This is a weird bug.

Jeija:
I can confirm that, @MT-Modder.
The reason for this is simple: Mesecons updates the wires as soon as a node is placed that has some connection rules registered. That, however, is not the case with the doors: If a door is placed, you in fact only trigger a on_place on a craftitem that then manually calls set_node to "build up" the door.
The best way to fix this is to get minetest_game to use `minetest.place_node` in `doors/init.lua` instead of `minetest.set_node`, I will send them a pull request.

MT-Modder:
Awesome. The doors mod really needs to be rewritten. I would do it but pilzadam said he would.

Jeija:
Hmm, I just had a closer look at the issue. Doesn't look that great, because minetest.place_node just completely ignores any param2 given to it. So doors would connect to mesecons just fine, but their facedir will be messed up. A solution to this might be to keep using `minetest.set_node` and trigger on-place callbacks in `core.registered_on_placenodes` manually, but I have no idea if that is event possible, it certainly doesn't feel right.

MT-Modder:
I think it's possible but I've read it's not recommended as it would be called every time any node is placed instead of just when that specific node is. You've tested the minetest.place_node method? If it does ignore parameters in the node table then I would think this is a bug in the engine, no?

MT-Modder:
I just did some testing with ```minetest.place_node`` and got the same results as you. I'm going to submit an issue for the engine. This doesn't seem right IMO.

MT-Modder:
Okay I think the best solution to this problem would be to use one of the door nodes as the inventory item instead of a craftitem. This of course would require an entire rewrite of the doors mod.

sofar:
[...]

PR #249, opened 2016-01-24
Doors: Use new mesh door API if available

sofar:
I've implemented a new door API in minetest_game that
performs all the needed things to assure doors are
properly openend and closed, without mods needing to
know the inner details of what needs to be done.

Mesecons can just fetch a reference to the door object
and call the appropriate open or close method, which
simplifies this code a lot.

For compatibility, this code retains the old code path
and tests whether the new API is available, so this
code remains functional if the new API is not available.

Since the metal trapdoor was only recently added, I added
it to the new API codepath only, it's unlikely to be
present on older versions of minetest_game anyway.

As a benefit from the new door API, there is now
absolutely no more accidental switching closed trapdoors
to open on power off, which could happen with the
old method - since that was just a simple toggle.

### DIFF
--- a/mesecons_doors/init.lua
+++ b/mesecons_doors/init.lua
@@ -20,31 +20,53 @@ local function on_rightclick(pos, dir, check_name, replace, replace_dir, params)
 end
 
 local function meseconify_door(name)
-	if not minetest.registered_items[name] then return end
+	if minetest.registered_items[name .. "_b_1"] then
+		-- old style double-node doors
+		local function toggle_state1 (pos, node)
+			on_rightclick(pos, 1, name.."_t_1", name.."_b_2", name.."_t_2", {1,2,3,0})
+		end
 
-	local function toggle_state1 (pos, node)
-		on_rightclick(pos, 1, name.."_t_1", name.."_b_2", name.."_t_2", {1,2,3,0})
+		local function toggle_state2 (pos, node)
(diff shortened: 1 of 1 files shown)


sofar:
@Jeija hey, since this door api is now merged in minetest_game, mesecons crashes the server if you don't merge this patch. I suggest you merge this ASAP.

Jeija:
Thank you, I have just merged your code (with some very slight style changes). However, there is one problem with doors that mesecons has had for a longer time, here is how to reproduce:
* Place some mesecon wires first
* Then place a door (any normal one, not a trapdoor) next to the wire
Issue: The wire won't automatically connect to the door!
This confuses many new users as the door just doesn't appear to work. When placing the door first, however, the wire connects just fine.

I would love to solve this problem, but there is no clean way for me to do it. As doors are craftitems, not nodes, no `minetest.register_on_placenode` event is fired since the doors code places the door nodes with `set_node`, not `place_node`. The only way to solve this I know is to override the whole `on_place` function of doors, which esentially means replicating most of the door code in mesecons.

@sofar Since you wrote the doors code, do you know a good solution to this problem? Maybe doors could use `place_node` instead or provide some kind of hook function?

Jeija:
I will close this PR since I merged it manually and we will track the issue I just mentioned in #220 / minetest_game.

sofar:
https://github.com/minetest/minetest_game/pull/852

PR #251, opened 2016-02-22
Fix crash due to force loading changes.

ssieb:


JurajVajda:
:+1:

Jeija:
Thank you, makes sense.

PR #252, opened 2016-02-27
fix door compatibility

HybridDog:
toggle doors instead of opening and closing them, fixes solar panel flowers

screenshot_20160227_114443

doors can be placed in a way that closing them means opening them, mesecons supported this, by closing and opening doors instead of toggling them, lots of automatically opening doors get broken, and if this gets merged, breaking those is stopped

### DIFF
--- a/mesecons_doors/init.lua
+++ b/mesecons_doors/init.lua
@@ -52,13 +52,13 @@ local function meseconify_door(name)
 				action_on = function(pos, node)
 					local door = doors.get(pos)
 					if door then
-						door:open()
+						door:toggle()
 					end
 				end,
 				action_off = function(pos, node)
 					local door = doors.get(pos)
 					if door then
-						door:close()
+						door:toggle()
 					end
 				end,
 				rules = mesecon.rules.pplate
@@ -95,13 +95,13 @@ if doors and doors.get then
 			action_on = function(pos, node)
 				local door = doors.get(pos)
 				if door then
-					door:open()
+					door:toggle()
 				end
 			end,
 			action_off = function(pos, node)
 				local door = doors.get(pos)
 				if door then
-					door:close()
+					door:toggle()
 				end
 			end,
 		}},


sofar:
This is unneeded, just rotate the trapdoor with the screwdriver to work the other way around.

Now you're making a mess of doors, where if a person opens it manually, the pressure plate closes the door in their face.

Not a fan, at all, of this chance. I purposely made screwdrivers rotate trapdoors in the first place...

Jeija:
I already feared there would be some obscure usecase for the toggle functionality when merging sofar's door patch...

There is just one problem here: This fixes one thing, but introduces a new bug. Steps to reproduce:
* Place a door
* Place a pressure plate in front of the door and behind it
* Step on the first pressure plate. The door opens.
* Now quickly get off the first pressure plate and step on the second pressure plate.
* Result: The door closes, even though it is powered by one pressure plate.
(You can also just do this with two switches / wires attached to the door)

This causes doors seemingly randomly changing their toggle status when just casually walking through them using the pressure plate system. This has been a well-known bug for quite some time now.

This happens, because mesecons requires each effector nodename to be either "onstate" or "offstate". Nodes with `action_on` in their effector definition are offstate, nodes with `action_off` in their definition are onstate. This means that `_a` doors should always be onstate and `_b` doors always offstate. When allowing for toggling, that is simply not the case anymore, since both can be both states. That essentially causes the `action_on` signal firing even though the door is already onstate and the `action_off` signal not firing because the door is powered which causes the state toggle bug.

As a workaround I have restored the old toggle behaviour for trapdoors, but kept the new behaviour for normal doors. If you need normal doors to behave as if they were inverted, you could (in most cases) just place them rotated by 90°.

If you (or @sofar as the door code author) have any idea on how to fix the onstate / offstate detection for normal doors (without rewriting mesecons core), I'd be happy to get your suggestions to bring back the old behaviour without bugs.

sofar:
![]
(http://i.imgur.com/i65KFOI.png)

without this patch. just used screwdriver.

Jeija:
@sofar Didn't know about the screwdriver honestly, but sounds like a far better solution. Currently trying to reproduce the sunflowers just using the screwdriver...

sofar:
rotating is a bit tricky, since it's hard to see which way the door will open, but trying to open/close the trapdoor after each rotate helps.

sofar:
The core problem is timer usage to detect players. node timers don't fire often enough to be accurate

the only way to work around it is to keep either a state list, or some queue of sorts. but that's terrible.

Jeija:
The best solution to the pressure plate problem are propably client-side mods. Only then can I accurately predict whether a player is standing on the plate or not. The core problem of the toggle bug is that with @HybridDog's patch doors no longer know what onstate and offstate is.

I've had some success with the screwdriver, but it certainly is way harder than with toggle trapdoors...

HybridDog:
using the screwdriver worked, but it took time to find out the right param2s ({[10]={-1,0}, [19]={0,-1}, [4]={1,0}, [13]={0,1}} -- [param2]={rz,rx})
l guess l need to use the replacer to fix my older flowers.

HybridDog:
screenshot_20160227_212952

Now the flowers work better without toggling, else this happens:
screenshot_20160227_213537

Jeija:
Okay, I reverted the toggling commit. I'm sorry to have introduced incompatibilities with the initial change, but I think it was a necessary and useful bugfix in the long run. Closing this issue.

PR #253, opened 2016-03-10
Remove 'do' and add 'function'.

pgimeno:
'do' is not a looping construct, and is useful for restricting scope context, as in
  local a = 2
  do
    local a = 3
  end
  -- a is 2 here
or to force early return (usually when debugging) as in
  do
    return
  end

'function' can be used to crash the server by making a function recursive or tail-recursive, as in

  function x()
    x()
  end

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -350,7 +350,7 @@ local function code_prohibited(code)
 	if not jit then
 		return false
 	end
-	local prohibited = {"while", "for", "do", "repeat", "until", "goto"}
+	local prohibited = {"while", "for", "repeat", "until", "goto", "function"}
 	code = " "..code.." "
 	for _, p in ipairs(prohibited) do
 		if string.find(code, "[^%w_]"..p.."[^%w_]") then


pgimeno:
I've just realized that the pull request from @ShadowNinja already adds the 'function' keyword. Removing 'do' is still advisable for the above reasons.

Jeija:
Removing `do` makes sense. Will do that when #241 is resolved.
Thanks for reporting!

C1ffisme:
While were here, any chance that these words in a string can be ignored? It's annoying when you have a digiline LCD and you need to write a sentence like "Cobblestone blocks **for** just 1 gold ingot!"

Jeija:
@C1ffisme I know that this is annoying and there have been many attempts at fixing this situation in the past (e.g. #115) which have all failed because it is a really difficult problem to solve and we can't afford to make any mistakes. #254 is a new attempt at this (and propably the best I have seen so far), but not thoroughly tested yet.

C1ffisme:
@Jeija An easy solution would be to not use any prohibitions, but overheat the controller if the server is lagging too much. This means other methods of lag (*cough* Blinky Plants *cough*) would also kill the luacontroller, but it would fix this problem.

I suppose if there was a way to check how much recurring happened.

Jeija:
@C1ffisme Yeah right, that is the problem. Since the luacontroller code basically runs on the same level as the mod code, but in a sandbox so that it can only access some specific functions, there is no way to replace prohibited commands with overheating, at least no way in LuaJIT that I know of.

Jeija:
Should have been solved by 08b14e3af0384bf23de3fa976ae94e212819218e, feel free to reopen if you don't think it was. Thanks for your contribution.

PR #254, opened 2016-03-10
Strip strings and comments prior to checking for prohibited keywords

pgimeno:
Prevents comments or strings containing words in the prohibited keywords set from considering the whole block invalid.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -352,6 +352,65 @@ local function timeout()
 end
 
 
+local function strip_squarebrackets(code, pos)
+	if code:byte(pos) ~= 91 then return end -- [
+	pos = pos + 1
+	local start = pos
+	while code:byte(pos) == 61 do -- =
+		pos = pos + 1
+	end
+	if code:byte(pos) ~= 91 then return end -- [
+	local _, last = code:find('%]'..string.rep('=', pos-start)..'%]', pos + 1)
+	last = last or #code
+	return code:sub(1, start - 1)..code:sub(last + 1, #code)
+end
+
+
+local function strip_strings_and_comments(code)
+	local i = 1
+	local len = #code
+	local start, new
+	while i <= len do
+		local c = code:byte(i)
+		if c == 34 or c == 39 then
+			-- " or '
+			start = i
+			i = i + 1
+			while i <= len and code:byte(i) ~= c do
+				if code:byte(i) == 92 then i = i + 1 end -- \ 
+				i = i + 1
+			end
+			code = code:sub(1, start - 1) .. code:sub(i + 1, #code)
+			i = start
+			len = #code
+		elseif c == 91 then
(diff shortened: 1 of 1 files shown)


Jeija:
While this is certainly not the first attempt at this ( #115 ) it is propably the best and cleanest I have seen so far, haven't been able to write code that breaks this just yet. I will be very cautious when merging this though.

Is there any way you can convince me that this is absolutely secure? E.g. is there any other project that uses similar code, is this code derived directly from the Lua source code or something like that?

Because I fear that we might just come across some weirdness in the Lua language (like `[[ ... ]]` strings, `[===[ ... ]===]` comments) that we didn't think of and this breaks immediately.

And there is one other issue. If we really get down to manually parsing the code, I think it would be beneficial to just invest a little more extra effort and parse the code in a way that allows for using the prohibited commands. At the current stage, e.g. all we need to allow for "while" and "for" is to have a dummy function call in every loop iteration. So if someone writes the code
```Lua
while true do end
```
or
```Lua
for i = 0, <very high number> do end
```
basically all we need to do is somehow add a dummy function call between `do` and `end`, like define a function
```Lua
function dummy()
end
```
and pass it over to the luacontroller environment, then instead of doing all the code parsing to strip strings and prohibited commands, just look for all `do` keywords and replace them with `do dummy()`. That way, the above endless while loop becomes:
```Lua
while true do dummy() end
```
which then causes a timeout as it should.

Am I missing something there? I mean, this code is really complex; I don't think the code to insert the dummy statements would be that much harder, would it?

pgimeno:
No, I don't have arguments to convince you. The code is not translated from the Lua source; it's a straightforward lexical analysis implemented in code, taken from the description of Lua. For performance it's not even implemented as a state machine.

There is in fact one security hole. The quotes are removed together with the contents; this enables an attack. For example:
```lua
function x() end
x"y"while true do end
```
will miss the keyword `while` because after removing the string `"y"`, that line looks like `xwhile true do end`. I'm fixing it by placing a space in the place of the string. The `--` in the comments already stays, so that fix is not necessary there.

To implement what you're requesting, it should be transformed into a proper lexer, a tokenizer, which would directly avoid that kind of issues. I'll see what I can do. It won't look simple, that's for sure. I'll open a new pull request if I manage to complete it. This one can stay as a fallback.

pgimeno:
(Rebased to resolve conflict)

pgimeno:
Found and fixed another security issue with the patch. If the user managed to somehow slip a CR (ASCII 13) after a double dash, all code between that and the next LF (ASCII 10) if any would be unchecked.

Jeija:
Since 6ba73babcbbce4d63d66158bc5879b7881e79060 is merged, there is no more need for this. I will reopen this and consider your code in case there is something fundamentally wrong with 6ba73babcbbce4d63d66158bc5879b7881e79060 and we have to fallback to prohibited commands.

PR #256, opened 2016-03-14
Disable JIT optimization for user code.

pgimeno:
That enables normal working of debug.sethook() even for loops. The drawback is that that code will run more slowly.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -354,22 +354,6 @@ local function timeout()
 end
 
 
-local function code_prohibited(code)
-	-- LuaJIT doesn't increment the instruction counter when running
-	-- loops, so we have to sanitize inputs if we're using LuaJIT.
-	if not rawget(_G, "jit") then
-		return false
-	end
-	local prohibited = {"while", "for", "repeat", "until", "goto"}
-	code = " "..code.." "
-	for _, p in ipairs(prohibited) do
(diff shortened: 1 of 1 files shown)


pgimeno:
Please check if this works. I think it may be the safest way to go. Otherwise I have prepared a patch along the lines discussed in #254 that inserts function calls in potential loops.

Jeija:
Seems to work for me, I have tested it with a small set of Luacontrollers that are programmed with all kinds of loops / gotos / function calls that could cause DoS. This certainly needs a lot more testing, but if it actually works in the end, the drawback of having code in the luacontroller run minimally slower is by far outweighed by the benefit of
* Having no more prohibited commands in the luacontroller
* Finally no feature disparity between Lua / LuaJIT
* Less code to maintain, which hopefully also means better security

Thank you so much for this patch! It is one of those easy solutions I didn't know existed, so I didn't even think of this!

pgimeno:
Yeah, it's been frustrating for me too to come up with it *after* writing the patch :)

Jeija:
Okay, so you and I (and anyone that came across this PR) haven't found any bugs so far. This code won't test itself, so I will have to merge it and see what happens. Worst case scenario is that some server using mesecons crashes and we have to revert this change... But I don't see that happening.

The only things I have changed from your original commit are some comments. I have also prepared a commit for the mesecons.net luacontroller documentation that I'll upload shortly.

PR #257, opened 2016-03-25
New feature: Allow string.find in plain mode.

pgimeno:
The fourth parameter of string.find indicates whether the second parameter should be interpreted literally (true) or as a pattern (false). Allowing patterns enables DoS attacks, but it's possible to allow literal matching with little effort, by disallowing the function only if the fourth parameter (plain mode) is not `true`.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -216,6 +216,17 @@ local function safe_string_rep(str, n)
 	return string.rep(str, n)
 end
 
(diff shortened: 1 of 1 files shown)


Jeija:
Sorry for my late response, I just returned from holidays. I think forcing the programmer to specify the fourth parameter is propably the cleanest solution to this problem (instead of just making `plain` true all the time) since this way the Luacontroller programmer is concious of the limitation.

I merged your PR with a slightly changed error message for clarity.

PR #259, opened 2016-04-05
Add wood and cobble ghoststones, for traps

amhadinger:
I added wood and cobble ghoststones, for traps

### DIFF
--- a/mesecons_random/init.lua
+++ b/mesecons_random/init.lua
@@ -1,85 +1,282 @@
--- REMOVESTONE
+­­ REMOVESTONE
 
 minetest.register_node("mesecons_random:removestone", {
-	tiles = {"jeija_removestone.png"},
-	inventory_image = minetest.inventorycube("jeija_removestone_inv.png"),
-	groups = {cracky=3},
-	description="Removestone",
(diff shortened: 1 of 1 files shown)


C1ffisme:
I think that the wood and cobble ghoststones should have a slightly different texture, so that you have to think about placing it in a strategic place. And it would also be pretty bad if you grabbed the ghoststone blocks in creative rather than the normal ones.

I don't want, nay, can't see someone's house made of ghoststone blocks when they turn the lights on.

sofar:
You completely removed all the indentation. Please fix that first.

Jeija:
I'm not interested in adding more blocks that do the same thing with a different texture to mesecons, there are already enough blocks to maintain as-is (exceptions are when the code is structured in a fashion that makes adding more blocks simply a matter of adding a few lines of code like lightstones, doors). You are free to maintain your mod as an extension to mesecons yourself and people will install it if they find it useful.

Also, when next submitting a PR please make sure you follow basic coding style rules such as intendation. I'm assuming you're using an editor that doesn't support tabs or something.

Anyways, thanks for your contribution. I encourage you to maintain the extra blocks seperately from mesecons, but I won't include them.

PR #263, opened 2016-04-20
fix `remove_functions` stack overflow bug

### LINKED ISSUE #262: Recursive table in luacontroller mem crash
Running the following code on a luacontroller crashes the game:
```lua
mem.t = {}
mem.t.t = mem.t
```

`remove_functions` in `mesecons_luacontroller/init.lua` runs into a stack overflow.  When called on `mem`, it calls itself on `mem.t`, which it calls itself on `mem.t.t`, which calls itself on `mem.t.t.t`, which calls itself on `mem.t.t.t.t`, ...

This problem is easy to detect, by having `remove_functions` maintain a table of tables that have been seen so far.  The question is, what should be done when it is detected?

When this happens, should the the luacontroller be burnt, or should the offending tables be silently removed?  What if you do `mem.t = {}; mem.t2 = mem.t`?  Should that burn the luacontroller, or should one of them be removed (which one?), or should they be left alone, both to be serialized by `save_memory` and then deserialized by `load_memory`, at which point they will have become separate tables?

I can fix this and make a PR for it once we decide how it should be fixed.

electrodude:
This fixes #262.

I added a `local function rfuncs(x)` inside of `remove_functions` that does what `remove_functions` used to do, but then I made `rfuncs(x)` first see if `seen[x]` is `true` and return if it is, and then set `seen[x] = true`, and then go on to do what `remove_functions` used to do.

Then, I made `remove_functions` declare a `local seen = {}` that becomes an upvalue of `rfuncs`, then call `rfuncs(x)`, and then finally return `x`, which now has all of its functions removed.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -229,23 +229,36 @@ end
 
 local function remove_functions(x)
 	local tp = type(x)
-	if tp == "table" then
-		for key, value in pairs(x) do
-			local key_t, val_t = type(key), type(value)
-			if key_t == "function" or val_t == "function" then
-				x[key] = nil
-			else
-				if key_t == "table" then
-					remove_functions(key)
-				end
-				if val_t == "table" then
-					remove_functions(value)
+	if tp == "function" then
+		return nil
+	end
+
+	local seen = {}
+
+	local function rfuncs(x)
+		if seen[x] then return end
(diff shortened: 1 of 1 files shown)


Jeija:
I merged this manually with some small changes to make the code more readable. Thank you very much for this much needed bugfix!

PR #264, opened 2016-04-21
Escape command block commands for formspec

cheapie:
This solves the issue with command blocks with commands containing characters like ; being uneditable.


Jeija:
Thank you!

PR #265, opened 2016-04-26
use the registered mese for powered mese, fixes missing light_source …

HybridDog:
…of powered mese


Jeija:
I liked this idea, but the implementation was unneccesarily complex since we have a neat little function called `mesecon.mergetable` that takes care of all of the logic here. I implemented my own solution in https://github.com/Jeija/minetest-mod-mesecons/commit/4249ed4986fd614019c0350de1808041b15a32cc. Thank you for your contribution!

PR #269, opened 2016-05-16
Add protection support to Luacontrollers

cheapie:
This simply disallows editing programs on Luacontrollers in someone else's protected area.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -508,8 +508,13 @@ local digiline = {
 		end
 	}
 }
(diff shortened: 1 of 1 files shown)


sofar:
Can you perhaps add the check for the `protection_bypass` priv as well here?

cheapie:
minetest.is_protected() doesn't do that?

cheapie:
Added.

sofar:
excellent, :+1:

Jeija:
I haven't really tested this as I'm not running any server that would need protection, but the code looks fine. I squashed to two commits and merged them manually.
Thank you!

PR #271, opened 2016-05-16
Spitted a string on two lines

Emon-Omen:
Cosmetic - prevents the line from going out of the configuration window


Jeija:
I like the idea, but there should be more space between text and the `Submit` button, and also a little more distance to the command window. Could you please make the window just a little larger and move the button and the text down accordingly?

commandblock

Jeija:
PR submitter account doesn't exist anymore, closing.

PR #272, opened 2016-05-17
Don't bother checking if a node can overheat before cooling it

cheapie:
This might /seem/ backwards, but the check is *very* expensive time-wise (as much as 30% of the total lag in the world I tested it with), and not entirely necessary since meta:get_int("heat") returns 0 if called on a meta that doesn't contain "heat".

I did test it quite thoroughly as well, and Nothing Bad Happened(TM).

### DIFF
--- a/mesecons/services.lua
+++ b/mesecons/services.lua
@@ -87,10 +87,6 @@ end
 
 
 mesecon.queue:add_function("cooldown", function (pos)
-	if minetest.get_item_group(minetest.get_node(pos).name, "overheat") == 0 then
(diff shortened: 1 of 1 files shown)


Jeija:
I'm sorry, I totally forgot about this one. This patch should work just fine since `meta:get_int` returns 0 if `heat` was not found. Thank you!
Merged as https://github.com/Jeija/minetest-mod-mesecons/commit/acd41b57733bee4c61729f4ec1ce1742402b59d8

PR #275, opened 2016-07-08
Command block bug fixed

numberZero:
Before that, command block would crash the server if activated when no
players are online


Jeija:
Thank you very much, I can't believe this critical bug went unnoticed (or wasn't reported reported) for so long. I changed your code so that commands that can't be resolved will also be deleted (just to make sure nothing unexpected can happen): https://github.com/Jeija/minetest-mod-mesecons/commit/f9f7600017967672ec28ecf139d9d520d64fcef0

PR #276, opened 2016-07-25
Fix an incorrect variable name.

Hawk777:
There is an incorrect variable name in `mesecon.turnon` and `mesecon.turnoff` functions. This results in errors being printed to the console and, presumably, some other problems. It appears in a rarely taken branch, so could easily not be noticed.

### DIFF
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -394,7 +394,7 @@ function mesecon.turnon(pos, link)
 
 				-- area not loaded, postpone action
 				if not mesecon.get_node_force(np) then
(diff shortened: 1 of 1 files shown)


Jeija:
Thank you very much for this report. However, the fix was more complicated:
We can't postpone the action of turning on /off the neighbor, since we can't possibly know if the rules of the neighbor actually link this condcutor's rules. This is why I made it so that turning on the current conductor is postponed until the neighbor has been loaded.

Hawk777:
Thanks; I didn’t actually take the time to understand the logic, I just looked at the variable names and assumed there had been a missed rename somewhere.

Issue #277, opened 2016-07-25
Forceloading is excessive


Hawk777:
This is a bit of a mix of feature request and discussion topic. I’m not sure if people agree or not.

As written, at present, Mesecons forceloads every mapblock connected along a wire to a signal source when the signal changes state. In my opinion, forceloading is a rather large hammer being used against a rather small target. All we care about is tracing through the mapblocks to see where the wire connects to (and to change some nodes), yet to accomplish this, we will burn the global forceloading budget and activate a lot of mapblocks.

What if, instead, in `get_node_force`, we called `get_node_or_nil` first, then, on a `nil` return, used a voxel manipulator to bring the target mapblock into the cache and simply called `get_node_or_nil` a second time? This would lead to all those mapblocks being loaded but inactive, which should be much cheaper. As long as signal edges occur more frequently than once every `server_unload_unused_data_timeout`, everything would live in cache permanently and so the initial `get_node_or_nil` would succeed and no voxel manipulator would ever be created, but we would not pay the cost of the mapblock being active. For slower signals, the blocks would eventually be unloaded, and subsequent signal edges would reload them via the VM.

Thoughts? If people are agreeable, I would be happy to look at implementing this myself and turning this into a PR.

Jeija:
This sound like an interesting idea! I'm not so sure about the performance of VoxelManip and if it is suitable for this purpose.

We are using forceloading for several reasons: First of all, it can be more performant to keep mapblocks in memory instead of loading them every time a signal passes by. But most importantly, not being able to trace back a wire to all of its sources can cause serious bugs: If we have two onstate receptors connected to a long wire at opposite ends, turning off only one of them shouldn't turn off the wire. If we can't trace back the whole wire quickly, we cannot act accordingly.

It sounds like it should be possible to use VoxelManip for this purpose. Just make sure that this doesn't impact performance. Especially in singleplayer it is more important for mesecons to function correctly and also be fast than to save server resources. So all in all, I would be willing to try out an implementation based on VoxelManip :+1:

Hawk777:
Do you have any guidelines for how to figure out what the performance impact is? I have a fairly fast computer, and I don’t personally have any large mesecons networks set up in my own worlds. Is there a sample world with a big machine in it somewhere I could download?

sofar:
[...]

PR #279, opened 2016-08-01
Use a VoxelManipulator to load unloaded areas

Hawk777:
This is a proof of concept of my simplest proposal on GH-277, using the VoxelManipulator to prime the mapblock cache when `get_node_or_nil` fails rather than using forceloading to keep all needed mapblocks in cache at all times.

### DIFF
--- a/mesecons/init.lua
+++ b/mesecons/init.lua
@@ -70,6 +70,8 @@ dofile(minetest.get_modpath("mesecons").."/internal.lua");
 -- these are the only functions you need to remember
 
 mesecon.queue:add_function("receptor_on", function (pos, rules)
+	mesecon.vm_begin()
+
 	rules = rules or mesecon.rules.default
 
 	-- if area (any of the rule targets) is not loaded, keep trying and call this again later
@@ -90,6 +92,8 @@ mesecon.queue:add_function("receptor_on", function (pos, rules)
 			mesecon.turnon(np, rulename)
 		end
 	end
+
+	mesecon.vm_commit()
 end)
 
 function mesecon.receptor_on(pos, rules)
@@ -97,6 +101,8 @@ function mesecon.receptor_on(pos, rules)
 end
 
 mesecon.queue:add_function("receptor_off", function (pos, rules)
+	mesecon.vm_begin()
+
 	rules = rules or mesecon.rules.default
 
 	-- if area (any of the rule targets) is not loaded, keep trying and call this again later
@@ -119,6 +125,8 @@ mesecon.queue:add_function("receptor_off", function (pos, rules)
 			end
 		end
 	end
+
+	mesecon.vm_commit()
 end)
 
 function mesecon.receptor_off(pos, rules)
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -379,42 +379,25 @@ function mesecon.turnon(pos, link)
 		local f = frontiers[depth]
 		local node = mesecon.get_node_force(f.pos)
 
-		-- area not loaded, postpone action
 		if not node then
-			mesecon.queue:add_action(f.pos, "turnon", {f.link}, nil, true)
+			-- Area does not exist; do nothing
 		elseif mesecon.is_conductor_off(node, f.link) then
 			local rules = mesecon.conductor_get_rules(node)
-
-			-- Success: If false, at least one neighboring node is unloaded,
-			-- postpone turning on action
-			local success = true
 			local neighborlinks = {}
 
 			-- call turnon on neighbors
 			for _, r in ipairs(mesecon.rule2meta(f.link, rules)) do
 				local np = vector.add(f.pos, r)
-
-				-- Neighboring node not loaded, postpone turning on current node
-				-- since we can't even know if neighboring node has matching rules
-				if not mesecon.get_node_force(np) then
-					success = false
-					break
-				else
-					neighborlinks[minetest.hash_node_position(np)] = mesecon.rules_link_rule_all(f.pos, r)
-				end
+				neighborlinks[minetest.hash_node_position(np)] = mesecon.rules_link_rule_all(f.pos, r)
 			end
 
-			if success then
-				mesecon.swap_node_force(f.pos, mesecon.get_conductor_on(node, f.link))
+			mesecon.swap_node_force(f.pos, mesecon.get_conductor_on(node, f.link))
 
-				for npos, links in pairs(neighborlinks) do
-					-- links = all links to node, l = each single link
-					for _, l in ipairs(links) do
-						table.insert(frontiers, {pos = minetest.get_position_from_hash(npos), link = l})
-					end
+			for npos, links in pairs(neighborlinks) do
+				-- links = all links to node, l = each single link
+				for _, l in ipairs(links) do
+					table.insert(frontiers, {pos = minetest.get_position_from_hash(npos), link = l})
 				end
-			else
-				mesecon.queue:add_action(f.pos, "turnon", {f.link}, nil, true)
 			end
 		elseif mesecon.is_effector(node.name) then
 			mesecon.changesignal(f.pos, node, f.link, mesecon.state.on, depth)
@@ -440,40 +423,24 @@ function mesecon.turnoff(pos, link)
 
 		-- area not loaded, postpone action
 		if not node then
-			mesecon.queue:add_action(f.pos, "turnoff", {f.link}, nil, true)
+			-- Area does not exist; do nothing
 		elseif mesecon.is_conductor_on(node, f.link) then
 			local rules = mesecon.conductor_get_rules(node)
-
-			-- Success: If false, at least one neighboring node is unloaded,
-			-- postpone turning on action
-			local success = true
 			local neighborlinks = {}
 
 			-- call turnoff on neighbors
 			for _, r in ipairs(mesecon.rule2meta(f.link, rules)) do
 				local np = vector.add(f.pos, r)
-
-				-- Neighboring node not loaded, postpone turning off current node
-				-- since we can't even know if neighboring node has matching rules
-				if not mesecon.get_node_force(np) then
-					success = false
-					break
-				else
-					neighborlinks[minetest.hash_node_position(np)] = mesecon.rules_link_rule_all(f.pos, r)
-				end
+				neighborlinks[minetest.hash_node_position(np)] = mesecon.rules_link_rule_all(f.pos, r)
 			end
 
-			if success then
-				mesecon.swap_node_force(f.pos, mesecon.get_conductor_off(node, f.link))
+			mesecon.swap_node_force(f.pos, mesecon.get_conductor_off(node, f.link))
 
-				for npos, links in pairs(neighborlinks) do
-					-- links = all links to node, l = each single link
-					for _, l in ipairs(links) do
-						table.insert(frontiers, {pos = minetest.get_position_from_hash(npos), link = l})
-					end
+			for npos, links in pairs(neighborlinks) do
+				-- links = all links to node, l = each single link
+				for _, l in ipairs(links) do
+					table.insert(frontiers, {pos = minetest.get_position_from_hash(npos), link = l})
 				end
-			else
-				mesecon.queue:add_action(f.pos, "turnoff", {f.link}, nil, true)
 			end
 		elseif mesecon.is_effector(node.name) then
 			mesecon.changesignal(f.pos, node, f.link, mesecon.state.off, depth)
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -405,8 +405,7 @@ function mesecon.turnon(pos, link)
 			end
 
 			if success then
-				minetest.swap_node(f.pos, {name = mesecon.get_conductor_on(node, f.link),
-					param2 = node.param2})
+				mesecon.swap_node_force(f.pos, mesecon.get_conductor_on(node, f.link))
 
 				for npos, links in pairs(neighborlinks) do
 					-- links = all links to node, l = each single link
@@ -465,8 +464,7 @@ function mesecon.turnoff(pos, link)
 			end
 
 			if success then
-				minetest.swap_node(f.pos, {name = mesecon.get_conductor_off(node, f.link),
-					param2 = node.param2})
+				mesecon.swap_node_force(f.pos, mesecon.get_conductor_off(node, f.link))
 
 				for npos, links in pairs(neighborlinks) do
 					-- links = all links to node, l = each single link
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -236,16 +236,131 @@ local function unhash_blockpos(hash)
 	return vector.multiply(minetest.get_position_from_hash(hash), BLOCKSIZE)
 end
 
--- get node and force-load area
-function mesecon.get_node_force(pos)
-	local node = minetest.get_node_or_nil(pos)
-	if node == nil then
-		-- Node is not currently loaded; use a VoxelManipulator to prime
-		-- the mapblock cache and try again.
-		minetest.get_voxel_manip(pos, pos)
-		node = minetest.get_node_or_nil(pos)
+-- Maps from a hashed mapblock position (as returned by hash_blockpos) to a
+-- table.
+--
+-- Contents of the table are:
+-- “vm” → the VoxelManipulator
+-- “va” → the VoxelArea
+-- “data” → the data array
+-- “param1” → the param1 array
+-- “param2” → the param2 array
+-- “dirty” → true if data has been modified
+--
+-- Nil if no VM-based transaction is in progress.
+local vm_cache = nil
+
+-- Starts a VoxelManipulator-based transaction.
+--
+-- During a VM transaction, calls to vm_get_node and vm_swap_node operate on a
+-- cached copy of the world loaded via VoxelManipulators. That cache can later
+-- be committed to the real map by means of vm_commit or discarded by means of
+-- vm_abort.
+function mesecon.vm_begin()
+	vm_cache = {}
+end
+
+-- Finishes a VoxelManipulator-based transaction, freeing the VMs and map data
+-- and writing back any modified areas.
+function mesecon.vm_commit()
+	for hash, tbl in pairs(vm_cache) do
+		if tbl.dirty then
+			local vm = tbl.vm
+			vm:set_data(tbl.data)
+			vm:write_to_map()
+			vm:update_map()
+		end
 	end
-	return node
+	vm_cache = nil
+end
+
+-- Finishes a VoxelManipulator-based transaction, freeing the VMs and throwing
+-- away any modified areas.
+function mesecon.vm_abort()
+	vm_cache = nil
+end
+
+-- Gets the cache entry covering a position, populating it if necessary.
+local function vm_get_or_create_entry(pos)
+	local hash = hash_blockpos(pos)
+	local tbl = vm_cache[hash]
+	if not tbl then
+		local vm = minetest.get_voxel_manip(pos, pos)
+		local min_pos, max_pos = vm:get_emerged_area()
+		local va = VoxelArea:new{MinEdge = min_pos, MaxEdge = max_pos}
+		tbl = {vm = vm, va = va, data = vm:get_data(), param1 = vm:get_light_data(), param2 = vm:get_param2_data(), dirty = false}
+		vm_cache[hash] = tbl
+	end
+	return tbl
+end
+
+-- Gets the node at a given position during a VoxelManipulator-based
+-- transaction.
+function mesecon.vm_get_node(pos)
+	local tbl = vm_get_or_create_entry(pos)
+	local index = tbl.va:indexp(pos)
+	local node_value = tbl.data[index]
+	if node_value == core.CONTENT_IGNORE then
+		return nil
+	else
+		local node_param1 = tbl.param1[index]
+		local node_param2 = tbl.param2[index]
+		return {name = minetest.get_name_from_content_id(node_value), param1 = node_param1, param2 = node_param2}
+	end
+end
+
+-- Sets a node’s name during a VoxelManipulator-based transaction.
+--
+-- Existing param1, param2, and metadata are left alone.
+function mesecon.vm_swap_node(pos, name)
+	local tbl = vm_get_or_create_entry(pos)
+	local index = tbl.va:indexp(pos)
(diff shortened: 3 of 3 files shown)


Hawk777:
I tried making a 300-metre-long wire with some lightstone at the end. I instrumented the lightstone to print out messages when turning on and off. The wire worked in that the lightstone printed its messages, after having ensured that nothing was forceloaded and even that the cache was cold (by waiting a minute before flipping the switch).

For some reason the onscreen copy of the wires and lightstone didn’t update far away from me, even having increased View Distance from 100 to 400 and Max Block Send Distance from 9 to 30. I’m not sure why that was. The lightstone definitely received the signal, though, as it printed the relevant message to my terminal.

Hawk777:
I tried instrumenting the code to check if, after doing the VM request, a node was still inaccessible. After a couple of hours running with a blinky plant, no such message had appeared. So I think it’s working.

Jeija:
Very nice! For me, this even seems to run more reliably than forceloading.

There is just one issue I see with this approach:
This totally relies on the fact that minetest will actually load mapblocks that are requested for voxel manipulation into the server cache. I don't think this is something we should rely on; maybe someone has the bright idea to somehow save memory by implementing voxelmanips differently and suddenly mesecons will break for no good reason. Also, I'd hope to get better performance by leveraging the power of VMs. So I'd prefer reading the data from the VMs themselves, which should be possible:
* Calculate the block position and size (make BlockSize `BS = 16` configurable in code, just in case)
* Get the VoxelManip for this mapblock and also make a VoxelArea for the mapblock for utility functions
* Read content IDs and param2 data from VoxelManip (`vm:get_data()`, `vm:get_param2_data()`)
* Get the index of the node (`va:index(p)`) inside the VManip data and use it to retrieve content ID and param2 from the previously read datasets
* Use `minetest.get_name_from_content_id(content_id)` to get a meaningful node name and put it into our own node table: `node = {name = nodename, param2 = param2 }`
* Somehow store content ID, param2 data and VoxelAreas in a global cache that is cleared every globalstep since it is very likely we will need them more than once. This shouldn't affect things like pistons pushing wires etc. since effector actions are executed one globalstep later anyways.
* Return node table

Since `rules` in the effector / conductor / receptor definitions and all other functions only use `name` and `param2` information of nodes, this should work. Node metadata cannot affect rules, this is not stored in the node object, actually your typical `node` table really only consists of these properties
```Lua
{
	param1 = 15,
	name = "mesecons:wire_10100000_off",
	param2 = 0
}
```
all of which can be read from a VM (param1 (light data) isn't even ever used in mesecons).

After having implemented this, all of the complicated logic in `internal.lua` that queues signals in case `get_node_force` doesn't give any result (which, until now, could still happen if a mapblock hadn't been used for some time) can be removed.

If you don't feel like implementing this, I will propably do it myself, but given that I have to study for several exams, I can't predict when that will happen…

Hawk777:
[This page](http://dev.minetest.net/get_node) appears to make voxel manipulators’ cache-filling behaviour part of the specification and not a lucky unspecified accident. I can still go ahead and do this if you think it’s worthwhile for performance reasons. I think if I were to do this I would still include param1 as, even if Mesecons doesn’t use it, there’s no reason some other Mesecons-compatible mod we don’t know about that adds its own conductors might use it (phototransistor?). I agree that queueing deferred edge propagation is no longer necessary if we do this as we can guarantee to see the whole equipotential node in one tick.
Are you sure clearing our local data structures on globaltick is sufficient? Is it not possible that something might happen in between, say, two sensors emitting edges on the same tick which modifies the world in a way that we care about?

Jeija:
I can't know if this will actually increase performance, but if there is a chance to, we better take it since bad performance is the major criticism I've heard from people using mesecons. Mesecons generally does not support conductors changing connection rules based on lighting or anything other than the conductor being dug and placed again, so I don't see any application for `param1` (a phototransistor would have to be implemented as effector + receptor); we can still include it though, even if it is just for following the specification.

Clearing the VMs every globaltick only is also one of my concerns. It should work, but there are several caveats. Every time any mesecon signal is emitted (which can only happen by calling `receptor_on` or `receptor_off`) the execution is queued in the ActionQueue until the next globalstep, when all mesecon actions are executed all at once; logic gates, pistons and all other effector actions are then also queued and will only be executed one globalstep later. So this means we don't have to think about other mods interfering with mesecons, because other mods just cannot be executed in that time.

What we have to care about though is mesecons interfering with itself. Imagine a situation like this one:
screenshot_20160809_112712

When that switch is toggled, both mesecon wires will be turned on or off during the same globalstep. If the left wire was turned on first and the right wire followed without the VoxelManips being updated with the new nodes from the left wire, we would end up with a crossing that appears to only be active for the right wire. So what we need to do is completely abstract everything that happens in mesecons during a globalstep (= all functions that are executed in `minetest.register_globalstep` in `actionqueue.lua`, which is essentially the interal mesecons logic) away from the real minetest world (the one that can be accessed with `get_node` / `set_node`) and make everything happen inside the VMs (an abstracted world that uses mesecon-internal functions).

This means that we basically have to follow this order to propagate all signal for the globalstep (this is basically what `minetest.register_globalstep` in `actionqueue.lua` does on a more abstract level):
* For each signal propagation (the globalstep function calls the queue functions `receptor_on` / `receptor_off` for this, see `mesecon.queue:add_function("receptor_on/off", ...)` in `init.lua`):
  * Get equipotential nodes, using `get_node_force` (propably rename that to `get_node_cache` or something), which also loads VMs. Abort in case we find an onstate receptor amongst those nodes and the current signal is `turnoff`.
  * Write new signal level to equipotential nodes **inside the VMs** (we need something like `set_node_cache`), call `activate` / `deactivate` on effectors. Their actions will be executed one globalstep later!
* For each effector signal (this is what `activate` / `deactivate` / `change` queue functions in `interal.lua`) do
  * Make sure we also use `set_node_cache` / `get_node_cache` for effectors that change themselves or their environment (pretty much every receptor does that), so that nodes are set **inside the VMs**. *This is critical and will break compatibility* as it means that also other mods implementing mesecon effectors would have to adapt to using these new functions!!! But it has to be that way, otherwise writing VMs to that map would overwrite these changes. (*)
* Write VMs to map

(`turnon` / `turnoff` queue functions can be dropped since they are no longer required, as they were only used if an area couldn't be loaded, `update_autoconnect` will propably also need to be adapted to use VMs).

So these were just some more of my thoughts on the change. *Thank you so much for asking these questions (and "forcing" me to write consistent answers)*, they really make me think about all the pitfalls that have to be dealt with. This might become a major change in how mesecons work, so we better get it right.

(*) We can make this transition less painful by giving other mods a timeframe to switch to `mesecon.set_node_cache` which only forwards to `minetest.set_node` before ultimately switching over to the VM system.

Jeija:
Oh, and before doing all of this, we should propably verify the claims that VMs are so much faster, especially for mesecons. Maybe write some test mod that only reads and writes like 10 nodes in a mapblock and see how VMs perform vs `get_node` / `set_node`. Though I could totally see how VMs can be faster given that even for small circuits, mesecons has to use *lots* of `get_node`s to find linking conductors (just toggling a switch is 12 `get_node`s already).

Hawk777:
Instead of using VMs for effectors, what if we kept track of all the effectors we encountered while scanning the equipotential nodes during turnon, committed the VMs, and then notified all the effectors afterwards (perhaps using the actionqueue if we want to do multiple turnons in a tick—we could add a “push-to-current” mechanism to the actionqueue that would push to the queue currently being popped or something)? That would mean breaking compatibility with third-party conductors but not third-party effectors, and the former are much less common. It shouldn’t hurt performance much because a large circuit is likely to have many more conductors than effectors, and a small circuit (where that rule might not be true) doesn’t matter anyway.
We could even make VM-based conductor updating opt-in on a per-conductor-type basis; we try it, and if we see a conductor of an old type then we commit the VMs and go back to `get_node`/`set_node` for the rest of that turnon/turnoff operation. Then we don’t break compatibility at all, and any third-party conductors would just have to opt in to get the performance improvements.

Jeija:
The scanning and keeping track of effectors is basically how the current system works already. Whenever an effector is encountered, an `activate` / `deactivate` / `changesignal` is performed which at that point only adds this function to the ActionQueue to be executed during the next globalstep. This system was implemented for multiple reason, one of them being that this effectively prevents any kind of infinite loop happening, since the server always get some time in between the recomputation steps of the circuit.

Not breaking effector compatibility means splitting the ActionQueue execution into two sections: First, execute all of the effector functions and *then* take care of the `receptor_on` / `receptor_off` functions. I don't like that because it adds complexity and there are valid reasons for effector actions to happen in between conductor signals. E.g. if a piston pushes into a wire, whether a short pulse is received on the other end should depend on wheter the wire was activated before the piston or afterwards. This is currently not implemented and a situation like in this picture will lead to bugs, but breaking the ActionQueue in such a way (by splitting it in two) means that this behaviour can never be defined:
screenshot_20160809_180743
In this picture, for instance, when turning on the switch, no signal pulse should be received on the top left end, since the wire to the NOT Gate is longer, so the piston takes a higher priority of execution and cuts the connection before the NOT gate can fire. When turning off the switch though, a signal will be received on the top left, since the piston retracts before the NOT gate can turn off. As I said, this situation leads to bugs, so that construction won't actually function properly.

A workaround the whole effector problem would be to only use VMs on a per-signal basis, so basically writing the VMs to the map after every single signal propagation. This is easier to implment, but sounds really bad performance-wise though.

I don't see any reason why we would break conductors actually. Conductors are the dumbest components of all, apart from `get_rules` they may not even implement any functions, and in get_rules they don't have access to their location, and therefore no reason to access `get_node` or `set_node`. All of the placing and rule-checking etc. of conductors is taken care of by the internal logic. At least that shouldn't be something to worry about.

VanessaE:
I've only been vaguely following this discussion because VMs tend to confuse me a bit, and this discussion is a bit beyond my Lua skills (else I'd try to help)... but if it helps, you could take the mechanical nature of the piston into consideration when computing the circuit.  No matter how fast it could move (even in the most outlandish theory), it'll still never be able to beat that NOT gate, so that bit of wire on the left should still go high for a moment before the piston cuts it off.  The same should apply to other effectors that are mechanical in nature.

Point being, as long as you're looking at possibly having to change the behavior of a circuit, you should probably aim for a realistic result, at least in this case.

Jeija:
I think the problem is not so much about whether the diode or the piston should be faster (or any other sort of combination, the piston was just an example), most importantly the one that is faster should be predictably be faster every time you build a similar circuit and that is what is currently missing.

Concerning the whole VoxelManip ideas: I'm currently starting to doubt VMs in general. First of all, I don't know how well this plays together with things like metadata, nodetimers etc. and if these could get lost somehow. And secondly, if VMs are so incredibly fast in the first place, why isn't all of the lua code just completely abstracted away from the map? I mean, if there performance gains are as big as the sound, why aren't `minetest.get_node` and `minetest.set_node` implemented in a way that they always load the current mapblock when a node is read / set and commit all their changes after the current globalstep? Has someone tried to implement this?

VanessaE:
VMs are particularly faster than get_/set_node once you get into the dozens-of-nodes range as I understand it, though even a few nodes' worth would probably still be faster via VM than get/set calls.  I've not heard of any problems with things like timers, meta data, and so on.

get_/set_node not loading a block has been a thorn in pretty much everyone's side for a long while; technic has something that might be useful to you:  https://github.com/minetest-technic/technic/blob/master/technic/helpers.lua#L49

This implements pretty much what you've been talking about doing, using VM's to force the engine to load a mapblock, without actually *doing* anything with the VM after that.  As a first step, you should be able to wedge this code into the existing mesecons codebase, i.e. just keep doing everything with get_/set_node.

Jeija:
Thanks for the information about the VM speed. Again, I have to wonder why minetest doesn't use VMs for `get_node` / `set_node` then if it is already faster for just a couple of nodes (and I would assume most of the time mods get / set multiple nodes within a mapblock per globalstep). We should thoroughly test and compare VM speed for our usecase.

The thing you pointed out technic does is basically the exact same thing this PR implements as it stands right now, and that is already a step forward. This PR is incomplete in a sense that with a properly working `get_node_force` we should continue to clean up the mesecons codebase and delete everything that assumes the area might be unloaded (additonally to the criticism I've shared above that this code kind of depends on the implementation of VMs).

VanessaE:
I need to pay better attention, I didn't even notice the title of this issue. :-/

Hawk777:
My original plan, when I was considering writing up the full VM-based implementation, was indeed to use a separate cache of VMs per signal edge propagated. That would ensure that all side effects of signal propagation were committed back to the map before anything else happened on the same tick, and that any other activities that happened between two signal edge propagations would properly be read into the new VMs and affect the second signal edge.

Performance-wise, I think it’s a tradeoff whether that or `get_node` is better. If you have a small number of signal edges being emitted per tick but each signal edge is being emitted into a large equipotential node, then the time spent propagating the signal edge is dominated by the time walking from node to node, and a VM minimizes the per-node cost of accessing those nodes. On the other hand, if you have a large number of signal edges being emitted into many small equipotential nodes, then the time spent propagating is dominated by per-signal-edge costs, which in the VM case means it’s dominated by the cost of shovelling data into and out of VMs.

If we could somehow use a single set of VMs for all signal edges emitted on a tick, that would be the ideal situation performance-wise, but as you pointed out, Jeija, this could change the order of operations in the world. Although if that order is already unreliable to start with, then would it be OK to just leave it unreliable? The idea would be for receptors to notify of their changing outputs, collect all of those into a table, then once per tick propagate all the signal edges to the effectors (remembering the effectors but not notifying them) using a single set of VMs, commit the VMs, and then go through and notify all the effectors.

Jeija:
> Although if that order is already unreliable to start with, then would it be OK to just leave it unreliable?

No. Just because it's broken, doesn't mean we should make it worse  :smirk:.

> My original plan, when I was considering writing up the full VM-based implementation, was indeed to use a separate cache of VMs per signal edge propagated.

I think this is a reasonable plan to start with, and I'm positive this will already be an improvement over the current system (remember that a single switch already has to access 12 nodes without any wires connected). And when we're at that point, it should be much easier to possibly switch to a single set of VMs, since we have a much better understanding of what the ramnifications are.

This plan basically means to commit the VM changes in `mesecons/init.lua` at the end of the functions `mesecon.queue:add_function("receptor_on/off", function (pos, rules) ... end)`. These functions can also be cleaned up, since we don't need to handle unloaded mapblocks anymore.

Unfortuantely, I think the best way to proceed  from that point would be to be to replace `mesecon.turnon` and `mesecon.turnoff` with a two-step process: First, get all of the equipotential nodes with breadth-first search (because `mesecon.turnon` and `mesecon.turnoff` at this point are pretty much doing the same thing) with the possibility to abort that equipotential-node function in case an onstate receptor was found (turnoff will be aborted). Second, apply the new state (on / off) to all conductors, call `activate` / `deactivate` / `changesignal` on the effectors (which just adds the real action to the queue).

That is because all of the functions that somehow access the map (`turnon`, `turnoff`, `connected_to_receptor`, `rules_link_rule_all`, ...) could be used in other places as well (such as `services.lua`), so we can't just replace the functions inside of these with the vmanip functions. It might theoretically work to just replace all of the get_nodes and set_nodes in those functions, but I'm really not sure. It's just a mess...

Jeija:
Actually, there might be a better solution than *having to* rewrite `turnon`, `turnoff` etc.:
We could just have two functions like `mesecon.start_vmanip_cache` and `mesecon.end_vmanip_cache`. We call the former at the beginning of `receptor_on/off` and the latter at the end. Our custom set_node/get_node functions then just check if the vmanip cache is currently active (which is set by `start_vmanip_cache`) and then write to the vmanips; otherwise, they just prime the mapblock cache and perform the normal `get_node` / `set_node`. `end_vmanip_cache` then commits the changes to the map and sets the vmanip cache to inactive.

With that system there is no need anymore to rewrite any of the internal functions, just make them all use our own get_node / set_node functions that take care of everything.

Hawk777:
Wow. I tried, but I got totally lost trying to implement `receptor_off`. Perhaps you could give me a little help by answering a couple of questions?
- What’s the difference between `action_on`/`action_off` and `action_change`?
- In `receptor_off`, if a particular equipotential node doesn’t have any other on-state receptors on it, you turn it off with `turnoff`. Cool. If it *does* have another on-state receptor somewhere, you call `changesignal` with `newstate` set to `mesecon.state.off`, which eventually calls `action_change` with same. Why? Shouldn’t you do nothing here? I mean, the conductor isn’t turning off, right?

Jeija:
Oh yes, always feel free to ask questions! Better ask before you get stuck and frustrated somehwere. Especially since internal mesecons code isn't exactly well-documented (and I do really feel bad about that).

`action_on` and `action_off` are intended for simle effectors such as lightstones, pistons, lamps etc. These effectors will be "on" if **any** of their inputs are powered. Therefore, before calling `action_off` in the receptor_off function, we must check if any other receptor (`connected_to_receptor`) powers them. If this is the case, we don't turn them off. Same is for conductors: If conductors are powered by any other receptor, we just ignore this.

Changesignals / `action_change` is what more complicated devices such as the luacontroller, gates or the delayer uses. These effectors want to be notified each time any of their inputs change so that they can set their outputs accordingly. Therefore, we must notify them no matter if they're connected to another receptor or not. Calling `action_change` on a conductor will just do nothing, it only does something for conductors.

Any effector can register some or all of the functions `action_on`, `action_off`, `action_change` and will be notified accordingly.

This stuff has some historical reasons (luacontrollers, delayers etc. were only implemented later, so there was only `action_on` and `action_off` in the beginning), but is also reasonable to keep since it makes writing effector and receptor code easier.

(Technically, you should be able to implement the caching functions without actually understanding any of the internal logic (I think) if you follow the procedure I outlined in my previous commit.)

Hawk777:
I thought I would have to understand what the actions are in order to implement the turnoff semantics described [here](https://github.com/Jeija/minetest-mod-mesecons/issues/277#issuecomment-235527700). I was going for the rather invasive complete rework. I can just do [the minimally invasive change](#issuecomment-238842790) if you prefer. Probably easier to start there. I’ll do that when I get some time.

Jeija:
I think the [minimally invasive change](https://github.com/Jeija/minetest-mod-mesecons/pull/279#issuecomment-238842790) is the place to start. We can always expand on that with better algorithms later on. Changing these algorithms without really understanding all the ramifications a simple change may have can be hard and frustrating, so it might be better if I get more involved in actually writing the code for that part.

Hawk777:
How about this? Of course I can squash if you prefer one commit; I’ve left it as three until we finish discussion as it’s easier to see the progression of bits and pieces.

Jeija:
That looks very good. It is certainly a lot more reliable than the current system. I haven't found a single issue in 30 minutes of testing, but I will leave this here for some more time so that more people have a chance to try it out, and I will also do some more performance-related testing.

Performance-wise, I haven't really seen any big improvement yet (maybe this could somehow be optimized further??). But notably, I think this indicates that the real bottleneck isn't the map so much, but rather the inefficient algorithms used for turning off equipotential nodes. Now that these algorithms can be made a lot less complex (we don't need to re-schedule the state change in the actionqueue for unloaded areas), it should be much easier to make them more efficient. (And I've also learned more about searching of graphs since initially writing that code).

There are just two little things I would like to discuss concerning this code: You implement `mesecon.vm_abort`, but don't seem to use it anywhere. Then why implement it in the first place? Keeping unused functions around is just confusing if you ask me. Also, in `vm_swap_node` you assume that the VM for the area is already loaded in `vm_cache`. Currently, there is no situation where this could possibly not be the case, but I still think it would be better to extract the logic for loading a VM from `mesecon.vm_get_node` into a seperate function and make use of this function in both `vm_swap_node` and `vm_get_node`, even if that is just for clarity.

So all in all, thanky for the great work!

Hawk777:
Thanks for the comments. I’ve now extracted vm_get_or_create_entry, which finds (populating if needed) a cache entry, and used this function in vm_get_node and vm_swap_node.

As for why implement vm_abort, it’s there so it can be used in some future work to clean up the turnoff code. Right now, we look for any other active receptors on the equipotential node and then, if none are found, turn off all the conductors. Instead, we could just start turning off conductors and, if an active receptor is found, abort the transaction. Yes, this would mean using a separate transaction for each equipotential node attached to the turning-off receptor, so I’m not sure if that would be beneficial or not, but the option is there.

I’ve also added another commit that removes the turnon/turnoff logic that used to deal with deferring actions on unloadable areas, because there is no longer such a thing as an unloadable area.

Hawk777:
Rebased.

Jeija:
I merged this into the [voxelmanip branch](https://github.com/Jeija/minetest-mod-mesecons/tree/voxelmanip), simplified it a little more and wrote the code for the new turnon / turnoff functions on top of it. Turnoff has become **a lot** more performant, it is approximately twice as fast now (for long wires or many nodes). But even more than that, the new internal code is much easier to understand and a lot more maintainable now.

We still need a lot of testing though. Not only locally, but if possible also on servers (@VanessaE ??). Everything seems to be working fine for now, but I can't guarantee that. I guess I will have to consistently keep asking people in the IRC to test out the branch for the next 3 days or so before I will just merge it into master to see what happens. The release for the debian package will happen 1-2 weeks afterwards if no bug reports are made.

VanessaE:
I've pushed the voxelmanip branch to my daconcepts.com:30000 server (the only one I currently run that has mesecons in full).  It'll go into service tonight at 1am US eastern time.

VanessaE:
2016-08-31 01:24:57: ERROR[Main]: ...World/worldmods/dreambuilder_modpack/mesecons/legacy.lua:5: attempt to call global 'unhash_blockpos' (a nil value)
2016-08-31 01:24:57: ERROR[Main]: stack traceback:
2016-08-31 01:24:57: ERROR[Main]:       ...World/worldmods/dreambuilder_modpack/mesecons/legacy.lua:5: in main chunk
2016-08-31 01:24:57: ERROR[Main]:       [C]: in function 'dofile'
2016-08-31 01:24:57: ERROR[Main]:       ...e_World/worldmods/dreambuilder_modpack/mesecons/init.lua:143: in main chunk

Jeija:
Oh no, I should have spotted that, sorry. I just copied some code into a different file and apparently forgot one function. It didn't show up in any of my tests since the un-forceloading had already been done before for my maps. It should be fixed now (https://github.com/Jeija/minetest-mod-mesecons/commit/54efc64558ec21847c0b8d668bd206e88403a410).

VanessaE:
No worries, it happens.  I've pulled in and put it into service on the aforementioned server.  At least, it doesn't crash at startup. :smile:

PR #282, opened 2016-08-16
Remove an obsolete comment.

Hawk777:
I think these two comment lines are left over from when the LuaController did some string scanning to look for forbidden text. It no longer does this, relying entirely on the debug subsystem to prevent taking too much time.


Jeija:
Thank you! Documentation fixes are also always appreciated.

PR #284, opened 2016-08-23
Move conductors into their own mod

cheapie:
This allows the conductors (the actual mesecons themselves) to be disabled individually from the rest of the mod if desired. If they are disabled, other mesecons devices still work if placed touching each other (same as before).

This also moves some textures shared between mods to the base "mesecons" mod. This is somewhat unrelated, but allows LuaCs to work properly with just themselves and the base mod enabled, which was the entire goal of this. As somewhat of a side effect, this also resolves an issue where the LuaC was missing textures if the command block is disabled.


Jeija:
Manually committed in https://github.com/Jeija/minetest-mod-mesecons/commit/912f17f33590df4ad76775a3511881ad4bbf4853 and https://github.com/Jeija/minetest-mod-mesecons/commit/4816dee396203dba1df9328c5565c145c4f2751f with the following changes compared to your original commit:
* Renamed from `mesecons_conductors` to `mesecons_wires`. "Conductor" in mesecons is used as a more general term describing all kinds of conductors including mese blocks, insulated wires and the default wires.
* The wire documentation files should also be moved to `mesecons_wire`
* Actual seperation using autoconnect hooks instead of creating a dummy function in case wires aren't installed, see commit description and source for more details
* Adapted other nodes in the mod to stop using `update_autoconnect`.

Thanks for your contribution!

PR #285, opened 2016-09-03
Cache forceload_timeout setting

cheapie:
While I was experimenting with the profiler on a local copy of VanessaE's creative server (which is quite mesecons-heavy), I noticed that this globalstep function was taking quite a long time to run. This change brought it from about 20% of the globalstep time down to about 5%.

### DIFF
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -255,9 +255,10 @@ function mesecon.get_node_force(pos)
 end
 
 minetest.register_globalstep(function (dtime)
+	local timeout = mesecon.setting("forceload_timeout", 600)
 	for hash, time in pairs(mesecon.forceloaded_blocks) do
 		-- unload forceloaded blocks after 10 minutes without usage
-		if (time > mesecon.setting("forceload_timeout", 600)) then
+		if (time > timeout) then
 			minetest.forceload_free_block(unhash_blockpos(hash))
 			mesecon.forceloaded_blocks[hash] = nil
 		else


Jeija:
Unfortunately I can't merge this PR since forceloading was about to (and is now) replaced with VoxelManip caching which should greatly improve performance and reliability. Your discovery is very interesting though!

Have you also noticed the same lag for other settings? Things like overheating ("overheat_max") and player detectors ("detector_radius") are also executed quite often and they also use `mesecon.setting`. Do you think it would be beneficial to cache all settings right in the `mesecon.setting` function? Do you happen to know what exactly takes so long about reading a setting?

cheapie:
I didn't look into a whole lot of what mesecon.setting was doing, but from what I did see, it's calling minetest.setting_get (or similar, depending on the setting type). I'm assuming that that's where most of the lag is coming from, but I haven't done any real testing of that by itself to find out.

Meanwhile, your/Hawk777's changes sound promising. I'll test those on the same map I was using before, as well as getting some profiling information to see how it compares.

I'll also try to get around to trying out caching those other settings, and seeing if they have any real impact on anything.

cheapie:
OK, I've tested the current state of mesecons (with the recent changes), and all of my circuits that I tried are still working normally.

However, performance is still not the greatest. The overhead of the globalstep function checking the forceloaded blocks is gone, of course, but the actionqueue globalstep is now at 37% of the globalstep time on my test world (same local copy of VE-C as before, standing in the same spot).

While this probably isn't the greatest and could still use some work, it's certainly a lot better than some of the other mods used there (technic, mostly).

t4im:
> what exactly takes so long about reading a setting?

Reading that setting once needs about 6 global lookups + 2 table lookups, compared to 1+1 if you'd use the minetest api directly. You could test if localizing those helps improve speed noticeable.

Btw, @cheapie  you should not take the relative times of the profiler. I actually removed them in the followup Profiler PR https://github.com/minetest/minetest/pull/4343, because they were misleading. Better take the absolute times in µs, they are independent of other instruments.

Jeija:
As stated earlier, this is really necessary anymore, closing.

PR #299, opened 2016-12-04
Fix crash with movestones

yyt16384:


yyt16384:
By the way, I'm not sure about the intended behavior here. Should it check rules with `x=-1, z=0` instead of `x=0, z=-1` for node `{x=pos.x+1, y=pos.y, z=pos.z}`?

Jeija:
Thank you very much, I didn't even know this problem existed.

Actually the whole `is_power_on` stuff is from the very early days of mesecons (and is admittedly a mess), it should be replaced with just normal `action_on` / `action_off` functions and different inputs for movestones, that should also make the code much simpler. So there is no real intended behaviour here, since the whole `is_power_on` etc. concept isn't compatible with how mesecons works internally these days anymore. But as it appears, it works just well enough for most people...

PR #301, opened 2016-12-23
Allow Luacontrollers to reprogram themselves

cheapie:
This allows Luacontrollers to call program(), which replaces the LuaC's code with a supplied string. This should, in theory, be safe, as it's run the same way and therefore subjected to the same checks as user-entered code. It behaves exactly as if a user had typed in the code and clicked "Execute", including sending it the "program" event.

The purpose of this is to allow LuaC programs to update themselves, in conjunction with a "modem" that I have in development for a different mod, a program sent from another Luacontroller, user input, or other means.

The corresponding documentation is in mesecons.net PR #12.
(and no, GitHub, that's not the right link...)

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -479,7 +479,8 @@ local function reset_meta(pos, code, errmsg)
 	meta:set_string("formspec", "size[10,8]"..
 		"background[-0.2,-0.25;10.4,8.75;jeija_luac_background.png]"..
 		"textarea[0.2,0.6;10.2,5;code;;"..code.."]"..
-		"image_button[3.75,6;2.5,1;jeija_luac_runbutton.png;program;]"..
+		"field[2.75,5.5;5,1;channel;Remote Update Channel;${channel}]"..
+		"image_button[3.75,6.25;2.5,1;jeija_luac_runbutton.png;program;]"..
 		"image_button_exit[9.72,-0.25;0.425,0.4;jeija_close_window.png;exit;]"..
 		"label[0.1,5;"..errmsg.."]")
 	meta:set_int("heat", 0)
@@ -525,7 +526,13 @@ local digiline = {
 	receptor = {},
 	effector = {
 		action = function(pos, node, channel, msg)
-			run(pos, {type = "digiline", channel = channel, msg = msg})
+			local meta = minetest.get_meta(pos)
+			local setchan = meta:get_string("channel")
+			if channel == setchan then
+				mesecon.program_luac(pos,msg)
+			else
+				run(pos, {type = "digiline", channel = channel, msg = msg})
+			end
 		end
 	}
 }
@@ -537,6 +544,10 @@ local function on_receive_fields(pos, form_name, fields, sender)
 	if minetest.is_protected(pos, name) and not minetest.check_player_privs(name, {protection_bypass=true}) then
 		minetest.record_protection_violation(pos, name)
 		return
+	end
+	if fields.channel then
+		local meta = minetest.get_meta(pos)
+		meta:set_string("channel",fields.channel)
 	end
 	mesecon.program_luac(pos,fields.code)
 end
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -280,6 +280,12 @@ local function get_digiline_send(pos)
 end
 
 
+function mesecon.program_luac()
+	-- This is a placeholder that will be redefined later.
+	-- It needs to be global so that the redefinition will be "noticed" by the function below.
+end
+
+
 local safe_globals = {
 	"assert", "error", "ipairs", "next", "pairs", "select",
 	"tonumber", "tostring", "type", "unpack", "_VERSION"
@@ -303,6 +309,7 @@ local function create_environment(pos, mem, event)
 		print = safe_print,
 		interrupt = get_interrupt(pos),
 		digiline_send = get_digiline_send(pos),
+		pos = pos,
 		string = {
 			byte = string.byte,
 			char = string.char,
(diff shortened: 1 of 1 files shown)


sofar:
FPGA-like behavior? fantastic!

cheapie:
If by "FPGA-like" you mean like dynamic reconfiguration, that's pretty much what this is (although on a software level, not hardware).

Jeija:
I generally like the idea, but I first need to make sure this doesn't introduce any new flaws (like possible DoS vulnerabilities etc.).

Also, does it make sense to completely overwrite the code in the luacontroller itself? If, say, you wanted to reprogram the luacontroller multiple times (e.g. have remote updating capabilities using digilines), you would have to include the code needed for reprogramming inside the parameter to `program()`. So I'm wondering if maybe there is a better way such as having a second , volatile code storage location that can be written to using `program()`.
This could even be as simple as just prepending the string contents of the volatile storage to the actual code in the controller. That way, the volatile code could contain functions that can be called from the actual luacontroller code.
Not saying it has to be done that way, just a suggestions that may make using this feature easier for players.

cheapie:
Being able to store functions into "mem" would accomplish that, but from what I've seen that doesn't appear to be possible.

Also, seeing as how the code you're loading onto the LuaC is likely a later version of what it's running at the time, wouldn't it contain the update code already, unless you manually deleted it?

Jeija:
The basic idea here is to make the code transmitted over digilines as minimal as possible, because transmitting the actual "payload" code you want to execute as well as the updater code can be quite difficult to debug.

Also, if you happen to introduce a bug in your code which makes it not run on the controller anymore, you would have to manually fix that bug on every Luacontroller before flashing new code. While this cannot be prevented when just appending the new code as a string (as in my proposal above), it is less likely since you would propably encapsulate your code in a function.

Another (better imo) idea would be to not even have a `program()` function, but rather have a designated digiline channel (the channel should be configured by the user in an input field) for reprogramming. Whenever a string is received on that channel, it is loaded into the luacontroller code box. That way, you cannot "brick" your luacontroller by updating it and you circumvent the weird situation where you have to include the `program()` statement inside the `program()` statement if you want to keep updating. What do you think about this idea?

cheapie:
The reason that I don't like the digiline channel idea for that is because I want to have my traffic light controllers update over the internet using a "modem" that I'm working on in another mod. The modem isn't "smart" enough to grab the code from the returned data (which also includes the HTTP response code and some other stuff), and that sort of intelligence doesn't really belong in the modem anyway.

Jeija:
Okay, but that is just one very specific use case. You could also just put another LuaController for parsing the HTTP response between the modem and all the other LuaControllers and you are done. I'd argue that this architecture is actually simpler and more maintainable than having LuaControllers that parse HTTP responses and update themselves where the update includes code to parse HTTP responses and update themselves. I don't seem to understand the downside of the digiline channel approach? (Maybe apart from the fact that you would need a "modem-driver" luacontroller in your scenario).

Also, if the goal of this change is to have luacontrollers update over the internet, this is also something I'd like to see as stated [here](https://github.com/minetest-mods/homedecor_modpack/pull/317#issuecomment-241247348). So if you'd like to integrate remote updating functionality right into luacontrollers, I'd welcome a feature of that kind (if you're interested in doing that, please let us discuss the objectives and the implementation in advance).

cheapie:
Maybe I could modify this to have both the program() function and the digiline channel idea? Since I've split writing code to a LuaC out into its own function, that would be easy enough to implement.

Jeija:
Yes, that sounds reasonable to me 👍 . You would also have to add a digiline programming channel input field to the luacontroller GUI and update the documentation accordingly. I'd be willing to merge a solution of that kind.

cheapie:
OK, I'll get started on that and update this PR when I have it done.

cheapie:
What should I call the field for the channel? Something like "Remote Update Channel" seems a bit long.

cheapie:
OK, that has been implemented, and both this PR and the documentation PR have been updated.

cheapie:
That most recent commit was per VanessaE's request (more or less), to prevent someone from replacing the code in a LuaC that doesn't have remote updating turned on by using the following code:
digiline_send("","evil code goes here")
Remote updates are now simply ignored if the channel is empty.

mesecons_luacontroller/init.lua
```diff
@@ -280,6 +280,12 @@ local function get_digiline_send(pos)
 end
 
 
+function mesecon.program_luac()
```
Jeija: Please don't use placeholders if you can somehow prevent it. Could you try to arrange the functions in an order so that this is not necessary?

mesecons_luacontroller/init.lua
```diff
@@ -359,6 +366,10 @@ local function create_environment(pos, mem, event)
 			time = os.time,
 			datetable = safe_date,
 		},
+		program = function(code)
```
Jeija: Please don't implement functions inside `env = { ... }`, this definition is long enough as it stands. Please put this line below `digiline_send = get_digiline_send(pos)` and implement this in a similar fashion as `get_digiline_send(pos)`.

mesecons_luacontroller/init.lua
```diff
@@ -303,6 +309,7 @@ local function create_environment(pos, mem, event)
 		print = safe_print,
 		interrupt = get_interrupt(pos),
 		digiline_send = get_digiline_send(pos),
+		pos = pos,
```
Jeija: The luacontroller code shouldn't know the controller's position.

Jeija:
This is something minor that concerns all of your pull requests: Could you please make sure to use a space after a comma (`something("foo", "bar")` not `something("foo","bar")`?

Jeija:
This also needs to impose some sort of code size restriction so that players cannot abuse the code memory for storage or try to crash / lag the server by writing lots of code (like lots of auto-generated print statements). A simple check to make sure code size is like below 200kB or something like that should suffice. 200kB ought to be enough for anybody, even the LTC-4000 doesn't use that much.

cheapie:
Since I've taken a different approach to solving my original problem, I'm not really all that interested in working on this PR any more. I will still keep my code around for a while in case someone wants to use it as a starting point for another attempt at this.

Issue #292, opened 2016-11-20
Pistons are not rotated correctly by screwdriver


Wuzzy2:
You can use the screwdriver from Minetest Game to screw up active pistons. You can rotate the piston arm and the piston itself seperately, which looks odd.
Minetest Game has API support for special screwdriver handling.

One possible bugfix would be to disable rotation for the piston arm completely and add special rotation code for the piston itself, if its arm is out (trying to rotate both piston and piston arm, if possible).

Jeija:
I don't think adding special code for rotating the piston is worth the effort. I'd say it is acceptable to just completely disable rotation in this case.

MahzunGozler:
I encounter the issue too and notice something :
When you move an extended piston with a screwdriver, you move only the piston and not the extended part indeed. Then if you break the piston, the extended arm will remain placed and become unbreakable, you can just rotate it with the screwdriver. World Edit was used to remove it.

I propose a simple (I think) bugfix : give the possibility to rotate a piston ONLY if the arm is retracted.

PR #302, opened 2016-12-26
Rotation fixes

### LINKED ISSUE #297: Mesecons can be rotated
You can use the screwdriver to rotate the mesecons themselves. I am pretty sure this is unintentional.

The Minetest Game API allows to handle (and prevent) screwdriver events, see `game_api.txt`.

cheapie:
First of all, this prevents wires from being rotated, which would fix #297.

This also overhauls the piston rotation system completely, making it use 6D facedir instead of separate up/down nodes. This allows pistons to be rotated (only when off, fixing #292) with the screwdriver without the "pusher" coming out of the wrong side/facing the wrong way/whatever. LBMs are also provided to convert pistons to the new system automatically.

### DIFF
--- a/mesecons/depends.txt
+++ b/mesecons/depends.txt
@@ -1,1 +1,2 @@
 default
+screwdriver?
--- a/mesecons/services.lua
+++ b/mesecons/services.lua
@@ -93,3 +93,11 @@ mesecon.queue:add_function("cooldown", function (pos)
 		meta:set_int("heat", heat - 1)
 	end
 end)
+
+-- "Shim" for simple rotation, will result in the following item in nodedefs using it:
+-- "on_rotate = screwdriver.rotate_simple" if screwdriver mod is installed
+-- "on_rotate = nil" (essentially not present) if screwdriver mod is missing
+
+if screwdriver then
+	mesecon.rotate_simple = screwdriver.rotate_simple
+end
--- a/mesecons_extrawires/corner.lua
+++ b/mesecons_extrawires/corner.lua
@@ -37,7 +37,7 @@ minetest.register_node("mesecons_extrawires:corner_on", {
 	sunlight_propagates = true,
 	selection_box = corner_selectionbox,
 	node_box = corner_nodebox,
-	on_rotate = screwdriver.rotate_simple,
+	on_rotate = mesecon.rotate_simple,
 	groups = {dig_immediate = 3, not_in_creative_inventory = 1},
 	drop = "mesecons_extrawires:corner_off",
 	mesecons = {conductor =
@@ -65,7 +65,7 @@ minetest.register_node("mesecons_extrawires:corner_off", {
 	sunlight_propagates = true,
 	selection_box = corner_selectionbox,
 	node_box = corner_nodebox,
-	on_rotate = screwdriver.rotate_simple,
+	on_rotate = mesecon.rotate_simple,
 	groups = {dig_immediate = 3},
 	mesecons = {conductor =
 	{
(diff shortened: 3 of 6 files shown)


mesecons/util.lua
```diff
@@ -179,6 +179,10 @@ end
 function mesecon.register_node(name, spec_common, spec_off, spec_on)
 	spec_common.drop = spec_common.drop or name .. "_off"
 	spec_common.__mesecon_basename = name
+	if spec_common.mesecon_wire then
```
Jeija: This does not belong here - please put this in `mesecons_wires/init.lua` (and you might also want to add `on_rotate = false` to the nodes in `mesecons_extrawires`).

mesecons_pistons/init.lua
```diff
@@ -765,3 +341,86 @@ minetest.register_craft({
 		{"mesecons_pistons:piston_normal_off"},
 	}
 })
+
+-- LBMs to convert old pistons to use facedir instead of separate up/down nodes
```
Jeija: Please put this in `mesecons/legacy.lua` so that we will remember to remove this at some point later (unless there is an important reason to put it here i missed?).

cheapie:
OK, I think these changes should address your concerns.

Jeija:
Something else: I'd prefer to not depend on the screwdriver mod, so that e.g. server admins who don't want people messing with their blocks can disable that mod. However, if we add `on_rotate = screwdriver.*`, we should propably add `screwdriver` to the dependencies (or am I missing something?).
Is there any non-ugly way to address this concern? We could copy over `rotate_simple`, but that also doesn't seem like a very good idea...

MT-Modder:
The best, least ugly, way would be this I think:
```
if not screwdriver_mod then
    mesecon.rotate = copied code
else
    mesecon.rotate = screwdriver.*
end

...
on_rotate = mesecon.rotate
...
```

Or something like that....

Jeija:
Or even better: We can add an optional dependency to `screwdriver` in depends.txt (I didn't know this was possible since I've been sort of out of touch with minetest development lately) and if the screwdriver mod doesn't get loaded (so it's missing from the system), we can just set `on_rotate = false` since `on_rotate` won't be used anyways.

MT-Modder:
True. I think you'll want ```on_rotate = return end,``` instead but I could be wrong.

cheapie:
I know I took forever, but that should be fixed now. I more or less implemented Jeija's way, but just let it default to nil if the screwdriver mod isn't present.

MT-Modder:
Nice 👍

Jeija:
Looks good at a first glance 👍 

I would like to properly test this and review the code given that this refactors quite a lot. So this PR might take a little longer than usual, but I'd much rather make sure everything works fine than merge this ASAP. Any help with testing is greatly appreciated, please report your results, no matter if they're positive or negative!

Jeija:
I only noticed now that I wanted to merge this that with this PR applied it seems to be no longer possible to place up / down pistons just by placing them underneath / above you. I'd say it's inconvenient for players to be required to use screwdrivers for placing up / down pistons, so I'd like to get that fixed.

I'm not sure if this is something that can be solved just by changing the facedir handling. It should certainly be possible to do something like we did before, where we had a function `piston_orientate` that determines the piston's direction:
```Lua
local pitch = placer:get_look_pitch() * (180 / math.pi) 
if pitch > 55 then <do something to get down piston> end
if pitch < -55 then <do something to get up piston> end
```

Also, the [piston-rotation-rebased](https://github.com/Jeija/minetest-mod-mesecons/tree/piston-rotation-rebased) branch should contain this PR rebased to the current master branch.

numberZero:
Obsolete.

PR #303, opened 2016-12-26
Allow water turbines to work with river water (fixes #294)

### LINKED ISSUE #294: Water turbine does not work with flowing river water
The water turbine does not go on if flowing river water is going through.

IMO the water turbine should accept all “flowing” nodes as long they are in the group “water”. Or is it intentional that only the “normal” water is accepted?

cheapie:
This allows for additional types of water to be somewhat more easily specified, and adds river water to the list.

### DIFF
--- a/mesecons_hydroturbine/init.lua
+++ b/mesecons_hydroturbine/init.lua
@@ -53,7 +53,12 @@ minetest.register_node("mesecons_hydroturbine:hydro_turbine_on", {
 })
 
 
-local water_nodes = {["default:water_flowing"]=true,["default:river_water_flowing"]=true}
+local function is_flowing_water(pos)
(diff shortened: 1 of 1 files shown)


Jeija:
Unfortunately, this is still hardcoded. I'd actually prefer a solution like this:
A liquid activates the turbine, if it is a member of the `water` group and has `liquidtype == "flowing"`. That way, other mods can also add their own types of liquids that may activate a turbine. For instance, a mod might want to add "pond water" or "glacier water".

cheapie:
This has now been modified as you requested.

Jeija:
Thank you for the contribution! I rebased the two commits into one and merged this manually.

PR #304, opened 2016-12-26
Properly handle rotation of buttons/levers

### LINKED ISSUE #298: Buttons and wall levers can be rotated around axis, which causes weirdness
You can use the screwdriver from Minetest Game to rotate the buttons and wall levers around their axis.

I think it is OK to horizontally rotate these things as the mod still works just fine. But if you start rotating these things around their axis, and then push the button or use the lever, odd things happen. The signal does not go where I expect it to go.
I think it would be best to disallow axis rotation (see Minetest Game API).

cheapie:
This fixes #298 by registering new "receivers" that point up/down (made to look like vertical mesecons). They are used whenever a button or "wall lever" (they're not just for walls any more :) is rotated such that it is mounted to the floor or ceiling. You can also have them upside-down on the wall and all that fun stuff, and the receivers are still right-side up for that.

As an added (and somewhat beneficial) side effect, floor-mounted buttons and levers work with vertical mesecons as well, in the same way as pressure plates.

### DIFF
--- a/mesecons_receiver/init.lua
+++ b/mesecons_receiver/init.lua
@@ -28,16 +28,8 @@ local receiver_get_rules = function (node)
 	return rules
 end
 
-minetest.register_node("mesecons_receiver:receiver_on", {
+mesecon.register_node("mesecons_receiver:receiver", {
 	drawtype = "nodebox",
-	tiles = {
-		"receiver_top_on.png",
-		"receiver_bottom_on.png",
-		"receiver_lr_on.png",
-		"receiver_lr_on.png",
-		"receiver_fb_on.png",
-		"receiver_fb_on.png",
-	},
 	paramtype = "light",
 	paramtype2 = "facedir",
 	sunlight_propagates = true,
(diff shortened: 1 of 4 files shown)


mesecons_receiver/init.lua
```diff
@@ -81,30 +92,153 @@ minetest.register_node("mesecons_receiver:receiver_off", {
 	}}
 })
 
+minetest.register_node("mesecons_receiver:receiver_up_on", {
```
Jeija: I like the idea of up / down wall levers. However, could you please try to implement the registration of normal wall levers, up wall levers and down wall levers with `mesecon.register_node` (https://github.com/Jeija/minetest-mod-mesecons/blob/master/mesecons/util.lua#L195), most other nodes use that already. Using `mesecon.register_node` you will only have to specify the differences between onstate and offstate nodes, which makes the code a lot shorter and more readable. And the way I see it, onstate and offstate receivers seem to have almost everything in common.

cheapie:
The receiver registration has now been changed to use mesecon.register_node().

Jeija:
This is looking very good now, just some small requests:
* Is it possible to make it so that when placing a button / lever, you can now also place it on the ground or at the ceiling? That would expose this new functionality to more users. If this isn't possible without breaking backward compatibility (or doing something like adding LBMs), then it's propably not worth it.
* I would suggest adding some sort of visual cue to the up / down receiver so that the player knows it is not just a vertical wire. E.g. have a little bump in the middle or remove the edges like this: 
mesecons_receiver_off. If you have a better idea, let me know.
* Tiny style fixes: In `mesecon.receiver_get_pos_from_rcpt`, please write `rules = mesecon.*`, not `rules=mesecon.*`. In the same file, add spaces after commas.

cheapie:
OK, it now has a better nodebox, and I also fixed all of the style issues I could find.

I wasn't able to do anything about the automatic wall/ceiling mounting, as it appears I would have to change the way those nodes handle facedir and add an LBM.

Jeija:
I merged this manually with the following changes:
* Few minor style fixes
* Add `on_rotate = false` to all receiver types while we're at it
* Change `mesecon.buttonlike_onrotate(pos, node, _, _, newparam2)` to `mesecon.buttonlike_onrotate(pos, node)` since `newparam2` is unused anyways

Thanks for your contribution 👍 !

PR #306, opened 2017-01-01
make insulated crossing wire to mesh

DS-Minetest:
with nodebox:
- more textures are needed
- (the code is more)
- there are holes in the insulating
- the upper wire makes a shadow on the other one if only one of both is enabled
- and more textures are needed

so, I've made a mesh:
screenshot_20170101_180931

I tried to make it as much as possible equal to the nodebox.


Jeija:
I really like this 👍 

Two more questions:
* You just added another commit to this PR. Could you please summarize why you changed the UV map and what exactly?
* I don't know much about modelling, but can you import .b3d files into blender without losing information compared to the .blend file? If so, we don't need the .blend file, otherwise I'd prefer to include it for future reference.

DS-Minetest:
I made some uv-map fixes, i turned the top and the down sides with 180° for each themselves, I'm still not sure if this will make problems with other textures. Also I disabled the vertex normals exporting on export, this halved the size. (I still have to make some tests.)
I don't have an importer for .b3d files and don't even know if you can even import it. Minetest has only an exporter. So, I will include it with another commit?

Jeija:
Yes, please also add the .blend file then. I'd suggest creating an additional `src` directory inside `mesecons_extrawires` to put that. Just tell me when you have finished testing and the model is ready for merging.

DS-Minetest:
OK, I think, it's ready to merge.
What I did in last commit:
I made a colorgrid to see if every side is turned right and corrected the wrong ones. With this I also saw that I don't need to test more. Also I included the .blend file in src dictionary.

Jeija:
I rebased your commits into one and merged them manually as 7415036f5b3e75e182bb3818b7e68718022319c1.

Thanks for your help!

Issue #307, opened 2017-01-03
add support for doc modpack?


DS-Minetest:
I've already started to add some support to Wuzzy's doc modpack and I recognized, it's very easy. (There are already doc folders in the mods.)
So, should I continue and open a pull request?

Jeija:
I have to disagree as long as the documentation system is not standardized. I'd rather not add some specific code for one particular other mod. How can I be sure that e.g. minetest_game don't add their own system in the future and Wuzzy's mod will become obsolete?

I will leave this open for discussion, but from my perspective I'd say there are propably better ways to improve mesecons (e.g. code cleanups, improve documentation text on website, ...) than to put much effort into this.

DS-Minetest:
ok

Wuzzy2:
:-(

I don't know what you mean by “standardized”.

But I do know this: Just because something is in Minetest Game it does not mean it's a “Standard”. It's only a “Standard” *within* Minetest Game, while other subgames can be wildly different.
The Minetest community really should get over its obsession with Minetest Game as if it would be the Lord himself and players must not have other subgames beside him. xD
There are subgames which are pretty much different and have completely different APIs or don't even have any API, so any mod which is in Minetest Game can hardly be seen as a “standard”.
Minetest Game doing its own Help system (rather than just adopting this modpack) would be an *extreme* case of Not Invented Here and I'll probably oppose it. I also don't have any reason to believe this scenario will happen in the near future.

If you have comments about the Help modpack in general, or about the Documentation System, I'd be glad to read your opinions in the forum thread. :-)

The Documentation System is currently the only mod which makes it possible to add any help pages. I have not seen any attempt to replace it so far. I also have not seen a criticism on the general concept of this modpack. But the mod is still pretty young.

Improving documentation texts on website is not a replacement to providing in-game help texts. IMO in-game help is better than help scattered on webpages. I am not opposed to web help at all, I just don't think it should be limited to this. In-game help is way more convenient than forcing the player to “tab out” of Minetest all the time. That's why I put so much effort into the Help modpack in the first place. Or imagine a newbie finding Mesecons the first time on a server. In this case, it is unlikely the newbie will actively search the web for Mesecons help. The chance is even lower if the server is full of difficult-to-understand mods, even if they all have web help. Now the newbie is forced to hunt down the help for each of the mods. Even if the newbie knows about `/mods`, this gets tedious quickly, especially since many mods don't have any help at all.

[...]

PR #308, opened 2017-01-06
support help modpack

DS-Minetest:
Here is a PR about what I did. The doc texts are mainly copied from the doc folders in the mods.

### DIFF
--- a/mesecons_blinkyplant/depends.txt
+++ b/mesecons_blinkyplant/depends.txt
@@ -1,1 +1,3 @@
 mesecons
+doc?
+doc_items?
--- a/mesecons_blinkyplant/init.lua
+++ b/mesecons_blinkyplant/init.lua
@@ -36,12 +36,23 @@ mesecon.register_node("mesecons_blinkyplant:blinky_plant", {
 },{
 	tiles = {"jeija_blinky_plant_off.png"},
 	groups = {dig_immediate=3},
-	mesecons = {receptor = { state = mesecon.state.off }}
+	mesecons = {receptor = { state = mesecon.state.off }},
+	-- doc support:
+	_doc_items_longdesc = "The blinky plants toggles between on and off state every three seconds."..
+		" Can be used to make clocks. Also works after having restarted the game.",
+	_doc_items_usagehelp = "Rightclick the blinky plants to en/disable it."
 },{
 	tiles = {"jeija_blinky_plant_on.png"},
 	groups = {dig_immediate=3, not_in_creative_inventory=1},
-	mesecons = {receptor = { state = mesecon.state.on }}
+	mesecons = {receptor = { state = mesecon.state.on }},
+	-- doc support:
+	_doc_items_create_entry = false
 })
+
+-- doc support:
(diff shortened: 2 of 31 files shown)


Wuzzy2:
Thanks for your time. I know it's not going to be merged at this moment, but I give my thoughts on this for the future:

Pretty much all texts need to be written from scratch for the in-game help. The help texts are not self-contained. If the user e.g. uses the lookup tool on any random mesecons item, not knowing about the other ones, the help text will usually just add to the confusion. I'm not saying that all help pages should explain the mesecons mod from scratch, but they should mention mesecons and their *relation* to other blocks. But this clearly.

When Jeija starts to approve this Help modpack (and I find some time), I might as well write some of the help texts myself. Well, at least as good as I can (I don't know all of the internals, so …). I'd rather not spend actual code work into something which I know is not going to be accepted.

The help texts must be written in a style where you can take any item completely out of context and it could still be understood. Most items interact with other items. The mesecons torch is the perfect example: If you look at this text without any context, it will be hard to understand anything at all.

For clarity, the word “mesecons” should be used on pretty much all help pages. Terms like “effector” should all be qualified to become e.g. “mesecon effector”. So the player is tempted to also look at the related help entries.
The entry on the mesecon item itself needs extra-care, as this is the most important item of the entire mod. It should be explained well and in detail. But putting the entire Mesecons explanation into the wire entry would be overkill, I guess.

The mese block text is very bad, as it destroy the original help text and even refers to external documentation, making the in-game help basically pointless. This is going to be tricky as there's already a help text from `doc_minetest_game` for the mese block. :/

Later, Mesecons could also benefit from adding some factoids:
First, piston stoppers. I think this is nice to know.
Second, whether a block is a mesecon receptor, effector or conductor. (Interestingly, some of the blocks are actually multiple of those, like ghoststones.)
Third, if it is a conductor, mention which directions are conducting. VERY important IMO.

Help for pistons don't seems to be broken completely.

Probably the most challenging block to explain would be the Lua controller. This thing can not be explained in just one entry alone. Although I want to note there is no limit on the text length.
The help entry on the actual item could just be a brief explanation on its purpose and maybe a very quick command reference or something like that.
Thankfully, the formspec already has a link to the relevant web page.

We might also conider adding a special entry or even a category only for the Mesecons mod itself. Not for the items themselves, but for a general introduction, and maybe some hints and simple examples. I am not sure how to implement this in detail, so I think we should first stick with item help for now.

PS: I honestly did not know that you can toggle the blinky plant with rightclick. Thanks for that! You learn something new every day.

DS-Minetest:
Note, that I didn't finish it and the descriptions could mostly be found in the doc folders in the mods, I just copied them.
Making an extra category for mesecons would be needed.
I used `minetest.override_item("mesecons:wire_"..nodeid.."_off", {_doc_items_create_entry = false})` because I thought overriding an item which was registered only a little before would not make problems and I didn't see that there's something for off- and onstate.

DS-Minetest:
I'll close this now, it makes no sense holding this open and also I used the master branch so, urgh.
Anyone who wants to have this just ask me (eg. via pm on forum) for it.
#307 will stay open.

PR #312, opened 2017-01-09
Add protection testing to the Piston and Movestones.

### LINKED ISSUE #281: Pistons move nodes into protected areas
Tested with Minetest 0.4.14 vanilla with Landrush and mesecons both git-cloned

After i defined a protected area with Landrush (it should work with other protection mod) i was able, with a user different from the protected area owner, to push nodes inside it to move its contents. 

Even if the contents stopped at the protected area limit, the protected area has contents moved and external nodes pushed inside, this may lead to griefs.

shacknetisp:
This adds protection checking to pistons and movestones. If a protected node is encountered when pushing, the action will fail. If it is encountered while pulling, only the stack up to that point will be pulled. Nodes placed before this change was implemented will continue to act as before.

Fixes #281

### DIFF
--- a/mesecons_movestones/init.lua
+++ b/mesecons_movestones/init.lua
@@ -73,7 +73,7 @@ function mesecon.register_movestone(name, def, is_sticky)
 		-- ### Step 1: Push nodes in front ###
 		local maxpush = mesecon.setting("movestone_max_push", 50)
 		local maxpull = mesecon.setting("movestone_max_pull", 50)
-		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, maxpush)
+		local success, stack, oldstack = mesecon.mvps_push(pos, frontpos, direction, maxpush)
 		if success then
 			mesecon.mvps_process_stack(stack)
 			mesecon.mvps_move_objects(frontpos, direction, oldstack)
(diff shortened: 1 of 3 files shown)


cheapie:
+1 to the idea, haven't actually tested it yet.

jhcole:
This makes pistons and movestones unusable in protected areas even by the area owner.

numberZero:
1. `local meta = minetest.get_meta(pos)`
2. Don’t use `to_table` unless you really want the whole table. And remember that it can return an empty table and not nil if there is no meta (what it actually does in latest dev).
3. tl;dr Does it check the node it pushes into?

shacknetisp:
> Don’t use to_table unless you really want the whole table. And remember that it can return an empty table and not nil if there is no meta (what it actually does in latest dev).
What is the method to determine if a node has metadata?
> tl;dr Does it check the node it pushes into?
Are you saying it doesn't? Or are you saying you aren't sure if it does or not?

numberZero:
@shacknetisp 

> What is the method to determine if a node has metadata?

None AFAIK, but you shouldn’t either. The simplest solution is to check `owner` only.

The problem arises, though, if you want to distinguish old (thus ownerless) pistons which you don’t want to break and new ownerless (for whatever reason) pistons that should not be able to move protected nodes. Some possible solutions:
* Use additional field like `new_style` or `ownerless`
* Use reserved owner name (maybe containing LF or even NUL so that it can’t be a valid player name; not sure can meta store NUL, though) for ownerless movestones.
* Forbid placing ownerless pistons completely.

> Or are you saying you aren't sure if it does or not?

Yes, I’m not sure, haven’t tested. Don’t even remember have you fixed that on your server.

shacknetisp:
I would not respect old pistons at all (they'd behave like an empty owner), but the discussion from #281 seemed to indicate that backwards compatibility would be required. I think I'll just go with setting the meta field `uses_owner` to 1 on node placement; reserved names seem a bit messier to me, and forbidding ownerless pistons could cause some trouble when other node-placing mods want to place a piston (this final solution may actually work best, but I'd need some agreement on the potential breakage).

shacknetisp:
I will commit without backwards-compatibility, the potential to exploit a protection-ignoring piston in a world where all other pistons and movestones obey protection is somewhat large. If there is a better alternative, I'd be happy to work on that, just let me know what it is.

cheapie:
I think this should probably be configurable, so that it doesn't break a ton of existing machines.

sofar:
Needs rebase, and probably a better solution -> backwards compat should preferably be kept, especially for mods as complex as this.

PR #313, opened 2017-01-15
Comment out tablecopy_stripfunctions

Thomas--S:
Because of enormous performance impact.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -280,7 +280,9 @@ local function get_digiline_send(pos)
 
 		-- No sending functions over the wire and make sure serialized version
 		-- of the data is not insanely long to prevent DoS-like attacks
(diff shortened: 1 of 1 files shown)


Thomas--S:
Not only the LTC-4000 times out, also the code for displaying stuff at the streets matrix screens.

cheapie:
The issue here, by the way, is that streets 2 has a 16x16 matrix display, and sending the table to control it is causing "code timed out!", even with nothing else being done on the same event.

Jeija:
Given that this issue seems to impact several people, I have decided to completely revert the function stripping addition. I have deleted that code in https://github.com/Jeija/minetest-mod-mesecons/commit/54daee236e680fa79b2dd31e812f7e063887a0bf, since the `mesecon.tablecopy_stripfunctions` function isn't used anywhere else.

Thomas--S:
Thank you very, very much!

PR #315, opened 2017-01-19
FPGAs

sfan5:
Sometimes you want this
screenshot_20170119_163347
...but more compact.

Since it's just some logic gates, LuaControllers are way too much (and inefficient). Microcontrollers are deprecated and more complicated to program.

Clearly Mesecons is missing something, introducing **Programmable Logic Gates**:
screenshot_20170119_163604

Similar to FPGAs they are configured by chaining some logic gates together.
There are 4 pins that can be used for both input or output (exactly like the µC).
To create more complex circuits ten "registers" named from 0 to 9 are available to hold intermediate results.
The available logic gate types are *AND*, *OR*, *NOT* and *XOR*.

Here is how the UI looks with the same circuit as displayed above configured (also showing the capability to warn the user about inconsistencies/errors in the circuit configuration):
bildschirmfoto_2017-01-19_16-38-15

Lastly, the craft recipe is a little simpler than the Lua- / Microcontroller. Programmable logic gates are supposed to partly replace microcontrollers when you don't need the flexibility of luacontrollers.


MT-Modder:
👍 Very nice!

numberZero:
Nice idea, but the implementation has flaws IMO.
If i understand the code correctly, it effectively runs a program like:
```
reg1 = portA and portB
reg3 = portA and portC
...
portD = reg9 or reg8
```
Such program is simpler to write, really. Moreover, you can copy-paste a program, but not a gate configuration. And, well, this is not a gate configuration anyway, really, as gates work in different order.

So, I can suggest:
1. Make destination fixed for each row,
2. Process the whole data flow tree on programming (making sure that it is actually a tree, i.e. has no cycles) for all 4 outputs.
3. Store the resulting program (as Lua code, like 'port.a = pin.a and (pin.b or pin.c)') in metadata (tree can easily be expanded to such program).

That would make it work more like gates, at least. (And would simplify configuration changes too: if you decided e.g. to negate some input port, you would not need to move all existing configuration one line down to make negation be applied earlier).

Also it would be nice to have some programming tool to make identical gate configurations on several controllers.

sfan5:
@numberZero 
> If i understand the code correctly, it effectively runs a program like [...]
> Such program is simpler to write, really. Moreover, you can copy-paste a program, but not a gate configuration. And, well, this is not a gate configuration anyway, really, as gates work in different order.

Indeed it is, however the point of this is to avoid writing a Lua program.
Reasons why I am avoiding lua controllers:
* Functions have to be re-parsed and compiled everytime
* Disabled JIT (due to sandboxing)
* Lots of heavy checking and sandboxing for problems that won't occurr anyway with such simple scripts

> 1. Make destination fixed for each row,

This would complicate configuration changes (something you suggested to make easier actually).

> 2. Process the whole data flow tree on programming (making sure that it is actually a tree, i.e. has no cycles) for all 4 outputs.

I initially wanted to do it like this, however writing code to do the reordering was more complicated than not writing it and I decided for the simpler way.

> 3. Store the resulting program (as Lua code, like 'port.a = pin.a and (pin.b or pin.c)') in metadata (tree can easily be expanded to such program).

Actually I have already written code to compile circuit configs to Lua code.

However at least one problem arises when using this: Lua functions would need to be re-parsed and compiled everytime when being run.
To really take advantage of JIT and avoid this problem they could be cached, however caching needs to depend on their position and changes to the world need to be accounted for too.

> (And would simplify configuration changes too: if you decided e.g. to negate some input port, you would not need to move all existing configuration one line down to make negation be applied earlier).
> Also it would be nice to have some programming tool to make identical gate configurations on several controllers.

I agree, however this is not a priority.

> BTW, LuaCtls are very cheap themselves... #316

That's not the point, I coded these because I wanted better tools for creative building.
Not because I thought getting LuaControllers in survival was too hard.

numberZero:
@sfan5 Just imagine you made a 10-gate configuration... and noticed that you forgot to add NOT gate on an input. What would you do?

numberZero:
@sfan5
> avoid writing a Lua program

I don’t mean a *Lua* program, it could be an extremely restricted language, allowing only several basic operations, like `0=a&b; 1=c&d; a=0|1; b=0^1`
But that’s still *sequenced*, unlike circuits.

numberZero:
@sfan5 never mind. It seems we just want different things. I will probably make another pull request or a separate mod.

sfan5:
>> avoid writing a Lua program
> 
> I don’t mean a *Lua* program, it could be an extremely restricted language, allowing only several basic > operations, like `0=a&b; 1=c&d; a=0|1; b=0^1`
> But that’s still *sequenced*, unlike circuits.

This is exactly what microcontrollers do, however those are unmaintained and deprectated.

Jeija:
I like the idea of this. It's propably not faster than writing code for luacontrollers, but many people use gates for the challenge and out of curiosity, so this certainly caters to that audience.

This is not a code review yet, just some of the things I discovered while testing this ingame:
* From what I understand, the programmable logic gate cannot be used for stateful logic circuits, like RS-Flops since that would result in a dependency cycle which results in an "undefined register" warning. That is obviously limiting in the kind of circuits you can build with this node. Is this intentional?
* Readings from output ports are wrong, they seem to always be 0. Once again, I'm not sure if this is intentional. Luacontrollers e.g. read pinstate 1 if the output port is set to true. If we allowed reading from output ports, this would make stateful circuits possible, since the state could be stored as an output value.
* Some more gate types like NAND, XNOR (=) and a simple buffer (e.g. to expose an internal register to an output port) would be useful
* I'd say we should just call it FPGA. I know an FPGA does more than this, but "Programmable Logic Gate" is kind of cumbersome
* I'd prefer the texture to look more visually distinct from the Luacontroller / Microcontroller. Maybe just make the texture grey [like these FPGAs](https://de.wikipedia.org/wiki/Field_Programmable_Gate_Array#/media/File:Altera_StratixIVGX_FPGA.jpg)?
* As @numberZero already mentioned, some way to copy-paste the configuration is essential. From what I see the gate configuration already gets serialized. It would be nice to either expose this directly using an input field or have some kind of inventory tool that can copy this data.

VanessaE:
It could also be called a GAL (Gate Array Logic for those who don't know the term), since the simpler ones like the classic GAL22V10 seem close to what this feature tries to accomplish.

sfan5:
> * From what I understand, the programmable logic gate cannot be used for stateful logic circuits, [...]. Is this intentional?

Yes that's intentional.

> * Readings from output ports are wrong, [...]
> * Some more gate types like NAND, XNOR (=) and a simple buffer (e.g. to expose an internal register to an output port) would be useful
> * I'd prefer the texture to look more visually distinct from the Luacontroller / Microcontroller. Maybe just make the texture grey like these FPGAs?

fixed :heavy_check_mark:

sfan5:
@Jeija would you prefer renaming to `GAL` (as suggested by VanessaE) or `FPGA`?

Besides a programming tool it's the only issue remaining and I'd not like this pull request to sit here for months unmerged..

Jeija:
I'd prefer FPGA, just because that is the more well-known term, even if it is slightly less correct. But I'm not going to insist on that name, if you want to have GAL, then I'd also be happy with that.

And yes, the open issues with this PR are:
* A copy-paste tool
* Documentation for this feature for http://mesecons.net. If you think this needs more than a short description + a picture, then just open a PR on https://github.com/Jeija/mesecons.net. Otherwise, just add documentation like all the other items have it and add the FPGA item to `documentation.json`.
* One small specific thing: You added the `print("[OK] mesecons > " .. minetest.get_current_modname())` line to this. I'd prefer to either not have that output or have it for all mods in the mesecons modpack, it's inconsistent otherwise.

I have seen you committing your own small fixes to the master branch in the past few weeks and I trust you to merge this PR yourself whenever you consider it to be finished (in that case let me know, so that I can update http://mesecons.net). If you prefer to have a second pair of eyes look over your code before commiting, write me a short notice and I will take a look. I have quickly glanced over the code before and haven't found any issues 👍 .

sfan5:
@Jeija mesecons.net can be updated

Jeija:
Thank you 👍 !

mesecons.net is now updated.

PR #321, opened 2017-02-12
Make "on" water turbine animated!

VanessaE:
This uses a fairly simple trick: the textures are animated, while the model is static and has a number of rotated copies of the parts that are seen to move (four copies of the blades, eight copies of the top/bottom
and their edges).  The opaque parts of the textures then decide which of those faces' copies are visible in each frame.


DS-Minetest:
This is great! :+1: 

But the rod in the middle should be animated, too.
screenshot_20170212_203715

(It's interesting, with sunlight and fancy lightning the turning top seems to blink.)

Jeija:
Amazing 👍 !

If you think @DS-Minetest's proposal to make the rod animated makes sense, then it might be a good idea to change that. I think it makes sense either way (the actual shaft that rotates could also be inside that rod in the middle). Also, feel free to merge this PR yourself, I'm happy to give you permission 😄 .

VanessaE:
@DS-Minetest  yep I know, I just got lazy with it. ;)

@Jeija Thanks!   I thought about the actual shaft being *inside* the visible part, as an explanation for said laziness.  Let's just go with that so that I don't have to redo the textures and add even more polys.  :smile:

DS-Minetest:
> Oh, also, update your engine build - you probably lack the new smooth lighting (which looks great on mesh nodes now).

I use the new smoth lightning.

numberZero:
Very nice idea, but it blinks. The reason is that
screenshot_20170213_200135
The blades don’t change the lightness gradually as they should. Are you sure your model has correct normals? (It blinks regardless of the smooth lighting setting)
Also, why do you use several mesh variants for the cap? That just makes the edge blink, rotating texture would be enough.

VanessaE:
I've pushed an update (to master branch) to fix some messed-up UV mapping.  Maybe this will fix your issue also?  I checked the normals before export just to make sure it was good.  Looks perfect to me now.

numberZero:
Sorry, it seems I have an older MT version (with buggy mesh shading). Rebuilding now.

numberZero:
Tested. Both old and new model work, but new model has better shading.
The blinking reason seems to be that MT skips animation frames at low FPS. That’s unavoidable. Although at lower speed (0.6/2.4) it looks better (it doesn’t need to be as fast as in this PR IMO).
Also, I still doesn’t understand why do you use the same trick for the caps? Rotating texture would be enough. And transparent polygons seem to affect FPS just like opaque ones.

VanessaE:
If by "caps" you mean the round top/bottom of the blades, actually I do use the same trick.  Since the texture has to rotate evenly to avoid weird artifacts, that meant making the top/bottom rotate around 90 degrees in 8 steps, after which the animation repeats but with the texture rotated (then again, and again, for four 8-frame sets).

The blades have four frames without the extra rotate-and-run-it-again cycling.  

In all cases, the frames are 11.25 degrees apart.

sofar:
Geesh, how many faces does this model have? Have you tried to 'limited dissolve" and clean it up?

VanessaE:
It has 407 faces, and unfortunately that's about as low as it can go and still provide a smooth-ish animation, at least until such time as the engine supports true node mesh animation anyway.

VanessaE:
Oh and I meant to say, @numberZero, I set the speed where I did to roughly match the speed at which water appears to flow.

PR #324, opened 2017-02-15
Fix active ghoststones drop as item when falling nodes fall into them

### LINKED ISSUE #289: Falling nodes can fall into active ghoststones and make them drop
The player can obtain the ACTIVE ghoststone node in the inventory with this trick:

* Place a ghost stone
* Activate it
* Drop a falling node (e.g. sand) on the spot where the active ghoststone is

Since the sand replaces the node, the *active* ghoststone will now drop as an item, which can be collected. You have to look closely to even recognize it.

My suggestion to fix this is simple:
* Change the drop of an active ghoststone to the normal ghoststone, so the player can't obtain weird items

Wuzzy2:
Fixes #289.

### DIFF
--- a/mesecons_random/init.lua
+++ b/mesecons_random/init.lua
@@ -53,6 +53,7 @@ minetest.register_node("mesecons_random:ghoststone_active", {
 	diggable = false,
 	sunlight_propagates = true,
 	paramtype = "light",
+	drop = "mesecons_random:ghoststone",
 	mesecons = {conductor = {
 		state = mesecon.state.on,
 		rules = {


DS-Minetest:
This doesn't really fix the issue totally.
The falling entity shouldn't become a node while it's in ghoststone. And when the ghoststone becomes inactive it should "press" the fall entity away.

Wuzzy2:
Any better ideas? At least this simple fix prevents the active ghoststone from being obtainable by the player,

Jeija:
Thank you 👍 , merged.
If someone thinks there is a better solution, feel free to provide some code. This is certainly better than what we had before!

PR #326, opened 2017-02-23
fix sticky movestone recipe

DS-Minetest:
You can gain two sticky out of one movestone, this makes not much sense.


Jeija:
Fair enough, that makes sense. Could you just please also make sure to update the [craft recipe documentation](https://github.com/Jeija/minetest-mod-mesecons/blob/master/mesecons_movestones/doc/movestone_sticky/recipe.png) for mesecons.net?

DS-Minetest:
Done.

DS-Minetest:
So, are there still any problems?

Jeija:
I squashed the commits and manually merged this. Thanks for your contribution and patience 👍

PR #327, opened 2017-03-04
Updated RECIPE for microcontroller to stop recipe collision

tanmayameher:
Due to collision of recipe of this microcontroller with mesecons Luacontroller , I have changed the recipe (not material, just the arrangement) and propose to merge into original. So, every one gets the benefit.

Thanks and Regards,
TM

### DIFF
--- a/mesecons_microcontroller/init.lua
+++ b/mesecons_microcontroller/init.lua
@@ -147,9 +147,9 @@ end
 minetest.register_craft({
 	output = 'craft "mesecons_microcontroller:microcontroller0000" 2',
 	recipe = {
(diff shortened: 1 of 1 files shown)


Jeija:
The microcontroller is not supposed to be craftable since it is a deprecated item that we don't want people to use. We only keep the microcontroller in mesecons for backwards compatibility reasons, but it will be removed eventually. So I'm sorry, but I'm not going to merge this PR.

tanmayameher:
oh its ok. it was there lying around. so,  i thought it might be a problem. :)
So, this new FGPA thing in new or ...

DS-Minetest:
> we don't want people to use

We could add group `not in creative inventory`.

Issue #323, opened 2017-02-15
Add more lightstone colors


Wuzzy2:
Any chance there will be more lightstone colors to match some of the missing Minetest Game dye colors? Would a PR have a chance?

* White
* Orange
* Pink
* Violet
* Cyan
* Magenta

(only bright colors have been chosen)

If not, I might release a separate mod. However, the lightstone API appears to be broken (I can't add lightstones from external mods).

VanessaE:
If you to the effort of adding more lightstone colors, you should consider adding [cheapie's RGBlightstone](https://github.com/cheapie/rgblightstone) mod and doing some creative coding to use it to supersede the existing colors.

numberZero:
cyan and magenta, at least, to match yellow. and probably orange and violet to make a rainbow and not just a color ring.
RGBlightstone is nice to have but sometimes you would prefer something simpler.

DS-Minetest:
- The RGBlightstone mod needs digilines, not only to be installed but also to be used by the players. Many players won't know enough about digilines to use it. (Btw. there's also the [colour_carrier](https://github.com/HybridDog/colour_carrier) mod.)
=> Having more different lightstones would be good.

- A contra point is that lightstones are not used for art but to see some different nodes activated and deactivated. Does anybody really want to use so many different lightstones?

Jeija:
I'd certainly agree to adding white, cyan and magenta so that our selection of colors makes more sense. Other colors depend on whether there is demand for it or not - I haven't really kept current with minetest mod development, but why couldn't someone use [Bobblocks](https://github.com/minetest-mods/BobBlocks) as a simple solution instead?

VanessaE:
Well if you're going to add more colors, you should switch them over to use hardware coloring.  Only one or two textures would be needed that way, and if you use param2 to colorize, then you only need one node and some creative code.

DS-Minetest:
> one node

It would be two nodes, on and off. Giving a node a bright color doesn't make it totally seem to be on.

VanessaE:
Correct.  That was a brain-o. :)

VanessaE:
@Jeija as for Bobblocks, actually that uses hardware coloring, so it would be a reasonable example of what I was talking about, though its textures don't lend themselves well to generic lightstones.

numberZero:
954262d0164692fa03fe48e9bd439c8e570726fb

PR #329, opened 2017-03-08
Add 6 new lightstone colors

Wuzzy2:
This PR adds 6 new lightstones: White, orange, pink, magenta, violet and cyan. Craftable with 3 dyes of the same color, 1 torch and 1 mesecon.

Screenshots:

New lightstones left, current lightstones right.

Colors intentionally left out: Dark green, black and brown. I have tried these colors but I figured they look very ugly; the green and (new) orange lightstone are just fine and don't need darker versions IMO.
Black doesn't feel right for something which glows and also doesn't look good.


Jeija:
I'm not sure about orange, pink and violet. Cyan, magenta and white are good for consistency and will be accepted [as I already mentioned](https://github.com/Jeija/minetest-mod-mesecons/issues/323#issuecomment-280574635). Everyone: Please comment with your thoughts on adding orange, pink and violet!

Also, please make sure to add documentation (3d views of the blocks, craft recipes, text) for mesecons.net as soon as it is decided which colors should be included.

Wuzzy2:
Why not orange? It's a secondary color. So is violet. If cyan (color between green and blue) is valid, then so is orange (between red and yellow) and violet (between red and blue).

I'm not happy about adding 6 near-identical pages, and I also don't know how to make lightstone images which don't look like shit.

VanessaE:
I don't like the idea only because it adds new textures and more nodes, when there's an engine feature that should be used instead.  At the very least, @Wuzzy2, you should be using the { name = "foo.png", color=bar } method to colorize the node, and just two base textures for all of them.

Jeija:
@Wuzzy2: No, since we're talking about additive color mixing here, we have [red, green and blue as primary colors and cyan, yellow and magenta in between](https://upload.wikimedia.org/wikipedia/commons/c/c2/AdditiveColor.svg). This is also the reason why I agreed to adding cyan, magenta and white in the first place, since yellow as a secondary color was already included without the other counterparts for balance.

@VanessaE: That sounds like a good idea 👍 . But we need to make sure to maintain backwards compatibility since `mesecon.lightstone_add` could potentially be used by other mods. So maybe we need a new lightstone-registration function for your proposal.

VanessaE:
It wouldn't require a new function.  You could simply hard-code the generic textures' filenames into the two register_node() calls and have the lightstone_add() function just discard the filenames given to it (that way the API doesn't change).

There might be a mod or two that looks for or uses the current lightstone textures for some reason, but those could easily be adapted (and such usage should be discouraged anyway).

VanessaE:
Hm, I forgot to account for passing along the actual color of the node...  So that's an extra field after the description.

If you wanted to avoid the need for the extra `nil` fields in the original calls, check the contents of the 'desc' field at the start of lightstone_add().  If that field is nil, assume the calling mod is passing a color name, dye, description, and color value, but no filenames.  E.g. four fields instead of five.

sofar:
Needs resolution (open for over a month, no comments).

I suggest merging as-is and resolving the paint/node coloring changes at a later point.

numberZero:
I like the idea, and while I don’t like the exact colors (orange is too brownish, pink is very strange, fuchsia could fit better than magenta), I don’t mind merging as-is.

*OT: my spellchecker accidentally switched from default `English (United States)` to `English (Hong Kong)`. How?!..*

Wuzzy2:
Still not resolved. :-(

numberZero:
Mergable, either as-is or with https://github.com/Wuzzy2/minetest-mod-mesecons/pull/1
screenshot_20170913_220814
screenshot_20170913_220802
@Wuzzy2

PR #330, opened 2017-03-13
Object and Node Detector Changes

joshuamars:
3 changes in one (sorry):
1. Support multiple player detection from the object detector
2. Adds a distance parameter for the node detector (# of spaces between node detector and node to detect)
3. Prevent players from editing the settings in either of the detectors, if they are in an owned area

### DIFF
--- a/mesecons_detector/init.lua
+++ b/mesecons_detector/init.lua
@@ -5,19 +5,33 @@ local GET_COMMAND = "GET"
 -- The radius can be specified in mesecons/settings.lua
 
 local function object_detector_make_formspec(pos)
-	minetest.get_meta(pos):set_string("formspec", "size[9,2.5]" ..
+	local meta = minetest.get_meta(pos)
+		meta:set_string("formspec", "size[9,2.5]" ..
 		"field[0.3,  0;9,2;scanname;Name of player to scan for (empty for any):;${scanname}]"..
 		"field[0.3,1.5;4,2;digiline_channel;Digiline Channel (optional):;${digiline_channel}]"..
 		"button_exit[7,0.75;2,3;;Save]")
+		if not meta:get_string("owner") then meta:set_string("owner", "") end
 end
 
-local function object_detector_on_receive_fields(pos, _, fields)
+local function after_place(pos, placer)
+	if placer then
+		minetest.get_meta(pos):set_string("owner", placer:get_player_name())
+	end
+end
+
+local function object_detector_on_receive_fields(pos, formname, fields, sender)
 	if not fields.scanname or not fields.digiline_channel then return end
 
 	local meta = minetest.get_meta(pos)
+	local owner = meta:get_string("owner")
+	if owner ~= "" and sender:get_player_name() ~= owner then
+		minetest.chat_send_player(sender:get_player_name(), "This detector is owned by "..owner.."!")
+		return
+	else
 	meta:set_string("scanname", fields.scanname)
 	meta:set_string("digiline_channel", fields.digiline_channel)
 	object_detector_make_formspec(pos)
+	end
 end
 
 -- returns true if player was found, false if not
@@ -28,16 +42,24 @@ local function object_detector_scan(pos)
 	if next(objs) == nil then return false end
 
 	local scanname = minetest.get_meta(pos):get_string("scanname")
+	local sep = "," -- Begin code to split scanname into array
+	local t={} ; i=1
+	for str in string.gmatch(scanname, "([^"..sep.."]+)") do 
+		t[i] = str:gsub("%s+", "")
+		i = i + 1
+	end -- end split code
 	local every_player = scanname == ""
 	for _, obj in pairs(objs) do
 		-- "" is returned if it is not a player; "" ~= nil; so only handle objects with foundname ~= ""
 		local foundname = obj:get_player_name()
 
 		if foundname ~= "" then
-			-- return true if scanning for any player or if specific playername was detected
-			if scanname == "" or foundname == scanname then
-				return true
+			for l = 1, i do
+				if (foundname == t[l] and foundname ~= "") or (foundname ~= "" and scanname == "") then
+					return true
+				end
 			end
+			-- return true if scanning for any player or if specific playername was detected
 		end
 	end
 
@@ -67,7 +89,8 @@ minetest.register_node("mesecons_detector:object_detector_off", {
 		state = mesecon.state.off,
 		rules = mesecon.rules.pplate
 	}},
-	on_construct = object_detector_make_formspec,
+	on_construct  = object_detector_make_formspec,
+	after_place_node = after_place,
 	on_receive_fields = object_detector_on_receive_fields,
 	sounds = default.node_sound_stone_defaults(),
 	digiline = object_detector_digiline
@@ -84,6 +107,7 @@ minetest.register_node("mesecons_detector:object_detector_on", {
 		rules = mesecon.rules.pplate
 	}},
 	on_construct = object_detector_make_formspec,
+	after_place_node = after_place,
 	on_receive_fields = object_detector_on_receive_fields,
 	sounds = default.node_sound_stone_defaults(),
 	digiline = object_detector_digiline
@@ -128,17 +152,26 @@ minetest.register_abm({
 -- Detects the node in front of it
 
 local function node_detector_make_formspec(pos)
-	minetest.get_meta(pos):set_string("formspec", "size[9,2.5]" ..
+	local meta = minetest.get_meta(pos)
+	meta:set_string("formspec", "size[9,2.5]" ..
 		"field[0.3,  0;9,2;scanname;Name of node to scan for (empty for any):;${scanname}]"..
-		"field[0.3,1.5;4,2;digiline_channel;Digiline Channel (optional):;${digiline_channel}]"..
+		"field[0.3,1.5;2,2;distance;Distance (0-10):;${distance}]"..
+		"field[2.5,1.5;4,2;digiline_channel;Digiline Channel (optional):;${digiline_channel}]"..
 		"button_exit[7,0.75;2,3;;Save]")
+	if not meta:get_string("owner") then meta:set_string("owner", "") end
 end
 
-local function node_detector_on_receive_fields(pos, _, fields)
-	if not fields.scanname or not fields.digiline_channel then return end
+local function node_detector_on_receive_fields(pos, fieldname, fields, sender)
+	if not fields.scanname or not fields.digiline_channel or not fields.distance then return end
 
 	local meta = minetest.get_meta(pos)
+	local owner = meta:get_string("owner")
+	if owner ~= "" and sender:get_player_name() ~= owner then
+		minetest.chat_send_player(sender:get_player_name(), "This detector is owned by "..owner.."!")
+		return
+	end
 	meta:set_string("scanname", fields.scanname)
+	meta:set_string("distance", fields.distance)
 	meta:set_string("digiline_channel", fields.digiline_channel)
 	node_detector_make_formspec(pos)
 end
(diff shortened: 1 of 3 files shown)


mesecons_detector/init.lua
```diff
@@ -5,19 +5,33 @@ local GET_COMMAND = "GET"
 -- The radius can be specified in mesecons/settings.lua
 
 local function object_detector_make_formspec(pos)
-	minetest.get_meta(pos):set_string("formspec", "size[9,2.5]" ..
+	local meta = minetest.get_meta(pos)
+		meta:set_string("formspec", "size[9,2.5]" ..
```
numberZero: wrong indentation

mesecons_detector/init.lua
```diff
@@ -5,19 +5,33 @@ local GET_COMMAND = "GET"
...
-local function object_detector_on_receive_fields(pos, _, fields)
+local function object_detector_on_receive_fields(pos, formname, fields, sender)
 	if not fields.scanname or not fields.digiline_channel then return end
 
 	local meta = minetest.get_meta(pos)
+	local owner = meta:get_string("owner")
+	if owner ~= "" and sender:get_player_name() ~= owner then
+		minetest.chat_send_player(sender:get_player_name(), "This detector is owned by "..owner.."!")
+		return
+	else
```
numberZero: `else` is not needed after `return`

mesecons_detector/init.lua
```diff
@@ -5,19 +5,33 @@ local GET_COMMAND = "GET"
...
+	end
 end
 
-local function object_detector_on_receive_fields(pos, _, fields)
+local function object_detector_on_receive_fields(pos, formname, fields, sender)
 	if not fields.scanname or not fields.digiline_channel then return end
 
 	local meta = minetest.get_meta(pos)
+	local owner = meta:get_string("owner")
+	if owner ~= "" and sender:get_player_name() ~= owner then
```
numberZero: why not `minetest.is_protected`?

mesecons_detector/init.lua
```diff
@@ -28,16 +42,24 @@ local function object_detector_scan(pos)
 	if next(objs) == nil then return false end
 
 	local scanname = minetest.get_meta(pos):get_string("scanname")
+	local sep = "," -- Begin code to split scanname into array
+	local t={} ; i=1
+	for str in string.gmatch(scanname, "([^"..sep.."]+)") do 
+		t[i] = str:gsub("%s+", "")
+		i = i + 1
+	end -- end split code
```
numberZero: use “set” and not “list”:
`t[playername] = true`
`if t[foundname] then...`

mesecons_detector/init.lua
```diff
@@ -28,16 +42,24 @@ local function object_detector_scan(pos)
 	if next(objs) == nil then return false end
 
 	local scanname = minetest.get_meta(pos):get_string("scanname")
+	local sep = "," -- Begin code to split scanname into array
+	local t={} ; i=1
+	for str in string.gmatch(scanname, "([^"..sep.."]+)") do 
+		t[i] = str:gsub("%s+", "")
```
numberZero: what if player name happens to contain spaces? (not sure is that possible, though)

mesecons_detector/init.lua
```diff
@@ -148,6 +181,14 @@ local function node_detector_scan(pos)
...
 	if not node then return end
 
+	local distance = minetest.get_meta(pos):get_string("distance")
+	if tonumber(distance) ~= nil then
+		distance = tonumber(distance)
+		if distance < 0 then distance = 0 end
+		if distance > 10 then distance = 10 end
+	else
+		distance = 0
+	end
```
numberZero: `get_int` returns 0 if that’s not an int, you would only need to clamp the value

mesecons_detector/init.lua
joshuamars: Not possible

mesecons_detector/init.lua
joshuamars: Fixed lol

mesecons_detector/init.lua
joshuamars: Right makes more sense

mesecons_detector/init.lua
joshuamars: Uhhhhhhhh.... what? What is get_int??

mesecons_detector/init.lua
joshuamars: Because it does not work.

mesecons_detector/init.lua
numberZero:
```lua
local distance = meta:get_int("distance")
if distance < 0 then distance = 0 end
if distance > 10 then distance = 10 end
```

mesecons_detector/init.lua
```diff
@@ -148,8 +178,16 @@ local function node_detector_scan(pos)
...
 	local frontname = minetest.get_node(
-		vector.subtract(pos, minetest.facedir_to_dir(node.param2))
+		vector.subtract(pos, vector.multiply(minetest.facedir_to_dir(node.param2), distance+1))
 	).name
 	local scanname = minetest.get_meta(pos):get_string("scanname")
```
numberZero: and use that `meta` here

mesecons_detector/init.lua
```diff
@@ -194,6 +232,11 @@ local function after_place_node_detector(pos, placer)
 	local node = minetest.get_node(pos)
 	node.param2 = minetest.dir_to_facedir(vector.subtract(pos, placer_pos), true)
 	minetest.set_node(pos, node)
+	
+	if placer then
+		minetest.get_meta(pos):set_string("owner", placer:get_player_name())
```
numberZero: and here, if you keep this

mesecons_detector/init.lua
numberZero: @joshuamars ? It *does* work if any protection mod is installed. Otherwise, one can simply break and place again that detector, becoming its owner, thus defeating the purpose of this patch.

joshuamars: That is the desired effect.  It is pointless to make it owned if all the wiring to and from it isn't.  So all the ownership does is prevent ppl from editing the meta data IF it is in an owned area.

mesecons_detector/init.lua
joshuamars: At the moment, it makes no sense b/c it is only used once

mesecons_detector/init.lua
joshuamars: I mean it is not hard to make the block fully owned, but it seems pointless ^^

mesecons_detector/init.lua
joshuamars: oh ok... I new that

mesecons_detector/init.lua
joshuamars: I guess this could go to a command block tho

joshuamars: Ok I used the can_dig function from the command blocks to make it fully owned.

numberZero: Er, I think we don’t want just another non-breakable node, locked chest is more than enough already. @Jeija your opinion is needed.

joshuamars: -.- so.... which is it? Should I undo what I just did?!?

joshuamars: Ok I redid this... the blocks are no longer "owned".  All they do is prevent players from changing their contents if it is in an owned area.

mesecons_detector/init.lua
```diff
@@ -28,17 +31,23 @@ local function object_detector_scan(pos)
...
 		-- "" is returned if it is not a player; "" ~= nil; so only handle objects with foundname ~= ""
 		local foundname = obj:get_player_name()
 
 		if foundname ~= "" then
-			-- return true if scanning for any player or if specific playername was detected
-			if scanname == "" or foundname == scanname then
-				return true
+			for _ in pairs(t) do
+				if t[foundname] then return true end
 			end
```
numberZero: It seems you forgot to remove the loop

mesecons_detector/init.lua
```diff
@@ -28,17 +31,23 @@ local function object_detector_scan(pos)
...
 		end
+		
+		if scanname == "" then return true end -- Return true if no player name was specified
```
numberZero: `if every_player ...`
or remove that variable as unused

numberZero:
Good =)
A few minor changes remain, I will do that myself now.

numberZero:
@joshuamars merge https://github.com/joshuamars/minetest-mod-mesecons/pull/2

numberZero:
@VanessaE please review

VanessaE:
I'll leave that to Jeija, I suck at code review :)

mesecons_detector/init.lua
```diff
@@ -28,14 +31,17 @@ local function object_detector_scan(pos)
 	if next(objs) == nil then return false end
 
 	local scanname = minetest.get_meta(pos):get_string("scanname")
+	local sep = "," -- Begin code to split scanname into array
+	local scan_for = {}
+	for str in string.gmatch(scanname, "([^"..sep.."]+)") do
```
Jeija: Can't we just use minetest's [string.split](http://dev.minetest.net/string.split) here? It is kind of hard to understand this way.

mesecons_detector/init.lua
```diff
@@ -148,10 +159,16 @@ local function node_detector_scan(pos)
 	local node = minetest.get_node_or_nil(pos)
 	if not node then return end
 
+	local meta = minetest.get_meta(pos)
+
+	local distance = meta:get_int("distance")
+	if distance < 0 then distance = 0 end
+	if distance > 10 then distance = 10 end
```
Jeija: It would be nice to make this maximum distance configurable, just use something like
```Lua
local distance_max = mesecon.setting("node_detector_distance_max", 10)
```

People can then change this value by setting `mesecon.node_detector_distance_max = something` in minetest.conf. [Here](https://github.com/Jeija/minetest-mod-mesecons/blob/master/mesecons/settings.lua) is the implementation of `mesecon.setting`.

Jeija:
Looking good 👍

Apart from what I mentioned in my code comments, I have to request some more minor changes:
* When sending the digiline `GET` command to the node detector, it should also respond with the node name at the specified distance. Basically, just update [this line](https://github.com/joshuamars/minetest-mod-mesecons/blob/6fcc332f60ff9538de00042f6647869ed09b0f43/mesecons_detector/init.lua#L186).
* Can we make it so that when placing a node detector the distance value 0 is pre-filled into the distance field?
* Please update [object detector documentation](https://github.com/Jeija/minetest-mod-mesecons/blob/master/mesecons_detector/doc/objectdetector/description.html) (add a note saying that you can also search for comma-separated lists of players where the detector gets activated if any of the names in the list are found) and [node detector documentation](https://github.com/Jeija/minetest-mod-mesecons/blob/master/mesecons_detector/doc/nodedetector/description.html) (document the distance feature).

@numberZero: Thank you for taking care of most of code review for me  🎉. I didn't comment before, but I generally do read all comments and I will step in if there is something I don't agree with.

@joshuamars: Having been on the "other side" of this too, I understand that this process of requesting many small changes instead of us fixing them ourselves is tedious, but it benefits both mesecons (we wouldn't have spotted everything if I had just taken the code and fixed everything I had noticed) and hopefully also yourself as a learning experience.

mesecons_detector/init.lua
joshuamars: string.split did not work.  This works fine, why change it?

Jeija: I want this changed, it is just unnecessarily complicated this way. `string.split` seems to work just fine for me:
```
print(string.split(string.gsub("test, foo, bar", " ", ""), ","))
  --> {"test", "foo", "bar"}
```
(the gsub function takes care of removing spaces)

joshuamars: Ok well it is not working.  string.split(string.gsub(scanname, " ", ""), ",") is returning nill.  Not to mention, I still need to loop through the resulting array to make it readable to the obj loop below

joshuamars: ok

joshuamars:
I don't know enough about digilines to do what your asking.

joshuamars:
Ok I think I got everything @Jeija 
Also, I still loop through the result of string.split and an reorganize it so it goes from {"Foo", "Bat"} to {Foo=true,Bat=true}

Jeija:
Thank you 👍! I squashed all your commits and merged this manually.

Issue #309, opened 2017-01-06
Controller overheating


numberZero:
When you move a controller (e.g. using a piston), it keeps its heat level, but sometimes it doens’t cool down afterwards. It seems that the `cooldown` functions doesn’t get executed for it due to position change. This way, the controllers accumulates heat until burn-up.

UPD: Something similar causes random overheating of controllers that use interrupts to run constantly: the heat value may not be reset sometimes. Probably the `cooldown` function is executed after mapblock unload sometimes.

@Jeija What’s about different overheating scheme? Something like this:
https://github.com/numberZero/digiline_routing/blob/master/overheating.lua
This doesn’t even need to access the metadata.
I will rewrite that for `mesecons` myself, if you agree.

numberZero:
Should be solved by #334

PR #334, opened 2017-03-24
Improve overheating

numberZero:
1. Removes the need of metadata for simple gates
2. Should solve #309
3. It’s more logical by itself.

@Jeija would you support this?

Note that I can implement continuous cool-down, if you want, (and that should eat less CPU), but that would be a bit more complicated and this implementation should be lighter than the current one already.


Jeija:
Looks good to me, I'd support this 👍 . I can't come up with any reason why this wouldn't work, but obviously this may introduce some new problems. Therefore I'm going to delay [an official release](https://github.com/Jeija/minetest-mod-mesecons/releases) until this has been tested for some time.

I've been looking at the code and quickly tested this, and haven't found any problems so far. Since you have been doing some great PR reviews and produce good quality code, I have now also given you permissions to push to this repository. So feel free to merge this yourself whenever you like 😄 (after squashing commits).

VanessaE:
Speaking of overheating, pistons need this as well - I was thinking mesecons could "overheat" one by swapping in an inert piston model that looks like it has a broken plunger.

numberZero:
@VanessaE Good idea) I think they should have a different `heat_max` value, though, so I need to add support for that.

@Jeija 
> given you permissions to push to this repository.

Thank you very much!

> I don't think continuous cooldown is necessary

That’s more about CPU time and not correctness. But if you consider current usage acceptable (haven’t profiled it myself), I won’t bother with that now.

numberZero:
@shacknetisp can you test this? It seems to work fine, and should reduce lag ;)

VanessaE:
I've been informed that under Dreambuilder, gates aren't properly overheating, but LuaC's do.  DB's copy of mesecons is currently up-to-date.

numberZero:
@VanessaE 
> overheating is unreliable

MT is unreliable, by nature. Timings vary a lot in an unpredictable manner. So a truly reliable system can’t even exist in a MT world.
I don’t want to say nothing may be done; but we need to decide how should things behave in such unstable environment. I could write a lot on that but don’t want to do that here and now.

> gates aren't properly overheating, but LuaC's do
> They overheat just fine with a fresh map

So the most probable reason is the mighty global lag :P Even short-circuited NOT-gate may live for a long time if the server is slow. But LuaCtl may overheat itself anyway as it may work faster (e.g. handling several interrupts or digiline messages in one globalstep. The latter are synchronous IIRC; if so, it can overheat in a single globalstep—easily). But this patch should not affect that considerably, does it?

numberZero:
Seems to work properly. But should there be a bug, it’s way easier to catch it after merge, you know...

PR #337, opened 2017-04-09
Fix crash when error()ing an invalid value

cheapie:


sofar:
:+1:

DS-Minetest:
Could you please use `tostring` instead of `dump`? It now adds extra `"`s and if it's a table, it would give new lines that don't fit into the formspec.

numberZero:
@DS-Minetest Er, the error message is unreadable anyway, so its formatting, basically, doesn’t matter. The crash must be prevented.

DS-Minetest:
Actually the error messages are readable and useful IMO and using `tostring` would also fix the bug.

numberZero:
564bdc2
As `error()` expects a message (that is, a string), handling of non-string values may be arbitrary as long as it works somehow.

PR #338, opened 2017-04-11
luacontroller: make formspec bigger

DS-Minetest:
This adds a much more efficient and convenient formspec.
<details>
<details>
<summary>screenshots</summary>

<details>
<summary>before:</summary>

screenshot_20170411_153421</details><details>
<summary>after:</summary>

screenshot_20170411_152251</details></details>(The thing on the right down edge is away.)
(The background is now in indicated mode.)

I thought of editing the "Micro" to "Lua".
</details>


VanessaE:
A bigger formspec is nice, but one problem that really irritates me with it is for one thing, you can't scroll with the mouse wheel (cursor keys work).  Secondly, the formspec insists on keeping the cursor at the bottom of the visible text area even if doing so means you can't see any of the code that follows it.  It should never force-scroll the text area.

These are both probably irrlicht's fault, but if workarounds could be found, it would be a major plus.

DS-Minetest:
@VanessaE I can't upgrade the textarea in lua without too much effort. We could cut the text in textarea to the hight of the textarea and save this separately when the scrollbar is used. Idk if this would be possible, at least it would be a challenge. Perhaps I'll try to do this in a new mod and port it to here if it works. For now the bigger formspec should be better than the small one.

sofar:
The large formspec is mostly a waste of screen space. I've never seen scripts larger than 5-10 lines, and those are huge. Largest I've built was maybe 12 lines total.

More efficient formspec space use: yes
Making it larger: nah.

numberZero:
@sofar my scripts (LuaCtl) are usually a bit longer. And as the formspec scrolling is, um, a bit weird, larger area means more scripts will fit it leaving no possibilities for incorrect scrolling. Although for anything more complex than a pulse generator I personally always use an external editor anyway.

DS-Minetest:
> The large formspec is mostly a waste of screen space.

Ehm, for what else do you need the screen space while you are looking into the formspec?
> Largest I've built was maybe 12 lines total.

How much do you write into one line?!
Very often I need  more than 30 lines.
Well, if someone only does something like `port.b = pin.a and pin.c` and such stuff, I can understand that not much space is needed. But if you use digilines and it's applications it can be much more.

(Since copying doesn't work mostly, I don't use an external editor.)

joshuamars:
Ditto about the more that 30 lines @DS-Minetest. I agree that there is a need to improve the editor.  I personally think that the suggested changes would be helpful.  Perfect? No, but a step in the right direction, and certainly nothing to lose by taking it.

cheapie:
I have some that are 1000+ lines, but I edit those with an external text editor and copy/paste in.

joshuamars:
Copy/paste doesn't work for everyone

DS-Minetest:
@cheapie Wow, 1000+ lines. O_o

Let's get back to topic, is anyone really against this PR to be merged?

DS-Minetest:
Conflict solved.

numberZero:
IMO that should be merged. That may be helpful (note that I have larger font, so very few lines fit in the current formspec). But I don’t want to merge stealthily. So @Jeija, what’s your opinion? Also, what are the rules for merging?

Jeija:
To be honest, I haven't been playing minetest / using mesecons much in the past few months / years apart from testing Pull Requests and fixing mesecons bugs. So I'm certainly not qualified to judge anything related to gameplay. Since most opinions here seem to be in favor of merging this, I'd say go for it, you have commit access for a reason.
I'm not sure though if there is some kind of voting procedure for these kinds of changes in the @minetest-mods community?

PR #342, opened 2017-05-12
Ignore Eclipse project files & directories

AntumDeluge:
Simplification for developers using the [Eclipse IDE](https://eclipse.org/ide/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 *~
+
+# Eclipse project files & directories
+.project
+.settings


sfan5:
People actually use eclipse to develop lua?

numberZero:
@sfan5 Well, there is some support for that. I tried to use it some time ago, and it seemed to work rather well, but as always with Eclipse, I just couldn’t configure it well enough, and Eclipse doesn’t forgive misconfiguration.
By the way, what’s the IDE you use for Lua?

numberZero:
On topic: I’m mostly neutral. Of course we can’t support each and every IDE, but the changes requested are minor, while Eclipse is a rather common product. On the other hand, `.git/info/exclude` may be used for that as well.

AntumDeluge:
Eclipse does not have good Lua support. Supposedly, the [LiClipse](https://marketplace.eclipse.org/content/liclipse) plugin has an editor that supports Lua, but it doesn't even appear to have syntax highlighting.

I did just install the [Lua Development Tools](https://marketplace.eclipse.org/content/lua-development-tools) plugin, & it *does* include a text editor with syntax highlighting. But other than that, I have not tested it & don't know what else it can do.

I just use Eclipse because it's what I use for my other projects that involve C++, Java, & Python. And it makes managing my Git repos easier (IMO).

AntumDeluge:
Decided to create a *.gitignore* in my home directory & added the global `core.excludesFile` in my *~/.gitconfig* so that globally I don't have to worry about ignoring files for individual projects.

```
$ git config --global --add core.excludesFile "~/gitignore"
$ cat ~/gitconfig
...
[core]
	editor = nano
	excludesFile = ~/.gitignore
...
```

PR #343, opened 2017-05-12
Add cornernote's 'craft_guide' mod as optional dependency:

AntumDeluge:
Ensures that craft recipes are registered in [cornernote's *craft_guide* mod](http://cornernote.github.io/minetest-craft_guide/).

Added to the following mods with craft recipes:
- blinkyplant
- button
- delayer
- detector
- extrawires
- fpga
- gates
- hydroturbine
- insulated
- lamp
- lightstone
- luacontroller
- materials
- microcontroller
- movestones
- noteblock
- pistons
- powerplant
- pressureplates
- random
- solarpanel
- switch
- torch
- walllever
- wires

### DIFF
--- a/mesecons_blinkyplant/depends.txt
+++ b/mesecons_blinkyplant/depends.txt
@@ -1,1 +1,2 @@
 mesecons
+craft_guide?
--- a/mesecons_button/depends.txt
+++ b/mesecons_button/depends.txt
@@ -1,2 +1,3 @@
 mesecons
 mesecons_receiver
+craft_guide?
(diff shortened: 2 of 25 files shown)


numberZero:
@AntumDeluge 
1. Why have you added it to depends of each and every mod? It should be enough to add it to `mesecons` only, as everything depends on it already (and that dependency is essential so won’t be removed; to be honest, except of `materials` which don’t use the API so may not depend on it).
2. https://github.com/cornernote/minetest-craft_guide/issues/20

AntumDeluge:
Thank you, I like the idea of getting list of registered crafts after server is loaded. The alternative [craftguide](https://github.com/minetest-mods/craftguide) mod, from the [Minetest Mods Team](https://github.com/minetest-mods), seems to do this.

PR #344, opened 2017-05-13
Replace deprecated methods:

AntumDeluge:
- 'setting_get' with 'settings:get'
- 'setting_getbool' with 'settings:get_bool'

May not want to merge this until after 0.4.16 release, as the new ['settings' object calls][l_settings] are not available in 0.4.15.

[l_settings]: https://github.com/minetest/minetest/blob/43d1f375d18a2fbc547a9b4f23d1354d645856ca/src/script/lua_api/l_settings.cpp#L267

### DIFF
--- a/mesecons/settings.lua
+++ b/mesecons/settings.lua
@@ -1,15 +1,15 @@
 -- SETTINGS
 function mesecon.setting(setting, default)
 	if type(default) == "boolean" then
-		local read = minetest.setting_getbool("mesecon."..setting)
+		local read = minetest.settings:get_bool("mesecon."..setting)
 		if read == nil then
 			return default
 		else
 			return read
 		end
 	elseif type(default) == "string" then
-		return minetest.setting_get("mesecon."..setting) or default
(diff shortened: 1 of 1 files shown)


DS-Minetest:
Btw. movement gravity should be saved in a local variable on load time.

sfan5:
Not sure how @Jeija plans to have Mesecons compatibility so I'll wait a little more before merging this.

Jeija:
Since there is no real reason to drop compatibility with minetest 0.4.15 other than getting rid of the deprecation warnings, I'd suggest either waiting at least ~5 months or until the next minetest version gets released before merging this, whichever comes first. This will give users a comfortable timeframe to update their minetest installations, which is especially critical for those installing minetest from their distribution's package manager (I mean not even Archlinux ships 0.4.16 right now, so it really wouldn't make any sense to merge this already).

sofar:
Needs a rebase, but we're well beyond the 5 months now! Let's start merging these type of changes asap.

AntumDeluge:
Rebased.

PR #346, opened 2017-05-19
add luacontroller to programmer and checks area protection

berengma:
use the programmer also to copy code from one luacontroller to an other
checks if you are the owner if luacontroller or fpga is in protected area
reading allowed from all, writing only if you are owner of it

### DIFF
--- a/mesecons_fpga/tool.lua
+++ b/mesecons_fpga/tool.lua
@@ -1,28 +1,53 @@
 return function(plg)
 
+local function reset_meta(pos, code, errmsg) 
+	local meta = minetest.get_meta(pos)
+	meta:set_string("code", code)
+	code = minetest.formspec_escape(code or "")
+	errmsg = minetest.formspec_escape(errmsg or "")
+	meta:set_string("formspec", "size[10,8]"..
+		"background[-0.2,-0.25;10.4,8.75;jeija_luac_background.png]"..
+		"textarea[0.2,0.6;10.2,5;code;;"..code.."]"..
+		"image_button[3.75,6;2.5,1;jeija_luac_runbutton.png;program;]"..
+		"image_button_exit[9.72,-0.25;0.425,0.4;jeija_close_window.png;exit;]"..
+		"label[0.1,5;"..errmsg.."]")
+	meta:set_int("heat", 0)
+	meta:set_int("luac_id", math.random(1, 65535))
+end
+
 
 minetest.register_tool("mesecons_fpga:programmer", {
-	description = "FPGA Programmer",
+	description = "Programmer for FPGA and luacontroller",
 	inventory_image = "jeija_fpga_programmer.png",
 	stack_max = 1,
 	on_place = function(itemstack, placer, pointed_thing)
 		if pointed_thing.type ~= "node" then
 			return itemstack
 		end
-
+		
 		local pos = pointed_thing.under
-		if minetest.get_node(pos).name:find("mesecons_fpga:fpga") ~= 1 then
-			return itemstack
-		end
-
-		local meta = minetest.get_meta(pos)
-		if meta:get_string("instr") == "//////////////" then
+		if minetest.get_node(pos).name:find("mesecons_fpga:fpga") == 1 then
+			local meta = minetest.get_meta(pos)
+			if meta:get_string("instr") == "//////////////" then
 			minetest.chat_send_player(placer:get_player_name(), "This FPGA is unprogrammed.")
 			return itemstack
+			end
+			itemstack:set_metadata(meta:get_string("instr"))
+			minetest.chat_send_player(placer:get_player_name(), "FPGA gate configuration was successfully copied!")
+			
+			return itemstack
+			
+		elseif minetest.get_node(pos).name:find("mesecons_luacontroller:luacontroller") == 1 then
+			local meta = minetest.get_meta(pos)
+			if meta:get_string("code") == "" then
+			  minetest.chat_send_player(placer:get_player_name(), "This luacontroller is unprogrammed.")
+			  return itemstack
+			end
+			itemstack:set_metadata(meta:get_string("code"))
+			minetest.chat_send_player(placer:get_player_name(), "luacontroller was successfully copied!")
+			return itemstack
 		end
-		itemstack:set_metadata(meta:get_string("instr"))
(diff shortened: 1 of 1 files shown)


numberZero:
Very nice idea, but the code doesn’t seem well.

1. You try to determine is the code stored for FPGA or LuaC. Instead, you should store that as well (or make different tools for FPGAs and LuaCs). `ItemStack:get_meta()` may be helpful (available since [minetest:f2aa2c6](https://github.com/minetest/minetest/tree/f2aa2c6a986dec47856c49ae5f54fbf3c688e027)), it returns object similar to [NodeMetaRef](http://dev.minetest.net/NodeMetaRef) but without inventory.
2. You add local `reset_meta` function; I suppose you copied it from the LuaC code (btw, it doesn’t use the `heat` meta value since 54b9eaf (#334)). If so, you should use that function instead. If it is local (almost sure it is)... well, I’d probably add it to the LuaC node definition, unsure.
3. You keep the tool in the FPGA mod, but now it is useful without FPGAs. Again, you should either make different tools for FPGAs and LuaCs and store the code in the corresponding mods, or make an all-in-one tool (as currently) and store it in mod of its own.

Also, is the `program` event generated when this tool works on a LuaC?

P.S. That’s very good that you check for protection, that’s often forgotten.
P.P.S. I feel that should work in a node breaker as well... Finally, programming automation...

mesecons_fpga/tool.lua
```diff
@@ -1,28 +1,53 @@
...
-	description = "FPGA Programmer",
+	description = "Programmer for FPGA and luacontroller",
 	inventory_image = "jeija_fpga_programmer.png",
 	stack_max = 1,
 	on_place = function(itemstack, placer, pointed_thing)
 		if pointed_thing.type ~= "node" then
 			return itemstack
 		end
-
+
```
DS-Minetest: No!

mesecons_fpga/tool.lua
```diff
@@ -1,28 +1,53 @@
...
-
+		
 		local pos = pointed_thing.under
-		if minetest.get_node(pos).name:find("mesecons_fpga:fpga") ~= 1 then
-			return itemstack
-		end
-
-		local meta = minetest.get_meta(pos)
-		if meta:get_string("instr") == "//////////////" then
+		if minetest.get_node(pos).name:find("mesecons_fpga:fpga") == 1 then
```
DS-Minetest: `minetest.get_node(pos).name:sub(1, 18) == "mesecons_fpga:fpga"`
I know, `find` was already used before, but this would be more efficient I think.

mesecons_fpga/tool.lua
```diff
@@ -1,28 +1,53 @@
...
+			itemstack:set_metadata(meta:get_string("instr"))
+			minetest.chat_send_player(placer:get_player_name(), "FPGA gate configuration was successfully copied!")
+			
+			return itemstack
+			
+		elseif minetest.get_node(pos).name:find("mesecons_luacontroller:luacontroller") == 1 then
+			local meta = minetest.get_meta(pos)
+			if meta:get_string("code") == "" then
+			  minetest.chat_send_player(placer:get_player_name(), "This luacontroller is unprogrammed.")
+			  return itemstack
```
DS-Minetest: Don't use space indent!

DS-Minetest:
I don't see the use for this. What's wrong with crtl+c and crtl+v?
> Finally, programming automation...

I don't think we need that.

mesecons_fpga/tool.lua
numberZero: String search is, generally, a wrong way. It could be better to add a property to the node definition, like `nodedef.mesecons.fpga = {}` (or `true`, but table is, generally, better as there may be properties to add)

numberZero:
I think there should be a separate programmer for LuaCs, so that it could be extended with LuaC-specific options.

>> Finally, programming automation...
>
> I don't think we need that.

@DS-Minetest But I do.

IRL, the only way to program a controller is to use some external device (maybe except of producing pre-programmed chips). In MT, that may be just a handy addition. (btw, copy-paste doesn’t work for everyone... not sure of MT→MT copying, though; for me, it all works well.) It may be used to store programs in-game, as your controllers may be 30km far from each other. It could setup controller memory, or just pass some data with the `program` event, to help to make controllers with the same program with different parameters. Oh well, there are many options...

mesecons_fpga/tool.lua
berengma: To 1: Original code was using string, so did I. That is why. It works pretty good, even though you are right. It is not the cleanest solution.

To 2: Yes the function was copied. All functions from LuaC are not global. That es why the program event is not called.

To 3: I do not like the idea having a tool for each device. Like in real life it is more convenient to have for all. I changed the description of the fpga programmer that should make it clear for players.

DS-Minetest: Actually I meant the two added useless indents. :P

berengma: And I meant @NumberZero :D 
I do not get how these comments work here

berengma:
copy and paste does not work everywhere. Specially on linux systems.
When you build big setups with lua controllers a tool which can copy
code from lua controllers is very handy. See on my server in /openlab for
examples like scroll text and realtime clock with display.
It makes it also easier for other players to copy code from public available
demo setups.

mesecons_fpga/tool.lua
```diff
@@ -1,28 +1,53 @@
...
+			end
+			itemstack:set_metadata(meta:get_string("code"))
```
DS-Minetest: Save `meta:get_string("code")` in a local variable.

numberZero:
To check is a node a LuaC, it has an `is_luacontroller` property in its node definition. But I would prefer using something like #348 (@DS-Minetest, are you fine with that?):
```lua
local def = minetest.registered_nodes[node.name]
if def.mesecons.luacontroller then
  def.mesecons.luacontroller.set_program(…
```
That way, you won’t need to copy-paste functions as well.

On using the same tool to program different devices: it’s really fine, as long as the tool is very simple, like this one. I might make my own, sophisticated, LuaC-only programmer for extraordinary needs, but in a separate mod.

On checking does it store a code for LuaC or FPGA... oh well, I’m not sure if we can use in-dev features like itemstack metadata...

numberZero:
Abandoned.

PR #348, opened 2017-05-20
Publish some LuaC API

numberZero:


mesecons_luacontroller/init.lua
```diff
@@ -522,6 +522,22 @@ local digiline = {
...
+
+local function set_program(pos, code)
+	reset(pos)
+	reset_meta(pos, code)
+	local err = run(pos, {type="program"})
+	if err then
+		print(err)
+		reset_meta(pos, code, err)
+	end
+end
```
numberZero: Should return whether the update was successful... or not?

DS-Minetest: Or return `err`. Or whether the update was successful... or not `, err`.

mesecons_luacontroller/init.lua
```diff
@@ -522,6 +522,23 @@ local digiline = {
...
+
+local function set_program(pos, code)
+	reset(pos)
+	reset_meta(pos, code)
+	local err = run(pos, {type="program"})
+	if err then
+		reset_meta(pos, code, err)
+		return false, err
+	end
+	return true
```
DS-Minetest: You could save one indention with
```lua
if not err then
	return true
end
```

DS-Minetest:
Btw does this close #346?

numberZero:
@DS-Minetest no ofc, but it makes #346 easier to do properly.

PR #350, opened 2017-06-07
Add 'settingtypes.txt'

AntumDeluge:
Add 'settingtypes.txt' file for configuring mod from client GUI.

### DIFF
--- a/settingtypes.txt
+++ b/settingtypes.txt
@@ -1,43 +1,43 @@
 [mesecons]
 
-mesecon.resumetime (Resume time) int 4
-mesecon.overheat_max (Device heat limit) int 20
-mesecon.cooldown_time (Device cooldown time) float 2.0
-mesecon.cooldown_granularity (Cooldown step length) float 0.5
+mesecon.resumetime (Startup delay) int 4 1 10
+mesecon.overheat_max (Device heat limit) int 20 1 100
+mesecon.cooldown_time (Device cooldown time) float 2.0 0.1 10.0
+mesecon.cooldown_granularity (Cooldown step length) float 0.5 0.0 1.0
 
 
 [mesecons_blinkyplant]
 
-mesecon.blinky_plant_interval (Plant blinking interval) int 3
+mesecon.blinky_plant_interval (Plant blinking interval) int 3 1 5
 
 
 [mesecons_detector]
 
-mesecon.detector_radius (Player detector radius) int 6
-mesecon.node_detector_distance_max (Node detector radius) int 10
+mesecon.detector_radius (Player detector scanning radius) int 6 3 16
+mesecon.node_detector_distance_max (Node detector distance limit) int 10 1 16
 
 
 [mesecons_luacontroller]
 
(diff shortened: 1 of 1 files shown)


settingtypes.txt
```diff
@@ -0,0 +1,43 @@
...
+
+[mesecons_blinkyplant]
+
+mesecon.blinky_plant_interval (Plant interval) int 3
+
+
+[mesecons_detector]
+
+mesecon.detector_radius (Detector radius) int 6
+mesecon.node_detector_distance_max (Detector max distance) int 10
```
numberZero: “Player detector” and “Node detector”, I suppose

settingtypes.txt
```diff
@@ -0,0 +1,43 @@
...
+
+
+[mesecons_luacontroller]
+
+mesecon.luacontroller_string_rep_max (Max) int 64000
+mesecon.luacontroller_digiline_maxlen (Digiline max length) int 50000
+mesecon.luacontroller_maxevents (Max events) int 10000
+mesecon.luacontroller_memsize (Memory size) int 100000
```
numberZero: - `string:rep` result length limit
- Digiline message size limit
- Events per ...?
- Add units (bytes (?))

settingtypes.txt
```diff
@@ -0,0 +1,43 @@
...
+
+[mesecons_pistons]
+
+mesecon.piston_max_push (Max push) int 15
+mesecon.piston_max_pull (Max pull) int 15
+
+
+[mesecons_pressureplates]
+
+mesecon.pplate_interval (Interval) float 0.1
```
numberZero: “Check interval”?

AntumDeluge:
Added some of the recommended changes by @numberZero. But "Max events" is still the same. Not sure what to put there. numberZero recommended something like "Events per ?".

numberZero:
@AntumDeluge er, only some my recommendations were literal descriptions. You should have read them, instead of copy-pasting blindly.

AntumDeluge:
I knew they were suggestions, I just didn't know what else to put there. Was waiting for more suggestions.

numberZero:
@AntumDeluge As you see, I changes some long names. Look at the settings list (in MT settings menu, not in editor, to make experiment cleaner) and ask yourself whether you understand what each option does just by reading its name. If so, names are probably good; otherwise, improvement may be needed.

P.S. Parentheses are illegal in setting names)

SmallJoker:
This PR is now open for a quite long time - how mergeable is it right now? All setting descriptions except `mesecon.resumetime` are clear to me. Some kind of delay in seconds?

settingtypes.txt
```diff
@@ -0,0 +1,43 @@
+[mesecons]
+
+mesecon.resumetime (Resume time) int 4
```
SmallJoker: Could be called `Startup delay time`. Minimal value 1s, max should be somewhere around 10s
The settings don't seem to be limited in the code, but let's at least define some limits here.

numberZero:
@SmallJoker is this OK now?

AntumDeluge:
Thank you for the help everyone. 😃

Issue #339, opened 2017-04-13
Luacontroller and Microcontroller have the same recipe


cheapie:
This makes one or the other impossible to craft.

DS-Minetest:
<details>
<summary>Suggestion:</summary>
Add two crafting recipes to craft the controller into the other one.
</details>

numberZero:
Or just remove microcontroller crafting recipe. It is deprecated, why is it still craftable?

C1ffisme:
@numberZero Because everyone related to Minetest, me included, has a little bit of a nostalgia problem. :stuck_out_tongue_closed_eyes:

DS-Minetest:
@C1ffisme True, me too.
But also it's a feature and so, why should it be removed? Also with mesecons (no more minetest-mods-mesecons) as modpack, the parts can be deactivated. AFAIK there's a mesecons_microcontroller mod, it can also be deactivated.
<details>
<summary>An additional suggestion:</summary>

Add to the description of microcontroller something like core.colorize("#FF0000", "\nDeprecated").</details>

C1ffisme:
@DS-Minetest The main reason why it should be removed is because it's old, and, AFAIK, nobody is going to maintain it. That brings up issues especially if Minetest 0.5.0 is going to "break mod compatibility". Plus, the lua controller uses a real language, and just for bonus points, the same language that the modding API is written in. If that's not a good way to introduce someone to programming, I dunno what is.

sofar:
One easy solution would be to make a crafting recipe out of the lua controller to make a microcontroller, and vice versa.

C1ffisme:
@sofar That's possible, and an easy solution, but what's the point of keeping the microcontroller anyway if the luacontroller is far superior? (and the microcontroller is more likely to break if nobody maintains the code)

benrob0329:
Microcontroller has been replaced by FGPA afaik

VanessaE:
Imho, the simplest solution is to delete the crafting recipe for the µC, and hide it from the creative inventory.   That way newer stuff is promoted and old maps still work.

Brackston:
PR 351 seems to need both the lua's and the uController's itint.lua files to be changed. My PR only changed the uController and leave the lua controller alone.
Please see and comment. Thank you. 
This change has been tested and is currently running on a public server
#353

PR #351, opened 2017-07-10
Changed Microcontroller and LuaController recipes

hdak1945git:
Since I've recently ran into the problem described in #339 I decided to take a crack at it. Only minor changes to recipes and an new recipe was added (so as not to waste any items).

---

This commit changes the recipes for the Microcontroller and LuaController.

The Microcontroller is slightly simplifyed, so it no longer masks the recipe
for the LuaController (which otherwise remains unchanged); now both objects
can be crafted freely.

**addendum** recipe for the microcontroller is now:

~~~
[silicon] [silicon] [mesecon]
[silicon] [mesecon] [mesecon]
[       ] [       ] [       ]
~~~

New: A new LuaController recipe, by using 3x Microcontroller parts in the top
row will also produce two LuaControllers.

**addendum**

~~~
[microcontroller] [microcontroller] [microcontroller]
[               ] [               ] [               ]
[               ] [               ] [               ]
~~~

See also: minetest-mods/mesecons#339

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -678,6 +678,15 @@ minetest.register_node(BASENAME .. "_burnt", {
 minetest.register_craft({
 	output = BASENAME.."0000 2",
 	recipe = {
+		{'mesecons_microcontroller:microcontroller0000', 'mesecons_microcontroller:microcontroller0000', 'mesecons_microcontroller:microcontroller0000'},
+		{'', '', ''},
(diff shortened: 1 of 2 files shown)


DS-Minetest:
I'd prefer not to have the 3 microcontrollers to 2 luacontrollers recipe
but 4 microcontrollers to 3 luacontrollers
and 3 luacontroller to 4 microcontrollers.

PR #353, opened 2017-08-14
Recipe change

Brackston:
Change recipe to allow players to build both the lua and micro controllers.
Currently the recipe is duplicated between two devices making it impossible for players to make the lua controller
This change replaces the second row of silicon with mesecon_conductors.

### DIFF
--- a/mesecons_microcontroller/init.lua
+++ b/mesecons_microcontroller/init.lua
@@ -148,7 +148,7 @@ minetest.register_craft({
 	output = 'craft "mesecons_microcontroller:microcontroller0000" 2',
 	recipe = {
 		{'mesecons_materials:silicon', 'mesecons_materials:silicon', 'group:mesecon_conductor_craftable'},
-		{'mesecons_materials:silicon', 'mesecons_materials:silicon', 'group:mesecon_conductor_craftable'},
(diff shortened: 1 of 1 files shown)


DS-Minetest:
Another version of #351.

Brackston:
True, except this PR only changes one file for the mesecon_microcontroller (init.lua) 
So if the device is removed there would be no need to update the lua controller as well. 
This is a simple fix that allows players to have the lua controller again. 
As it is now, players on survival servers can not craft the lua controller. 


Brackston 
----- Original Message -----

Billre:
Good idea.  I find that both lua controllers and microcontrollers have their uses - I mix both of them in my mesecon circuits.  Works great - but, as an admin, I'm not restricted by recipe conflicts.  I would suppose a better question to ask would be:  Do we want players to enjoy lua coding?  If so, then we need to resolve this conflict.  As this is a simple change that does that, I'm giving it a thumbs up.  Thanks Brackston.

DS-Minetest:
>True, except this PR only changes one file for the mesecon_microcontroller (init.lua)

Well, the conversation recipes can also be done at microcontroller's `init.lua`. But that doesn't matter much.

Cat5TV:
TPS would also like to upvote this. We have several players who have requested it.

PR #354, opened 2017-08-14
add conversation recipes for lua- and microcontroller

### LINKED ISSUE #339: Luacontroller and Microcontroller have the same recipe

DS-Minetest:
I made my own PR for this now.
Fixes #339, closes #353 and closes #351.
Look to "Files changed".

### DIFF
--- a/mesecons_microcontroller/init.lua
+++ b/mesecons_microcontroller/init.lua
@@ -144,14 +144,27 @@ end
 end
 end
 
-minetest.register_craft({
-	output = 'craft "mesecons_microcontroller:microcontroller0000" 2',
-	recipe = {
-		{'mesecons_materials:silicon', 'mesecons_materials:silicon', 'group:mesecon_conductor_craftable'},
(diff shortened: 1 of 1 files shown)


Brackston:
This appears to make the devices interchangeable. It that correct?

DS-Minetest:
Well, they have the same value based on their recipe, so why not make it interchangeable?
The microcontroller is only directly craftable if the luacontroller is deactivated. This solves the overlapping problem.

Brackston:
> The microcontroller is only directly craftable if the luacontroller is deactivated. This solves the overlapping problem.
What happens if both of the controllers are desired?

DS-Minetest:
>What happens if both of the controllers are desired?

Then it isn't directly craftable but indirectly via a luacontroller and the conversation recipe.

Brackston:
Thank you, DS-Minetest.

PR #355, opened 2017-08-22
fix unloaded active button staying on for ever

### LINKED ISSUE #325: button stays on if unloaded while active
If a button was unloaded while it was activated, it doesn't get deactivated again and stays activated.
To test this you can press a button and press escape (in singleplayer) and leave and then join again.

DS-Minetest:
Fixes #325.

### DIFF
--- a/mesecons_button/init.lua
+++ b/mesecons_button/init.lua
@@ -4,11 +4,11 @@
 
 mesecon.button_turnoff = function (pos)
 	local node = minetest.get_node(pos)
-	if node.name ~= "mesecons_button:button_on" then --has been dug
+	if node.name ~= "mesecons_button:button_on" then -- has been dug
 		return
 	end
-	minetest.swap_node(pos, {name = "mesecons_button:button_off", param2=node.param2})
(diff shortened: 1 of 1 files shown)


mesecons_button/init.lua
```diff
@@ -45,7 +46,7 @@ minetest.register_node("mesecons_button:button_off", {
 		minetest.swap_node(pos, {name = "mesecons_button:button_on", param2=node.param2})
 		mesecon.receptor_on(pos, mesecon.rules.buttonlike_get(node))
 		minetest.sound_play("mesecons_button_push", {pos=pos})
-		minetest.after(1, mesecon.button_turnoff, pos)
+		minetest.get_node_timer(pos):start(1)
```
numberZero: hard-coded value

DS-Minetest: Is there any reason to make it configurable?

numberZero: Don’t think so, just don’t like magic numbers. But that’s not really a problem, at least as long as such number is used in one place only ;)

PR #356, opened 2017-09-05
handle blasts

### LINKED ISSUE #291: Piston arm can by dropped by exploding it with a TNT
It is possible to make the piston arm drop as an item (and thus collectible):

* Use Minetest Game
* Place some pistons somewhere
* Activate them all so that the arm is out
* Place and ignite TNT nearby

If you're lucky, you will soon find some dropped piston arms lying around.

I think this bug could be easiest fixed if the drop of the piston arm is disabled (`drop = ""`).

DS-Minetest:
Before eg. wires did not update.
Also fixes #291.


mesecons_lightstone/init.lua
```diff
@@ -28,20 +28,22 @@ function mesecon.lightstone_add(name, base_item, texture_off, texture_on, desc)
...
-	light_source = default.LIGHT_MAX-2,
-	sounds = default.node_sound_stone_defaults(),
-	mesecons = {effector = {
-		rules = lightstone_rules,
-		action_off = function (pos, node)
-			minetest.swap_node(pos, {name = "mesecons_lightstone:lightstone_" .. name .. "_off", param2 = node.param2})
-		end,
-	}}
+		tiles = {texture_on},
+		groups = {cracky=2,not_in_creative_inventory=1, mesecon = 2},
```
numberZero: Inconsistent styling.
Either keep as is (including wrong indentation, to reduce diff stats), or fix it.

numberZero:
@DS-Minetest 
Rebase needed.

DS-Minetest:
Rebased.

PR #357, opened 2017-09-05
make removestone check nodes above for falling

### LINKED ISSUE #290: Falling nodes don't fall if ghoststone or removestone below get activated
If you place a falling node on a ghoststone or removestone, then activate either of them, the falling node floats in mid-air and doesn't fall.
Also, attached nodes at removestones or ghoststones stay attached when the main node is gone.

Hint: In Minetest, you can use the undocumented function `nodeupdate` to “remind” nodes that they should fall. However, this function will be removed at least in Minetest 0.4.16. The current dev version of Minetest has the properly documented functions `minetest.check_for_falling` and `minetest.check_single_for_falling`. I suggest to test your code with falling nodes and attached nodes at the ghoststone or removestone.

DS-Minetest:
Fixes a part of #290.

### DIFF
--- a/mesecons_random/init.lua
+++ b/mesecons_random/init.lua
@@ -10,6 +10,7 @@ minetest.register_node("mesecons_random:removestone", {
 		action_on = function (pos, node)
 			minetest.remove_node(pos)
 			mesecon.on_dignode(pos, node)
+			minetest.check_for_falling(vector.add(pos, vector.new(0, 1, 0)))
 		end
 	}}
 })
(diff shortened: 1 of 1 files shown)


numberZero:
LGTM.
Don’t forget of GitHub keyword handling; it treats “fixes #290” in a special way...

DS-Minetest:
Btw. the keyword isn't there anymore since long.

numberZero:
@DS-Minetest Notifications aren’t sent when a comment is edited, only when a new comment or new commit is added.

DS-Minetest:
I know, sorry for commenting the change so late. >_<

PR #358, opened 2017-09-05
make fpga rotatable

DS-Minetest:


mesecons_fpga/init.lua
```diff
@@ -116,6 +116,43 @@ plg.register_nodes({
 	after_dig_node = function(pos, node)
 		mesecon.receptor_off(pos, plg.rules[node.name])
 	end,
+	on_rotate = minetest.global_exists("screwdriver") and function(pos, node, user, mode)
```
sfan5: why not set this independently of whether the screwdriver mod exists?

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,43 @@ plg.register_nodes({
...
+			if user and user:is_player() then
+				minetest.chat_send_player(user:get_player_name(),
+						"FPGA has been rotated counter-clock-wise.")
+			end
+		end
+		local meta = minetest.get_meta(pos)
+		local instr = meta:get_string("instr")
+		local out = ""
+		for i = 1, #instr do
+			local char = instr:sub(i, i)
```
sfan5: no no no no no

you should deserialize the instruction string, modify the structure and re-serialize

mesecons_fpga/init.lua
DS-Minetest: ~~This is because `screwdriver.ROTATE_FACE` and `screwdriver.ROTATE_FACE`.~~
You're right, this isn't executed without screwdriver mod and so, `screwdriver` is never indexed.

DS-Minetest:
At current state nil is returned. Should the wrench wear off when turning a fpga?

sfan5:
it should if it does that normally

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,46 @@ plg.register_nodes({
...
+					elseif num < 1 then
+						num = 4
+					end
+					instr[i][op].port = abcd1[num]
+				end
+			end
+		end
+		local is = lcore.serialize(instr)
+		meta:set_string("instr", is)
+		plg.update_formspec(pos, is)
```
sfan5: you can save a deserialization operation by calling `update_formspec` with `instr` instead

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,47 @@ plg.register_nodes({
...
+						num = 1
+					elseif num < 1 then
+						num = 4
+					end
+					instr[i][op].port = abcd1[num]
+				end
+			end
+		end
+		local is = lcore.serialize(instr)
+		meta:set_string("instr", is)
```
sfan5: ...which also means you no longer need `is` to be a variable

mesecons_fpga/init.lua
```diff
@@ -116,42 +116,45 @@ plg.register_nodes({
...
 		for i = 1, #instr do
-			local char = instr:sub(i, i)
-			local num = abcd2[char]
-			if num then
-				num = num + dir
-				if num > 4 then
-					num = 1
-				elseif num < 1 then
-					num = 4
+			for k = 1, #ops do
```
numberZero: `for k, op in ipairs(ops) do`
or
`for _, op in ipairs(ops) do`
as the counter is never used

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,46 @@ plg.register_nodes({
...
+				local op = ops[k]
+				local o = instr[i][op]
+				if o and o.type == "io" then
+					local num = abcd2[o.port]
+					num = num + dir
+					if num > 4 then
+						num = 1
+					elseif num < 1 then
+						num = 4
+					end
```
numberZero: May be simplified using some ~ℤ₄ magic~ modulo arithmetic. Necessary?

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,46 @@ plg.register_nodes({
...
+			dir = -1
+			if user and user:is_player() then
+				minetest.chat_send_player(user:get_player_name(),
+						"FPGA ports have been rotated counter-clock-wise.")
+			end
+		end
+		local meta = minetest.get_meta(pos)
+		local instr = lcore.deserialize(meta:get_string("instr"))
+		for i = 1, #instr do
+			for k = 1, #ops do
```
numberZero: `for _, op in ipairs(ops) do`
That’s Lua, not JS or even C.

DS-Minetest: But afaik `ipairs` is slower.

DS-Minetest:
@numberZero I'm not sure if this is what you wanted…

mesecons_fpga/init.lua
sfan5: this is not a "hot" code path, performance does not matter much here

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,42 @@ plg.register_nodes({
...
+				local op = ops[k]
+				local o = instr[i][op]
+				if o and o.type == "io" then
+					local num = abcd2[o.port]
+					num = num + dir
+					num = (2 ^ num % 2) * 4 + num + math.floor(num / 5)
```
sfan5: i'm not sure if this is better than just having two ifs

DS-Minetest: It's a little bit shorter.

mesecons_fpga/init.lua
DS-Minetest: Ok.

mesecons_fpga/init.lua
sfan5: heh, my point was that it's unreadable and obscure

DS-Minetest: That makes it more interesting to read.

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,41 @@ plg.register_nodes({
...
+		end
+		local meta = minetest.get_meta(pos)
+		local instr = lcore.deserialize(meta:get_string("instr"))
+		for i = 1, #instr do
+			for _, op in ipairs(ops) do
+				local o = instr[i][op]
+				if o and o.type == "io" then
+					local num = abcd2[o.port]
+					num = num + dir
+					num = (2 ^ num % 2) * 4 + num - math.floor(num / 5) * 4
```
sfan5: this is still too unreadable, just change it to:
```lua
if num > 4 then num = num - 4 end
if num < 1 then num = num + 4 end
```

DS-Minetest: I'll change it back.

sfan5:
fa58607546dc6b27c88f9a54f4560fdb3a55c344

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,45 @@ plg.register_nodes({
...
+		elseif mode == screwdriver.ROTATE_AXIS then -- counter-clock-wise
+			dir = -1
+			if user and user:is_player() then
+				minetest.chat_send_player(user:get_player_name(),
+						"FPGA ports have been rotated counter-clock-wise.")
+			end
+		end
+		local meta = minetest.get_meta(pos)
+		local instr = lcore.deserialize(meta:get_string("instr"))
+		for i = 1, #instr do
```
numberZero: `for _, operation in ipairs(instr) do`

mesecons_fpga/init.lua
numberZero: *This* looks terribly, I’m glad that sfan5 noticed that. I meant something simpler, like `num = (num + dir) % 4` (that would need changing a few lines above as well).
Also, I didn’t mean *you* to do that; I do understand that that may not be easy for a non-mathematician.

mesecons_fpga/init.lua
```diff
@@ -116,6 +116,45 @@ plg.register_nodes({
...
+				local o = instr[i][op]
+				if o and o.type == "io" then
+					local num = abcd2[o.port]
+					num = num + dir
+					if num > 4 then
+						num = 1
+					elseif num < 1 then
+						num = 4
+					end
+					instr[i][op].port = abcd1[num]
```
numberZero: `o.port = ...`

mesecons_fpga/init.lua
DS-Minetest: That wouldn't give much profit.

mesecons_fpga/init.lua
DS-Minetest: Problematically `0%4` would be `0`, not `4` and `4%4` would be `0`.

mesecons_fpga/init.lua
DS-Minetest: I think, that wouldn't work somehow.

sfan5: pretty sure that would work since `o` is a reference of some sort

DS-Minetest: If I remember correctly, I had tried it with `o` without success.

DS-Minetest: Hm, seems like I don't remember correctly, it works with `o`.
And somehow cw and ccw are swapped. >_< Why didn't I see that before?!¡¿

mesecons_fpga/init.lua
numberZero: >0%4 would be 0, not 4 and 4%4 would be 0.

>that would need changing a few lines above as well

PR #360, opened 2017-09-06
Replace old nodeupdate() with new minetest.check_for_falling()

Thomas--S:
Otherwise mesecons crashes on 0.5.0-dev.

This is not tested, but should work.

### DIFF
--- a/mesecons_hydroturbine/init.lua
+++ b/mesecons_hydroturbine/init.lua
@@ -70,7 +70,7 @@ nodenames = {"mesecons_hydroturbine:hydro_turbine_off"},
 		local waterpos={x=pos.x, y=pos.y+1, z=pos.z}
 		if is_flowing_water(waterpos) then
 			minetest.set_node(pos, {name="mesecons_hydroturbine:hydro_turbine_on"})
-			nodeupdate(pos)
+			minetest.check_for_falling(pos)
 			mesecon.receptor_on(pos)
 		end
 	end,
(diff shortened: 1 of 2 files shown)


mesecons_hydroturbine/init.lua
```diff
@@ -70,7 +70,7 @@ nodenames = {"mesecons_hydroturbine:hydro_turbine_off"},
 		local waterpos={x=pos.x, y=pos.y+1, z=pos.z}
 		if is_flowing_water(waterpos) then
 			minetest.set_node(pos, {name="mesecons_hydroturbine:hydro_turbine_on"})
-			nodeupdate(pos)
+			minetest.check_for_falling(pos)
```
numberZero: is this call necessary at all?

mesecons_pistons/init.lua
```diff
@@ -74,7 +74,7 @@ local piston_remove_pusher = function(pos, node)
 		max_hear_distance = 20,
 		gain = 0.3,
 	})
-	nodeupdate(pusherpos)
+	minetest.check_for_falling(pusherpos)
```
numberZero: this works

numberZero: just noticed that #362 contains this change too

numberZero: #362 merged

numberZero:
928f04f
b08e93f

PR #362, opened 2017-09-07
rework pistons

### LINKED ISSUE #292: Pistons are not rotated correctly by screwdriver

### LINKED ISSUE #314: Breaking a rotated extended piston create an unbreakable extended arm
Over the issue #292 there is a biggest problem which is really a bug

When you move an extended piston with a screwdriver, you move only the piston and not the extended part indeed. Then if you break the piston, the extended arm will remain placed and become unbreakable, you can just rotate it with the screwdriver. World Edit was used to remove it.

bug_rotating_piston

I propose a simple (I think) bugfix : give the possibility to rotate a piston ONLY if the arm is retracted.

DS-Minetest:
- There are no up and down pistons anymore, instead facedir is used.
- User friendly rotation.
- Replaces a part of #302.
- Fixes #292.
- Fixes #314.
- Legacy code ignores face rotation since it would make much more effort and isn't important imo.
- Problematically, if a wire is connected and you rotate the piston face to the side, the wire is connnected to, the wire doesn't disconnect. Maybe I should call `mesecon.execute_autoconnect_hooks_now` also for the wire?


mesecons_pistons/init.lua
```diff
@@ -1,65 +1,63 @@
...
+
+-- Get mesecon rules of pistons
+local function piston_get_rules(node)
+	local dir = minetest.facedir_to_dir(node.param2)
+	for k, v in pairs(dir) do
+		if v ~= 0 then
+			dir = {k, -v}
+			break
+		end
+	end
```
numberZero: I feel I saw this code somewhere already... It could be better to make a function.

mesecons_pistons/init.lua
```diff
@@ -74,67 +72,158 @@ local piston_remove_pusher = function(pos, node)
...
-		mesecon.mvps_move_objects(np, dir, oldstack)
+	local pistonspec = get_pistonspec(node.name, "offname")
+	local dir = vector.multiply(minetest.facedir_to_dir(node.param2), -1)
+	local pusher_pos = vector.add(pos, dir)
+	local success, stack, oldstack = mesecon.mvps_push(pusher_pos, dir, max_push)
+	if not success then
+		return
 	end
+	minetest.set_node(pos, {param2 = node.param2, name = pistonspec.onname})
+	minetest.set_node(pusher_pos,  {param2 = node.param2, name = pistonspec.pusher})
```
numberZero: double space

mesecons_pistons/init.lua
```diff
@@ -74,67 +72,158 @@ local piston_remove_pusher = function(pos, node)
...
+	minetest.sound_play("piston_extend", {
+		pos = pos,
+		max_hear_distance = 20,
+		gain = 0.3,
+	})
+	mesecon.mvps_process_stack(stack)
+	mesecon.mvps_move_objects(pusher_pos, dir, oldstack)
 end
 
 local piston_off = function(pos, node)
```
numberZero: Inconsistency: `local name = function(` vs `local function name(`

mesecons_pistons/init.lua
```diff
@@ -74,67 +72,158 @@ local piston_remove_pusher = function(pos, node)
...
-		minetest.after(0, mesecon.on_placenode, pos, nn)
+local rotations = {
+	{0, 16, 20, 12},
+	{2, 14, 22, 18},
+	{1,  5, 23,  9},
+	{3, 11, 21,  7},
+	{4, 13, 10, 19},
+	{6, 15,  8, 17},
+}
+local function get_rotation(param2)
```
numberZero: missing empty line (and some more below)

mesecons_pistons/init.lua
```diff
@@ -74,67 +72,158 @@ local piston_remove_pusher = function(pos, node)
...
+		end
+	end
+	if not ok then
+		return false
+	end
+	pusher_node.param2 = node.param2
+	minetest.remove_node(pusher_pos)
+	minetest.set_node(pusher_pos_after, pusher_node)
+	minetest.swap_node(pos, node)
+	mesecon.execute_autoconnect_hooks_now(pos, node)
```
numberZero: That’s not enough. When rotating the piston, connected wires don’t disconnect when they should, and piston power state is not changed.

mesecons_pistons/init.lua
DS-Minetest: This is intended since `get_rotation` uses the table above.

mesecons_pistons/init.lua
DS-Minetest: I know. What should I do here?

mesecons_pistons/init.lua
DS-Minetest: Do you mean [here](https://github.com/minetest-mods/mesecons/pull/367/files#diff-fe45ad82229c80f9a7352e533b6c311aR216)?

DS-Minetest:
Note to last commit: mvps should check for falling. #371

mesecons_pistons/init.lua
numberZero: #372

PR #364, opened 2017-09-07
do not override effector param2 on place

### LINKED ISSUE #332: mesecons overrides effector param2 on-place
In bobblocks for example, a block is supposed to be placed as white, by way of place_param2 = 240, but mesecons is forcing the block's param2 to 0 (which is a shade of pink in the palette they use).  Similarly, attempting to alter the param2 of the node in after_place_node is being denied, with the same result.

This is preventing Unified Dyes' continuous-recolor feature from working.

DS-Minetest:
Fixes #332.

### DIFF
--- a/mesecons/services.lua
+++ b/mesecons/services.lua
@@ -21,9 +21,8 @@ mesecon.on_placenode = function(pos, node)
 			end
 			--mesecon.receptor_on (pos, mesecon.conductor_get_rules(node))
 		elseif mesecon.is_conductor_on(node) then
-			local new_node = table.copy(node)
-			new_node.name = mesecon.get_conductor_off(new_node)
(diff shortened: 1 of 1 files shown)


mesecons/services.lua
```diff
@@ -21,7 +21,9 @@ mesecon.on_placenode = function(pos, node)
 			end
 			--mesecon.receptor_on (pos, mesecon.conductor_get_rules(node))
 		elseif mesecon.is_conductor_on(node) then
-			minetest.swap_node(pos, {name = mesecon.get_conductor_off(node)})
+			local new_node = table.copy(node)
```
numberZero: Is copy necessary here?

DS-Minetest: I don't think so.

PR #366, opened 2017-09-07
Improve movestones

### LINKED ISSUE #261: [Feature Request] Vertical Movestones
I've seen this mentioned a lot on the forums; vertical movestones and sticky vertical movestones would mainly be useful for elevators, but I'm sure there are other uses too.
It would also be useful if they can stick to other movestones, so you could have square movestone platforms all moving in one direction or the other.

DS-Minetest:
Closes #261.


numberZero:
Code split is not necessary, they are almost identical.

DS-Minetest:
Ok, now it isn't split anymore.

numberZero:
@DS-Minetest `register_movestone` and `register_vertical_movestone` differ in 1 line only; why are they separate functions? They aren’t that short!
That line is
```lua
-               local direction = get_movestone_direction_h(rulename)
+               local direction = get_movestone_direction_v(rulename)
```
Why not to pass that function as an argument?
```lua
function mesecon.register_movestone(name, def, is_sticky,
                direction_getter = get_movestone_direction_h)
        local function movestone_move(pos, node, rulename)
                local direction = direction_getter(rulename)
```
If you want `register_vertical_movestone` to be globally available but `get_movestone_direction_v` to be local, you can make the former a wrapper for the backend (`register_movestone`).

DS-Minetest:
@numberZero You are right, thanks.
Changed to use one function.

mesecons_movestones/init.lua
```diff
@@ -8,124 +8,148 @@
...
-		local maxpull = mesecon.setting("movestone_max_pull", 50)
-		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, maxpush)
-		if success then
-			mesecon.mvps_process_stack(stack)
-			mesecon.mvps_move_objects(frontpos, direction, oldstack)
-		-- Too large stack/stopper in the way: try again very soon
-		else
-			minetest.after(0.05, movestone_move, pos)
+		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, max_push)
+		if not success then
```
numberZero: Seems to be behavior change: if there is a stopper in the way, which is removed later, the movestone won’t continue moving how it would do before. Am I right?

DS-Minetest: You're right, I haven't thought about that. >_<

numberZero:
check_for_falling is needed

DS-Minetest:
Shouldn't mesecons_mvps call check_for_falling by using `mesecon.register_on_mvps_move`?

DS-Minetest:
#371

PR #367, opened 2017-09-09
mvps: improve object move

### LINKED ISSUE #361: Piston with node doesn't push players/entities with most drawtypes
If you attach a node with a non-default drawtype to the end of a piston, it won't push entities, as shown below:

screenshot_20170907_025221
screenshot_20170907_025232

If there is no node or a full block node, the player/entity will be pushed as expected.

With a nodebox of any size or shape (or any other non-default drawtype e.g. raillike, plantlike), the node will simply clip through any players/entities standing on them, failing to push them up and potentially trapping them.

Edit: It seems that all nodes are treated as if they are full nodes, so if there is a 1/16th slab on a piston, it won't move the entity unless that entity is 2 nodes away from the piston, even if there is space in the adjacent node.

DS-Minetest:
Objects with big collision boxes are moved.
Objects in stack are moved, not only at front. Ergo fixes #361.
~~Objects lying on top of the nodestack aren't moved anymore. Bad?~~ Edit: fixed.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -205,37 +205,50 @@ end)
 
 function mesecon.mvps_move_objects(pos, dir, nodestack)
 	local objects_to_move = {}
-
-	-- Move object at tip of stack, pushpos is position at tip of stack
-	local pushpos = vector.add(pos, vector.multiply(dir, #nodestack))
-
-	local objects = minetest.get_objects_inside_radius(pushpos, 1)
-	for _, obj in ipairs(objects) do
-		table.insert(objects_to_move, obj)
-	end
-
-	-- Move objects lying/standing on the stack (before it was pushed - oldstack)
-	if tonumber(minetest.setting_get("movement_gravity")) > 0 and dir.y == 0 then
-		-- If gravity positive and dir horizontal, push players standing on the stack
-		for _, n in ipairs(nodestack) do
-			local p_above = vector.add(n.pos, {x=0, y=1, z=0})
-			local objects = minetest.get_objects_inside_radius(p_above, 1)
-			for _, obj in ipairs(objects) do
-				table.insert(objects_to_move, obj)
-			end
+	local dir_k
+	local dir_l
+	for k, v in pairs(dir) do
+		if v ~= 0 then
+			dir_k = k
+			dir_l = v
+			break
 		end
 	end
-
-	for _, obj in ipairs(objects_to_move) do
-		local entity = obj:get_luaentity()
-		if not entity or not mesecon.is_mvps_unmov(entity.name) then
-			local np = vector.add(obj:getpos(), dir)
-
-			--move only if destination is not solid
-			local nn = minetest.get_node(np)
-			if not ((not minetest.registered_nodes[nn.name])
-			or minetest.registered_nodes[nn.name].walkable) then
-				obj:setpos(np)
+	for id, obj in pairs(minetest.object_refs) do
+		local obj_pos = obj:get_pos()
+		local cbox = obj:get_properties().collisionbox
+		local min_pos = vector.add(obj_pos, vector.new(cbox[1], cbox[2], cbox[3]))
+		local max_pos = vector.add(obj_pos, vector.new(cbox[4], cbox[5], cbox[6]))
+		local ok = true
+		for k, v in pairs(pos) do
+			local edge1, edge2
+			if k ~= dir_k then
+				edge1 = v - 0.5
+				edge2 = v + 0.5
+			else
+				edge1 = v - 0.5 * dir_l
+				edge2 = v + (#nodestack + 0.5) * dir_l
+				-- Make sure, edge1 is more than edge2:
+				if edge1 > edge2 then
+					edge1, edge2 = edge2, edge1
+				end
+			end
+			if min_pos[k] > edge2 or max_pos[k] < edge1 then
+				ok = false
+				break
+			end
+		end
+		if ok then
+			local ent = obj:get_luaentity()
+			if not ent or not mesecon.is_mvps_unmov(ent.name) then
+				local np = vector.add(obj_pos, dir)
+				-- Move only if destination is not solid or object is inside stack:
+				local nn = minetest.get_node(np)
+				local node_def = minetest.registered_nodes[nn.name]
+				if (node_def and not node_def.walkable) or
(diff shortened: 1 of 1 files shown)


mesecons_mvps/init.lua
```diff
@@ -205,37 +205,50 @@ end)
...
-				obj:setpos(np)
+		if ok then
+			local ent = obj:get_luaentity()
+			if not ent or not mesecon.is_mvps_unmov(ent.name) then
+				local np = vector.add(obj_pos, dir)
+				-- Move only if destination is not solid or object is inside stack:
+				local nn = minetest.get_node(np)
+				local node_def = minetest.registered_nodes[nn.name]
+				if (node_def and not node_def.walkable) or
+						math.abs(obj_pos[dir_k]) <= #nodestack - 0.5 + math.abs(pos[dir_k]) then
```
numberZero: What if signs of `pos[dir_k]` and `obj_pos[dir_k]` are different?

mesecons_mvps/init.lua
```diff
@@ -205,37 +205,50 @@ end)
...
+		local obj_pos = obj:get_pos()
+		local cbox = obj:get_properties().collisionbox
+		local min_pos = vector.add(obj_pos, vector.new(cbox[1], cbox[2], cbox[3]))
+		local max_pos = vector.add(obj_pos, vector.new(cbox[4], cbox[5], cbox[6]))
+		local ok = true
+		for k, v in pairs(pos) do
+			local edge1, edge2
+			if k ~= dir_k then
+				edge1 = v - 0.5
+				edge2 = v + 0.5
```
numberZero: > Objects lying on top of the nodestack aren't moved anymore. Bad?

You mean, when pushing horizontally? That’s acceptable, but may be fixed by increasing (slightly) constant here, if I understand the code correctly.

mesecons_mvps/init.lua
```diff
@@ -205,37 +205,50 @@ end)
...
+			else
+				edge1 = v - 0.5 * dir_l
+				edge2 = v + (#nodestack + 0.5) * dir_l
+				-- Make sure, edge1 is more than edge2:
+				if edge1 > edge2 then
+					edge1, edge2 = edge2, edge1
+				end
+			end
```
numberZero: `edge`s may be pre-calculated

mesecons_mvps/init.lua
```diff
@@ -205,37 +205,50 @@ end)
...
-
-	-- Move objects lying/standing on the stack (before it was pushed - oldstack)
-	if tonumber(minetest.setting_get("movement_gravity")) > 0 and dir.y == 0 then
-		-- If gravity positive and dir horizontal, push players standing on the stack
-		for _, n in ipairs(nodestack) do
-			local p_above = vector.add(n.pos, {x=0, y=1, z=0})
-			local objects = minetest.get_objects_inside_radius(p_above, 1)
-			for _, obj in ipairs(objects) do
-				table.insert(objects_to_move, obj)
+	for id, obj in pairs(minetest.object_refs) do
```
numberZero: Seems horribly, but that’s the only way; MT itself does the same, all the time...

mesecons_mvps/init.lua
```diff
@@ -205,37 +205,50 @@ end)
...
-			local np = vector.add(obj:getpos(), dir)
-
-			--move only if destination is not solid
-			local nn = minetest.get_node(np)
-			if not ((not minetest.registered_nodes[nn.name])
-			or minetest.registered_nodes[nn.name].walkable) then
-				obj:setpos(np)
+		if ok then
+			local ent = obj:get_luaentity()
+			if not ent or not mesecon.is_mvps_unmov(ent.name) then
```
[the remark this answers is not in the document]
DS-Minetest: `not (ent and …)`

mesecons_mvps/init.lua
DS-Minetest: ?

DS-Minetest: ? This works well enough imo, I can't find a simpler way.

DS-Minetest:
Comments addressed.

mesecons_mvps/init.lua
```diff
@@ -205,37 +205,55 @@ end)
...
-			local np = vector.add(obj:getpos(), dir)
-
-			--move only if destination is not solid
-			local nn = minetest.get_node(np)
-			if not ((not minetest.registered_nodes[nn.name])
-			or minetest.registered_nodes[nn.name].walkable) then
-				obj:setpos(np)
+		if ok then
+			local ent = obj:get_luaentity()
+			if not (ent and mesecon.is_mvps_unmov(ent.name)) then
```
numberZero: Does this work at all?
The correct way seems to be `if ent and not mesecon.is_mvps_unmov(ent.name) then`

mesecons_mvps/init.lua
numberZero: This exact way, but before `for id, obj in pairs(minetest.object_refs) do` and not in the inner loop.

mesecons_mvps/init.lua
```diff
@@ -240,13 +240,18 @@ function mesecon.mvps_move_objects(pos, dir, nodestack)
...
 		if ok then
 			local ent = obj:get_luaentity()
-			if not ent or not mesecon.is_mvps_unmov(ent.name) then
+			if not (ent and mesecon.is_mvps_unmov(ent.name)) then
 				local np = vector.add(obj_pos, dir)
 				-- Move only if destination is not solid or object is inside stack:
 				local nn = minetest.get_node(np)
 				local node_def = minetest.registered_nodes[nn.name]
+				local nodestack_front = (#nodestack - 0.5 + math.abs(pos[dir_k]))
+				nodestack_front = nodestack_front * pos[dir_k]/math.abs(pos[dir_k])
```
numberZero: Weird math. Can you write something simpler? Remember that you have `dir_l`.
Or let me do that (a bit later).

mesecons_mvps/init.lua
DS-Minetest: This is correct, don't forget players.

numberZero: Oh, sorry.
Then the previous variant was better. But such clever code needs a comment, at least.
And, well, there is `is_player`.

DS-Minetest: So, `if obj:is_player() or not mesecon.is_mvps_unmov(obj:get_luaentity().name) then`?

PR #370, opened 2017-09-12
improve some rules

### LINKED ISSUE #369: Pressure plate and solar panel don't power block directly below
Presure plates and solar panels are weird. While they propagate power to the 4 adjacent blocks *below* themselves, they do *not* power the block directly beneath it (nor do they power the block 2 blocks beneath it; I'm not sure if this is intentional).

DS-Minetest:
Fixes #369.


mesecons/presets.lua
```diff
@@ -1,59 +1,84 @@
...
+		rules = mesecon.rotate_rules_up(rules)
+	elseif dir.x == 1 then
+		rules = mesecon.rotate_rules_down(rules)
+	elseif dir.z == -1 then
+		rules = mesecon.rotate_rules_right(mesecon.rotate_rules_up(rules))
+	elseif dir.z == 1 then
+		rules = mesecon.rotate_rules_left(mesecon.rotate_rules_up(rules))
 	end
 	return rules
 end
```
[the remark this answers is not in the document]
DS-Minetest: Problematically `presets.lua` is loaden after `internal.lua` since `internal.lua` needs the preset rules.

mesecons/presets.lua
```diff
@@ -45,42 +45,44 @@ mesecon.rules.alldirs = {
...
-	elseif dir.z == 1 then
-		rules = mesecon.rotate_rules_right(rules)
-	elseif dir.y == -1 then
-		rules = mesecon.rotate_rules_up(rules)
-	elseif dir.y == 1 then
-		rules = mesecon.rotate_rules_down(rules)
-	end
-	return rules
+local rules_wallmounted, rules_buttonlike = {}, {}
+minetest.after(0, function() -- The rotate_rules functions are not yet loaden.
```
numberZero: Er, what? Are you really sure these tables may not be ever used before the server loads?

Really, `rotate_rules_*` should be in `util.lua`, somewhere near `flattenrules`. Dunno how could they appear in `internal.lua`.

DS-Minetest: You're right, I'll move them.

PR #371, opened 2017-10-08
check for falling on mvps_move

DS-Minetest:
~~The falling object is pushed after spawning by current pistons and movestones. Is that bad?~~

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -205,12 +205,6 @@ function mesecon.mvps_push_or_pull(pos, stackdir, movedir, maximum, all_pull_sti
 	return true, nodes, oldstack
 end
 
-mesecon.register_on_mvps_move(function(moved_nodes)
-	for _, n in ipairs(moved_nodes) do
-		mesecon.on_placenode(n.pos, n.node)
-	end
-end)
-
 function mesecon.mvps_move_objects(pos, dir, nodestack)
 	local objects_to_move = {}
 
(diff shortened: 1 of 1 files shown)


numberZero:
Yes that’s bad. E.g. it can skip an 1-node-width hole.
To fix that, you need to call `check_for_falling` after moving objects.

DS-Minetest:
Can maybe be done with `minetest.after`.

DS-Minetest:
Done, works.

numberZero:
screenshot_20171015_165832

DS-Minetest:
Can this also be reproduced with only one movestone?

Hm, maybe the first one fell but was pushed and as the signal turned off the second one fell.

numberZero:
I think (haven’t tested, though) that the function you passed to `minetest.after` was called *after* all these movestones moved the stack. That might also happen with several pistons chained.

BTW, you call `minetest.after` for each and every node in the stack... Isn’t that too much?

*Just realized:* the effect I noticed is desirable. Would the node start falling *between* moves, it still would be moved as an object. So the only problem remains is tons of `after`, but it seems to be fairly lightweight. And while it would be lost on server crash, the whole movement may not be considered as performed correctly in that case, so not a problem IMO.

mesecons_mvps/init.lua
```diff
@@ -258,6 +252,11 @@ mesecon.register_on_mvps_move(mesecon.move_hot_nodes)
 mesecon.register_on_mvps_move(function(moved_nodes)
 	for i = 1, #moved_nodes do
 		local moved_node = moved_nodes[i]
+		mesecon.on_placenode(moved_node.pos, moved_node.node)
+		minetest.after(0, function()
+			minetest.check_for_falling(moved_node.oldpos)
+			minetest.check_for_falling(moved_node.pos)
```
numberZero: Why checking `oldpos`? The node isn’t there anymore.

DS-Minetest: After a node is pulled, there might have been falling nodes above it.

PR #374, opened 2017-10-08
mvps: pull objects

### LINKED ISSUE #373: Sticky MVPS don’t pull objects
@DS-Minetest

DS-Minetest:
Fixes #373.

### DIFF
--- a/mesecons_movestones/init.lua
+++ b/mesecons_movestones/init.lua
@@ -50,7 +50,6 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 			minetest.get_node_timer(pos):start(timer_interval)
 			return
 		end
-		mesecon.mvps_process_stack(stack)
 		mesecon.mvps_move_objects(frontpos, direction, oldstack)
 
 		-- ### Step 2: Move the movestone ###
@@ -61,9 +60,13 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 		minetest.get_node_timer(frontpos):start(timer_interval)
 
 		-- ### Step 3: If sticky, pull stack behind ###
-		if is_sticky then
-			local backpos = vector.subtract(pos, direction)
-			mesecon.mvps_pull_all(backpos, direction, max_pull)
+		if not is_sticky then
+			return
+		end
+		local backpos = vector.subtract(pos, direction)
+		success, stack, oldstack = mesecon.mvps_pull_all(backpos, direction, max_pull)
+		if success then
+			mesecon.mvps_move_objects(backpos, vector.multiply(direction, -1), oldstack, -1)
 		end
 	end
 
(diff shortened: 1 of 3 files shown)


mesecons_movestones/init.lua
```diff
@@ -50,7 +50,6 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 			minetest.get_node_timer(pos):start(timer_interval)
 			return
 		end
-		mesecon.mvps_process_stack(stack)
```
numberZero: Why removed?

mesecons_movestones/init.lua
```diff
@@ -61,9 +60,13 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
...
-		if is_sticky then
-			local backpos = vector.subtract(pos, direction)
-			mesecon.mvps_pull_all(backpos, direction, max_pull)
+		if not is_sticky then
+			return
+		end
+		local backpos = vector.subtract(pos, direction)
+		success, stack, oldstack = mesecon.mvps_pull_all(backpos, direction, max_pull)
+		if success then
+			mesecon.mvps_move_objects(backpos, vector.multiply(direction, -1), oldstack, -1)
```
numberZero: Looks strange, but acceptable.

mesecons_movestones/init.lua
DS-Minetest: This function isn't needed anymore since `mesecon.mvps_push` was before already called [which calls `mesecon.mvps_push_or_pull`](https://github.com/minetest-mods/mesecons/blob/7094f0b6065cb657de743fc6a919129a67bc079f/mesecons_mvps/init.lua#L138L140) [which calls `on_mvps_move`](https://github.com/minetest-mods/mesecons/blob/7094f0b6065cb657de743fc6a919129a67bc079f/mesecons_mvps/init.lua#L203) [which calls the same](https://github.com/minetest-mods/mesecons/blob/7094f0b6065cb657de743fc6a919129a67bc079f/mesecons_mvps/init.lua#L208L212) [as `mesecon.mvps_process_stack` does](https://github.com/minetest-mods/mesecons/blob/7094f0b6065cb657de743fc6a919129a67bc079f/mesecons_mvps/init.lua#L49L54).

numberZero: Okay then

numberZero:
It was pulling objects 1 node far from the stack being pulled (that’s OK for pushing, though). I fixed that.
@DS-Minetest okay?

DS-Minetest:
Um, shouldn't it pull anything in front of the stack?

numberZero:
IMO, that doesn’t make much sense. While pushing objects in front of the stack is obviously correct, as they would be crossed by the stack otherwise (assuming it consists of solid nodes), pulling objects not in the stack looks strangely.

numberZero:
59273d5

Issue #376, opened 2017-10-17
TOCTOU vulnerability on Digiline serialized message size limit


Hawk777:
There is a time-of-check-to-time-of-use vulnerability in the Digiline serialized message limit. This code exploits it:

```lua
if event.type == "program" then
  local s = "hello hello hello hello hello asdf asdf das fa fas fadsf dsf adsf dsafa sd fasd fdsa"
  for i = 1, 10 do
    s = s .. s
  end
  local t = {}
  for i = 1, 150 do
    t[i] = s
  end
  local msg = {x = 27, y = 29}
  digiline_send("foo", msg)
  msg.z = t
end
```

If you were to move the `msg.z = t` line above the `digiline_send` line, no message would be sent because the serialized form is too long. By keeping it below, the message *as it exists when `digiline_send` is called* is small enough to pass the check, but because the table is never copied but only passed by reference into the Digiline network, the `msg.z = t` line adds a new key, with huge value, to the live table in the Digiline network and affects the message delivered to another node (verified by having another Luacontroller on the network print out whether `z` is a key in the received table or not, and it is).

DS-Minetest:
Possible fix:
Edit [these lines](https://github.com/minetest-mods/mesecons/blob/4f400f3fae1955a5a54e710167a91a2cba48d280/mesecons_luacontroller/init.lua#L291L293) to:
```lua
		minetest.after(0, digilines.receptor_send, pos, digiline.rules.default, channel, type(msg) == "table" and table.copy(msg) or msg)
```

numberZero:
@DS-Minetest `table.copy` does shallow copy, so won’t help.
Also, don’t forget there was a colon, thus
`minetest.after(0, digilines.receptor_send, digilines, pos,`

Note that you can check size limit at the same time as making the copy.

Hawk777:
`table.copy` is defined in Minetest core at `builtin/common/misc_helpers.lua`, and it’s a deep copy, not a shallow copy.

The Luacontroller source code includes the following comment:

> It is technically possible to send functions over the wire since the high performance impact of stripping those from the data has been decided to not be worth the added realism.

Yet the performance cost of stripping functions is just a table walk AFAICT, which is precisely what `table.copy` does. Is that comment now obsolete (i.e. is the performance cost now considered acceptable)? Or is it still a legitimate reason why we can’t do a table walk?

[...]

PR #379, opened 2017-10-23
[WIP] Close vulnerability and optimize digiline_send

Hawk777:
Fixes GH-376.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -270,6 +270,88 @@ local function get_interrupt(pos)
 end
 
 
+-- Given a message object passed to digiline_send, clean it up into a form
+-- which is safe to transmit over the network and compute its “cost” (a very
+-- rough estimate of its memory usage).
+--
+-- The cleaning comprises the following:
+-- 1. Functions (and userdata, though user scripts ought not to get hold of
+--    those in the first place) are removed, because they break the model of
+--    Digilines as a network that carries basic data, and they could exfiltrate
+--    references to mutable objects from one Luacontroller to another, allowing
+--    inappropriate high-bandwidth, no-wires communication.
+-- 2. Tables are duplicated because, being mutable, they could otherwise be
+--    modified after the send is complete in order to change what data arrives
+--    at the recipient, perhaps in violation of the previous cleaning rule or
+--    in violation of the message size limit.
+--
+-- The cost indication is only approximate; it’s not a perfect measurement of
+-- the number of bytes of memory used by the message object.
+--
+-- Parameters:
+-- msg -- the message to clean
+-- back_references -- for internal use only; do not provide
+--
+-- Returns:
+-- 1. The cleaned object.
+-- 2. The approximate cost of the object.
+local function clean_and_weigh_digiline_message(msg, back_references)
+	back_references = back_references or {}
+	local t = type(msg)
+	if t == "string" then
+		-- Strings are immutable so can be passed by reference, and cost their
+		-- length.
+		return msg, #msg
+	elseif t == "number" then
+		-- Numbers are passed by value so need not be touched, and cost 8 bytes
+		-- as all numbers in Lua are doubles.
+		return msg, 8
+	elseif t == "table" then
+		-- Tables are duplicated. Check if this table has been seen before
+		-- (self-referential or shared table); if so, reuse the cleaned value
+		-- of the previous occurrence, maintaining table topology and avoiding
+		-- infinite recursion, and charge zero bytes for this as the object has
+		-- already been counted.
+		local bref = back_references[msg]
+		if bref then
+			return bref, 0
+		end
+		-- Construct a new table by cleaning all the keys and values and adding
+		-- up their costs, plus 8 bytes as a rough estimate of table overhead.
+		local cost = 8
+		local ret = {}
+		back_references[msg] = ret
+		for k, v in pairs(msg) do
+			local k_cost, v_cost
+			k, k_cost = clean_and_weigh_digiline_message(k, back_references)
+			v, v_cost = clean_and_weigh_digiline_message(v, back_references)
+			if k ~= nil and v ~= nil then
+				-- Only include an element if its key and value are of legal
+				-- types.
+				ret[k] = v
+			end
+			-- If we only counted the cost of a table element when we actually
+			-- used it, we would be vulnerable to the following attack:
+			-- 1. Construct a huge table (too large to pass the cost limit).
+			-- 2. Insert it somewhere in a table, with a function as a key.
+			-- 3. Insert it somewhere in another table, with a number as a key.
+			-- 4. The first occurrence doesn’t pay the cost because functions
+			--    are stripped and therefore the element is dropped.
+			-- 5. The second occurrence doesn’t pay the cost because it’s in
+			--    back_references.
+			-- By counting the costs regardless of whether the objects will be
+			-- included, we avoid this attack; it may overestimate the cost of
+			-- some messages, but only those that won’t be delivered intact
+			-- anyway because they contain illegal object types.
+			cost = cost + k_cost + v_cost
+		end
+		return ret, cost
+	else
+		return nil, 0
+	end
+end
+
+
 local function get_digiline_send(pos)
 	if not digiline then return end
 	return function(channel, msg)
@@ -278,13 +360,9 @@ local function get_digiline_send(pos)
 			return false
 		end
 
-		-- It is technically possible to send functions over the wire since
-		-- the high performance impact of stripping those from the data has
-		-- been decided to not be worth the added realism.
-		-- Make sure serialized version of the data is not insanely long to
-		-- prevent DoS-like attacks
-		local msg_ser = minetest.serialize(msg)
-		if #msg_ser > mesecon.setting("luacontroller_digiline_maxlen", 50000) then
+		local msg_cost
+		msg, msg_cost = clean_and_weigh_digiline_message(msg)
+		if msg_cost > mesecon.setting("luacontroller_digiline_maxlen", 50000) then
 			return false
 		end
 
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -271,7 +271,7 @@ end
 
 
 -- Given a message object passed to digiline_send, clean it up into a form
--- which is safe to transmit over the network and compute its “cost” (a very
+-- which is safe to transmit over the network and compute its "cost" (a very
 -- rough estimate of its memory usage).
 --
 -- The cleaning comprises the following:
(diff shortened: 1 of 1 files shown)


mesecons_luacontroller/init.lua
```diff
@@ -270,6 +270,88 @@ local function get_interrupt(pos)
...
+			-- 2. Insert it somewhere in a table, with a function as a key.
+			-- 3. Insert it somewhere in another table, with a number as a key.
+			-- 4. The first occurrence doesn’t pay the cost because functions
+			--    are stripped and therefore the element is dropped.
+			-- 5. The second occurrence doesn’t pay the cost because it’s in
+			--    back_references.
+			-- By counting the costs regardless of whether the objects will be
+			-- included, we avoid this attack; it may overestimate the cost of
+			-- some messages, but only those that won’t be delivered intact
+			-- anyway because they contain illegal object types.
```
numberZero: That is, you compute weight of what is requested to be sent, and not of what will be sent. That’s correct.

mesecons_luacontroller/init.lua
```diff
@@ -270,6 +270,88 @@ local function get_interrupt(pos)
...
+-- Given a message object passed to digiline_send, clean it up into a form
+-- which is safe to transmit over the network and compute its “cost” (a very
+-- rough estimate of its memory usage).
+--
+-- The cleaning comprises the following:
+-- 1. Functions (and userdata, though user scripts ought not to get hold of
+--    those in the first place) are removed, because they break the model of
+--    Digilines as a network that carries basic data, and they could exfiltrate
+--    references to mutable objects from one Luacontroller to another, allowing
+--    inappropriate high-bandwidth, no-wires communication.
```
numberZero: Oh well, that’s real. I imagine one scheme:

1. One LuaC sends a function that references some variable through digiline channel.
2. **Two** controllers receive it.
3. One of them (those who received it first) is able to pass arbitrary data to another one, bypassing any protection.

The problem is that a mere table would have the same effect, if not cloned...

Hawk777:
I guess there is a possible vulnerability where node A sends a message, Luacontroller B receives it and modifies it, and then object C sees the modified message. Is that what you mean by done on data reception? I don’t know that this belongs in Digilines core, because as long as the sender and recipient are both trusted, there isn’t really any problem with them sending functions, userdata, or huge tables to each other as far as I can see. It’s really only a problem when you introduce Luacontrollers and give them access to the network, isn’t it?

mesecons_luacontroller/init.lua
```diff
@@ -270,6 +270,88 @@ local function get_interrupt(pos)
 end
 
 
+-- Given a message object passed to digiline_send, clean it up into a form
+-- which is safe to transmit over the network and compute its “cost” (a very
```
DS-Minetest: I'm not sure if it's good to have a `“` here instead of a `"`.

mesecons_luacontroller/init.lua
```diff
@@ -270,6 +270,88 @@ local function get_interrupt(pos)
...
+			--    back_references.
+			-- By counting the costs regardless of whether the objects will be
+			-- included, we avoid this attack; it may overestimate the cost of
+			-- some messages, but only those that won’t be delivered intact
+			-- anyway because they contain illegal object types.
+			cost = cost + k_cost + v_cost
+		end
+		return ret, cost
+	else
+		return nil, 0
```
DS-Minetest: I suggest returning t, #t instead. Else, there will be _empty_ digiline signals.

DS-Minetest:
As you are optimizing digiline_send, could you please replace
```lua
minetest.after(0, function()
 	digiline:receptor_send(pos, digiline.rules.default, channel, msg)
end)
```
[here](https://github.com/minetest-mods/mesecons/pull/379/files#diff-0875d6fbb7ce4b468db9d73b9f3889a9R369R371) with the new `digilines.receptor_send`?
_(Can't make in-line comment because it's out of range.)_

mesecons_luacontroller/init.lua
Hawk777: Meh. I default to nice quotation marks when typing. I’ll delete it if you don’t like it.

mesecons_luacontroller/init.lua
Hawk777: This is the branch that prevents functions and userdata from being put in Digilines messages; remember this function is recursive so this branch can also be taken for keys and values in tables. Do you want their stringified forms kept in? I thought it would be more clear to the player that functions are prohibited if they see them completely disappear, rather than become stringified. Or even just completely fail if you try to send a message with a function anywhere in it; as I see it, it should be up to the user to write code that obeys the rules, and we should be able to just kill execution if they don’t. I agree, though, that I forgot to put a `nil` check in `get_digiline_send` to avoid sending an empty message if the value the script provides is entirely prohibited, rather than just containing a prohibited value somewhere in a table.

Hawk777:
@DS-Minetest, I can’t reply directly to your last comment about `digilines.receptor_send`, but fixed in 79d61f552a52de6691b4d3743fc619adb07a8598.

mesecons_luacontroller/init.lua
Hawk777: Shall I add a call to `clean_and_weigh_digiline_message` to `digiline.effector.action`? I think that should do the job nicely. It wouldn’t count against anyone’s event quota because it’s called outside the scope of a user script, and for the same reason that invocation, in contrast to the others, could be JITed. So hopefully the overhead wouldn’t be too high.

mesecons_luacontroller/init.lua
Hawk777: `nil` check added in cf59a8e90d0c.

mesecons_luacontroller/init.lua
```diff
@@ -270,26 +270,108 @@ local function get_interrupt(pos)
...
+		local msg_cost
+		msg, msg_cost = clean_and_weigh_digiline_message(msg)
+		if msg == nil or msg_cost > mesecon.setting("luacontroller_digiline_maxlen", 50000) then
 			return false
 		end
 
 		minetest.after(0, function()
-			digiline:receptor_send(pos, digiline.rules.default, channel, msg)
+			digilines.receptor_send(pos, digilines.rules.default, channel, msg)
 		end)
```
DS-Minetest: Why not `minetest.after(0, digilines.receptor_send, pos, digilines.rules.default, channel, msg)` instead of making a new function?

mesecons_luacontroller/init.lua
DS-Minetest: :+1:
(But `return` would be enough here I think since afaik it's the same as `return nil, nil` and `msg_cost` doesn't need to be set.)

mesecons_luacontroller/init.lua
```diff
@@ -270,26 +270,108 @@ local function get_interrupt(pos)
...
+--
+-- Parameters:
+-- msg -- the message to clean
+-- back_references -- for internal use only; do not provide
+--
+-- Returns:
+-- 1. The cleaned object.
+-- 2. The approximate cost of the object.
+local function clean_and_weigh_digiline_message(msg, back_references)
+	back_references = back_references or {}
```
DS-Minetest: Maybe move this down since it's only used for tables.

mesecons_luacontroller/init.lua
```diff
@@ -270,26 +270,108 @@ local function get_interrupt(pos)
...
+	local t = type(msg)
+	if t == "string" then
+		-- Strings are immutable so can be passed by reference, and cost their
+		-- length.
+		return msg, #msg
```
DS-Minetest: Shouldn't this be a little ~~bit~~ more? Does lua have escape characters? And do lua characters only use one byte? I don't know…

mesecons_luacontroller/init.lua
DS-Minetest: As I said, I'm not sure. But maybe it can give problems with constancy because other people might not use those marks but the normal ones. But it's not important imo, I'm neutral.

Hawk777: So am I. So fixed in 80fc385e7183.

mesecons_luacontroller/init.lua
Hawk777: Actually we do need to return a proper numeric cost. Not for the case when we’re returning directly to `digiline_send`, but rather when we’re returning to a recursive call to clean up a table key or value; in that case, if the key or value is cleaned to `nil` then the table element is discarded, but the cost is still added onto the running total.

mesecons_luacontroller/init.lua
Hawk777: Because I didn’t know you could do that :smile: fixed in b7fa47994529.

mesecons_luacontroller/init.lua
Hawk777: Fixed in d55884dee592.

mesecons_luacontroller/init.lua
Hawk777: > Shouldn’t this be a little bit more?

Probably yes, by a constant added amount I think, to represent the overhead of the string management data structures. In plain Lua (not sure about Luajit), it looks like strings are allocated as a single contiguous block of memory with a header followed by the characters followed by a NUL (for C compatibility). The header is a pointer, a few bytes, an `unsigned int`, and a `size_t`. On a 64-bit platform, that header will be 24 bytes long. Fixed in 56dee6dc414f.

> Does lua have escape characters?

Yes, but that’s irrelevant to this discussion. Escape characters only apply in source code, and the cost function is meant to measure runtime memory cost, not source code size.

> And do lua characters use only one byte?

Yes, according to [the manual](https://www.lua.org/manual/5.1/manual.html#2.2):

> String represents arrays of characters. Lua is 8-bit clean: strings can contain any 8-bit character, including embedded zeros ('\0')

This also agrees with my inspection of the source code, that all string lengths are measured in terms of bytes.

mesecons_luacontroller/init.lua
Hawk777: @numberZero ping?

mesecons_luacontroller/init.lua
```diff
@@ -270,27 +270,108 @@ local function get_interrupt(pos)
...
+		-- Construct a new table by cleaning all the keys and values and adding
+		-- up their costs, plus 8 bytes as a rough estimate of table overhead.
+		local cost = 8
+		local ret = {}
+		back_references[msg] = ret
+		for k, v in pairs(msg) do
+			local k_cost, v_cost
+			k, k_cost = clean_and_weigh_digiline_message(k, back_references)
+			v, v_cost = clean_and_weigh_digiline_message(v, back_references)
+			if k ~= nil and v ~= nil then
```
numberZero: `if k == nil or v == nil then return nil, 0 end`?

mesecons_luacontroller/init.lua
numberZero: Maybe. Double-clonning shouldn’t be any slower than serializing used currently.

Hawk777: Fixed in ca03f8193fbf.

mesecons_luacontroller/init.lua
Hawk777: So you want to abort on any invalid elements rather than filter them out of the table as I do here?

PR #383, opened 2017-10-29
FPGA: overheat

### LINKED ISSUE #320: FPGA: Doesn't overheat
Even if I connect two pins and set one to be the inverse of the other, the FPGA won't overheat and instead just sits there oscillating (and making the server lag a bit).

DS-Minetest:
Fixes #320.

### DIFF
--- a/mesecons_fpga/init.lua
+++ b/mesecons_fpga/init.lua
@@ -72,7 +72,7 @@ plg.register_nodes({
 	sunlight_propagates = true,
 	paramtype = "light",
 	walkable = true,
-	groups = {dig_immediate = 2, mesecon = 3},
+	groups = {dig_immediate = 2, mesecon = 3, overheat = 1},
 	drop = "mesecons_fpga:fpga0000",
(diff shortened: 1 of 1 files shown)


mesecons_fpga/init.lua
```diff
@@ -309,6 +309,10 @@ plg.update = function(pos)
 	local meta = minetest.get_meta(pos)
 	if meta:get_int("valid") ~= 1 then
 		return
+	elseif mesecon.do_overheat(pos) then
```
sfan5: how expensive is this function?

DS-Minetest: Afaik, it's cheap. And luacontroller and gates use it, too.
https://github.com/minetest-mods/mesecons/blob/f78937b351d42c04bb2f07ca4a365d456289f30e/mesecons/services.lua#L85L95

sfan5:
db39eef2f6583c91d3d2c8823eaa1b627110980d

PR #384, opened 2017-10-29
gates: update wires after overheat

DS-Minetest:


numberZero:
Sorry for long wait, but I’m just not sure should it work this way. If a gate overheats, it leaves a circuit broken. Why should wires be updated to a “proper” state in such event?

numberZero:
No support from anyone, including the author.

Also note that updating wires might break them if the configuration is fragile (i.e. using special wires/states, like button’s receiver). Would that be a bug or a feature, well... dunno. I’d rather not do that.

DS-Minetest:
I think, you're right, not updated wires shouldn't be seen as buggy.

PR #386, opened 2017-10-29
make is_ground_content to false

### LINKED ISSUE #295: Mesecons nodes can be destroyed by cave generator
I noticed that many Mesecons nodes do not have `is_ground_content = false` set (default is `true`). This means, all those nodes are possible victims for the cave generator.
Please add `is_ground_content = false` to all nodes.

DS-Minetest:
Fixes #295.


numberZero:
`is_ground_content` defaults to false.

sfan5:
[it doesn't](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L4524)

numberZero:
@sfan5 https://github.com/minetest/minetest/blob/master/src/nodedef.cpp#L356

sfan5:
@numberZero https://github.com/minetest/minetest/blob/master/builtin/game/item.lua#L689

PR #390, opened 2017-12-04
Craft fallbacks for detectors without luacontroller enabled

CloudyProton:
As of now, users who disable mesecons_luacontroller do not have an option to craft node or object detectors. Add check.

### DIFF
--- a/mesecons_detector/init.lua
+++ b/mesecons_detector/init.lua
@@ -99,14 +99,25 @@ minetest.register_node("mesecons_detector:object_detector_on", {
 	on_blast = mesecon.on_blastnode,
 })
 
-minetest.register_craft({
-	output = 'mesecons_detector:object_detector_off',
-	recipe = {
-		{"default:steel_ingot", "default:steel_ingot", "default:steel_ingot"},
-		{"default:steel_ingot", "mesecons_luacontroller:luacontroller0000", "default:steel_ingot"},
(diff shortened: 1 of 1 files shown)


numberZero:
What if they disable mesecons_microcontroller as well?

CloudyProton:
Hmm hadn't thought of that. It might be unreasonable to expect a craft recipe for every combination of enabling/disabling between 30 something sub-mods. Maybe it should ultimately have a default:<materials> fallback?

numberZero:
Er, nvm; it was that way for 5 years.

> Maybe it should ultimately have a default: fallback?

Maybe; e.g. mese crystal fragment. Or some mesecon material, like silicon; dunno. @DS-Minetest any ideas?

DS-Minetest:
Perhaps add use for a second fallback fpga and for a third one a gate and for last one insulated mesecon wire?

numberZero:
@CloudyProton don’t check, register both recipes.

CloudyProton:
While that is more streamlined, a disabled component still shows up as "unknown item" in crafting guides. Or should it be the job of craft guide mods to cull recipes with unknown ingredients?

numberZero:
> Or should it be the job of craft guide mods to cull recipes with unknown ingredients?

Exactly. Unified Inventory does that well AFAIK.

numberZero:
@CloudyProton Thanks for the code.
P.S. Please always make sure you don’t add trailing spaces etc.

PR #393, opened 2018-01-02
Hopefully fix LuaController security in a way that doesn't break compat. (NOTE: Conflicts with #379.)

20kdc:
EDIT: Fixes, post-rebase:

1. Lack of 'safe' on minetest.deserialize usage
2. String sandbox bypass via (""):evil()
3. Loss of upcoming digilines messages on server shutdown
4. LCs failing to show information on some errors
5. Interrupt IDs as infinite data storage

With any luck something hasn't been broken this time.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -13,8 +13,10 @@
 -- newport = merge_port_states(state1, state2): just does result = state1 or state2 for every port
 -- set_port(pos, rule, state): activates/deactivates the mesecons according to the port states
 -- set_port_states(pos, ports): Applies new port states to a Luacontroller at pos
--- run(pos): runs the code in the controller at pos
--- reset_meta(pos, code, errmsg): performs a software-reset, installs new code and prints error messages
+-- run_inner(pos, code, event): runs code on the controller at pos and event
+-- reset_formspec(pos, code, errmsg): installs new code and prints error messages, without resetting LCID
+-- reset_meta(pos, code, errmsg): performs a software-reset, installs new code and prints error message
+-- run(pos, event): a wrapper for run_inner which gets code & handles errors via reset_meta
 -- resetn(pos): performs a hardware reset, turns off all ports
 --
 -- The Sandbox
@@ -141,7 +143,7 @@ local function set_port_states(pos, ports)
 		-- Solution / Workaround:
 		-- Remember which output was turned off and ignore next "off" event.
 		local meta = minetest.get_meta(pos)
-		local ign = minetest.deserialize(meta:get_string("ignore_offevents")) or {}
+		local ign = minetest.deserialize(meta:get_string("ignore_offevents"), true) or {}
 		if ports.a and not vports.a and not mesecon.is_powered(pos, rules.a) then ign.A = true end
 		if ports.b and not vports.b and not mesecon.is_powered(pos, rules.b) then ign.B = true end
 		if ports.c and not vports.c and not mesecon.is_powered(pos, rules.c) then ign.C = true end
@@ -183,7 +185,7 @@ end
 
 local function ignore_event(event, meta)
 	if event.type ~= "off" then return false end
-	local ignore_offevents = minetest.deserialize(meta:get_string("ignore_offevents")) or {}
+	local ignore_offevents = minetest.deserialize(meta:get_string("ignore_offevents"), true) or {}
 	if ignore_offevents[event.pin.name] then
 		ignore_offevents[event.pin.name] = nil
 		meta:set_string("ignore_offevents", minetest.serialize(ignore_offevents))
@@ -236,6 +238,7 @@ local function remove_functions(x)
 	local seen = {}
 
 	local function rfuncs(x)
+		if x == nil then return end
 		if seen[x] then return end
 		seen[x] = true
 		if type(x) ~= "table" then return end
@@ -259,12 +262,27 @@ local function remove_functions(x)
 	return x
 end
 
-local function get_interrupt(pos)
+-- itbl: Flat table of functions to run after sandbox cleanup, used to prevent various security hazards
+local function get_interrupt(pos, itbl, send_warning)
 	-- iid = interrupt id
 	local function interrupt(time, iid)
+		-- NOTE: This runs within string metatable sandbox, so don't *rely* on anything of the form (""):y
+		-- Hence the values get moved out. Should take less time than original, so totally compatible
 		if type(time) ~= "number" then return end
-		local luac_id = minetest.get_meta(pos):get_int("luac_id")
-		mesecon.queue:add_action(pos, "lc_interrupt", {luac_id, iid}, time, iid, 1)
+		table.insert(itbl, function ()
+			-- Outside string metatable sandbox, can safely run this now
+			local luac_id = minetest.get_meta(pos):get_int("luac_id")
+			-- Check if IID is dodgy, so you can't use interrupts to store an infinite amount of data.
+			-- Note that this is safe from alter-after-free because this code gets run after the sandbox has ended.
+			-- This runs outside of the timer and *shouldn't* harm perf. unless dodgy data is being sent in the first place
+			iid = remove_functions(iid)
+			local msg_ser = minetest.serialize(iid)
+			if #msg_ser <= mesecon.setting("luacontroller_interruptid_maxlen", 256) then
+				mesecon.queue:add_action(pos, "lc_interrupt", {luac_id, iid}, time, iid, 1)
+			else
+				send_warning("An interrupt ID was too large!")
+			end
+		end)
 	end
 	return interrupt
 end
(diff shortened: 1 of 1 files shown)


mesecons_luacontroller/init.lua
```diff
@@ -259,49 +260,69 @@ local function remove_functions(x)
...
-		-- Make sure serialized version of the data is not insanely long to
-		-- prevent DoS-like attacks
+		-- To previous commenter: NO NO NO NO NO! ADDED REALISM OR NOT, STRIPPING FUNCTIONS IS IMPORTANT FOR ANTI-DoS!
+		-- Anyway, the string sandbox gets disabled because external code is being called which ofc could use (""):y forms.
+		-- If we run out of time here and thus leave the sandbox off, out of time state must not allow user code to get called.
+		local onetruestring = getmetatable("")
+		onetruestring.__index = string
 		local msg_ser = minetest.serialize(msg)
-		if #msg_ser > mesecon.setting("luacontroller_digiline_maxlen", 50000) then
+		local setting = mesecon.setting("luacontroller_digiline_maxlen", 50000)
```
numberZero: Why is `mesecon.setting` called in the inner function?

mesecons_luacontroller/init.lua
```diff
@@ -474,6 +512,16 @@ mesecon.queue:add_function("lc_interrupt", function (pos, luac_id, iid)
 	run(pos, {type="interrupt", iid = iid})
 end)
 
+mesecon.queue:add_function("lc_digiline_relay", function (pos, channel, msg)
+	if not digiline then return end
+	-- This check is only really necessary because in case of server crash, old actions can be thrown into the future
+	if (minetest.get_meta(pos):get_int("luac_id") ~= luac_id) then return end
```
numberZero: Where is `luac_id` declared?

mesecons_luacontroller/init.lua
```diff
@@ -259,49 +260,69 @@ local function remove_functions(x)
...
 		if (type(channel) ~= "string" and type(channel) ~= "number" and type(channel) ~= "boolean") then
 			return false
 		end
 
-		-- It is technically possible to send functions over the wire since
-		-- the high performance impact of stripping those from the data has
-		-- been decided to not be worth the added realism.
-		-- Make sure serialized version of the data is not insanely long to
-		-- prevent DoS-like attacks
+		-- To previous commenter: NO NO NO NO NO! ADDED REALISM OR NOT, STRIPPING FUNCTIONS IS IMPORTANT FOR ANTI-DoS!
```
numberZero: Without history your comment is pointless.

mesecons_luacontroller/init.lua
```diff
@@ -259,49 +260,69 @@ local function remove_functions(x)
...
+		-- Anyway, the string sandbox gets disabled because external code is being called which ofc could use (""):y forms.
+		-- If we run out of time here and thus leave the sandbox off, out of time state must not allow user code to get called.
+		local onetruestring = getmetatable("")
+		onetruestring.__index = string
```
numberZero: I’d prefer some general gateway-like approach. May not fit the current code well, though.

mesecons_luacontroller/init.lua
```diff
@@ -259,49 +260,69 @@ local function remove_functions(x)
...
+		-- NOTE: This runs within string metatable sandbox, so don't *rely* on anything of the form (""):y
+		-- Hence the values get moved out. Should take less time than original, so totally compatible
 		if type(time) ~= "number" then return end
-		local luac_id = minetest.get_meta(pos):get_int("luac_id")
-		mesecon.queue:add_action(pos, "lc_interrupt", {luac_id, iid}, time, iid, 1)
+		table.insert(itbl, function ()
+			-- Outside string metatable sandbox, can safely run this now
+			local luac_id = minetest.get_meta(pos):get_int("luac_id")
+			-- Check if IID is dodgy, so you can't use interrupts to store an infinite amount of data.
+			-- Note that this is safe from alter-after-free because this code gets run after the sandbox has ended.
```
numberZero: Yes it can’t be made arbitrarily big, but still IID can be changed after call to `interrupt`.
BTW, didn’t know that tables could be used as IID.

mesecons_luacontroller/init.lua
20kdc: Relying on external functions in the string sandbox will probably result in hard to debug crashes.

mesecons_luacontroller/init.lua
20kdc: ...whoops! Apparently I screwed this up somehow. This was a rather last-minute change, I'll undo it.

mesecons_luacontroller/init.lua
20kdc: Neither did I. Also note that functions being put in `itbl` only run after the sandbox ends - I.E. after the code is unable to change it.

numberZero: >  I.E. after the code is unable to change it.

Yes I understand that. That is, there is no security problem. The problem I noticed is just inconvenience: it’s highly unobvious that the table is captured by reference, esp. remembering the fact `digiline_send` does the opposite.

mesecons_luacontroller/init.lua
```diff
@@ -259,49 +260,69 @@ local function remove_functions(x)
...
+			-- This runs outside of the timer and *shouldn't* harm perf. unless dodgy data is being sent in the first place
+			iid = remove_functions(iid)
+			local msg_ser = minetest.serialize(iid)
+			if #msg_ser > mesecon.setting("luacontroller_interruptid_maxlen", 256) then
+				return false
```
numberZero: Return value is not used anymore.

mesecons_luacontroller/init.lua
numberZero: Also now it may fail silently (if IID is too large), that’s not very good.

20kdc: That's more an unfortunate side effect of attempting to keep compatibility with `digiline_send`'s "return false if message too large" behavior, but also trying to avoid breaking compatibility by making the actual process of setting an interrupt use serialize.
Keep in mind last time someone tried these kinds of changes, they got reverted because of what amounted to event-counter compatibility (insane reason IK).

mesecons_luacontroller/init.lua
20kdc: ...oh, right. As for IID failing silently, not much I can do here without turning off the event counter and turning it on again (see previous note on compat.)

Hawk777:
You pinged me, but I’m not sure what you want from me. Are you hoping I’ll review this PR? Or something else? You mentioned that your changes conflict with my changes in other PRs, but AFAICT that’s neither your nor my business, but rather up to the powers that be which PR they decide to accept (if either), and then whoever’s left fixes up their leftovers afterwards.

Hawk777:
> However, if we don't want every single thing that uses Digilines to make this mistake I believe it would be better implemented there

IMO the table copying is better done in Luacontrollers, for a simple reason: *not* everything that uses Digilines is going to make a mistake. Far more likely IMO is that most things that send messages will create a fresh table for each message and then send it, and most things that receive messages will examine, but not modify, the received message (if we talk of fixed-functionality objects like chests and LCDs, where the code is in a mod and not user-editable). For those network endpoints, copying the table is a total waste of time.

20kdc:
Hawk777: Regarding first message: It's because if this PR gets closed without resolution, and given my track record with PRs that's kind of likely, I want to make sure you're up to date.
Regarding second message, fair enough.

mesecons_luacontroller/init.lua
numberZero: It may be better to complain loudly, e.g. showing an error message in the formspec. That may actually be a better (for players) way to report failures because it doesn’t require checking return value (the thing no one wants to do).

mesecons_luacontroller/init.lua
```diff
@@ -449,12 +473,21 @@ local function run(pos, event)
...
+	local itbl = {}
+	local env = create_environment(pos, mem, event, itbl)
 
 	-- Create the sandbox and execute code
 	local f, msg = create_sandbox(code, env)
 	if not f then return msg end
+	-- Start string true sandboxing
+	local onetruestring = getmetatable("")
+	if onetruestring.__index ~= string then
+		return "Cannot continue - string sandboxing appears nested?"
```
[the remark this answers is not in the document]
20kdc: Idea here was to fail safe, but good point, and will implement.

mesecons_luacontroller/init.lua
```diff
@@ -518,7 +609,7 @@ local digiline = {
 	receptor = {},
 	effector = {
 		action = function(pos, node, channel, msg)
-			run(pos, {type = "digiline", channel = channel, msg = msg})
+			run(pos, {type = "digiline", channel = channel, msg = table.copy(msg)})
```
Hawk777: `table.copy` is a shallow copy. GH-379 does a deep copy instead. If you do a shallow copy, I can just put a table in a table and use the inner table for my cheaty data sharing.

20kdc: Ah, didn't know it was a shallow copy. Will fix this.

20kdc: Uh, I just checked table.copy - it's *not* a shallow copy. It does preserve situations like:

```
local a = {}
local b = {
 fork1 = a,
 fork2 = a
}
```

but fork1 and fork2 are copied.

See: https://github.com/minetest/minetest/blob/master/builtin/common/misc_helpers.lua#L539

Hawk777: My mistake; I have no idea where I got that thought from.

20kdc: Don't worry about it. Helped me fix a bug anyway because I paid attention to table.copy.

numberZero:
@20kdc Rebase needed.

20kdc:
Rebased and squashed. Sorry for the wait.

20kdc:
@numberZero *pokes* I don't know if the previous notification was lost, so pinging directly.
As far as I can tell this PR is ready - have I missed anything?

numberZero:
Sorry, was too busy to review.
The code doesn’t look that well, but works well enough to be mergable.

PR #394, opened 2018-01-02
Improve settings

DS-Minetest:
Replaces and so closes #344.
Replaces and so closes #350.
- `minetest.settings` is used instead of the deprecated `minetest.setting_*` functions.
- There won't be crashes for old minetest versions because a fake `minetest.settings` is created for them. Ergo this PR can be merged earlier than #344.
- Settings are only loaded once per file (at load time).
- A `settingtypes.txt` file is added based on the one from #350 but improved.

I didn't test this yet perfectly.

### DIFF
--- a/mesecons/settings.lua
+++ b/mesecons/settings.lua
@@ -18,14 +18,8 @@ minetest.settings = minetest.settings or {
 	set_bool = function(_, k, v)
 		return minetest.setting_setbool(k, v)
 	end,
-	get_names = function()
-		return {}
-	end,
-	write = function()
-		return minetest.setting_save()
-	end,
 }
--- do not use get_np_group, set_np_group, remove or to_table
+-- do not use get_np_group, set_np_group, remove, get_names, write or to_table
 
 
 function mesecon.setting(setting, default)
@@ -33,6 +27,7 @@ function mesecon.setting(setting, default)
 		local read = minetest.settings:get_bool("mesecon."..setting, default)
 		if read == nil then -- legacy
 			return default
+		end
 		return read
 	elseif type(default) == "string" then
 		return minetest.settings:get("mesecon."..setting) or default
--- a/mesecons/settings.lua
+++ b/mesecons/settings.lua
@@ -1,15 +1,45 @@
 -- SETTINGS
+
+-- legacy:
+minetest.settings = minetest.settings or {
+	get = function(_, k)
+		return minetest.setting_get(k)
+	end,
+	get_bool = function(_, k, default)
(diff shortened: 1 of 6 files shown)


mesecons/settings.lua
```diff
@@ -1,15 +1,37 @@
...
+}
+-- do not use get_np_group, set_np_group, remove, get_names, write or to_table
+
+
 function mesecon.setting(setting, default)
 	if type(default) == "boolean" then
-		local read = minetest.setting_getbool("mesecon."..setting)
-		if read == nil then
+		local read = minetest.settings:get_bool("mesecon."..setting, default)
+		if read == nil then -- legacy
```
numberZero: That is, for early Settings implementation, right?

settingtypes.txt
```diff
@@ -0,0 +1,51 @@
...
+mesecon.luacontroller_maxevents (Max debug hook events per event) int 10000
+#    This is the maximum number of characters of serialized mem per luacontroller.
+mesecon.luacontroller_memsize (Memory size) int 100000
+
+[mesecons_movestones]
+#    This is the speed of movestones in nodes per second.
+mesecon.movestone_speed (Speed) int 3
+#    This is the maximum amount of nodes that can be pushed by a movestone.
+mesecon.movestone_max_push (Max push) int 50
+#    This is the maximum amount of nodes that can be pulled by a sticky movestone.
```
numberZero: This is this is this is this is...

mesecons_detector/init.lua
```diff
@@ -167,9 +170,8 @@ local function node_detector_scan(pos)
 	local meta = minetest.get_meta(pos)
 
 	local distance = meta:get_int("distance")
-	local distance_max = mesecon.setting("node_detector_distance_max", 10)
 	if distance < 0 then distance = 0 end
-	if distance > distance_max then distance = distance_max end
+	distance = math.min(distance, node_detector_distance_max)
```
numberZero: Why do you mix `if...then` with `math.min`?

mesecons/settings.lua
```diff
@@ -1,15 +1,37 @@
 -- SETTINGS
+
+-- legacy:
+minetest.settings = minetest.settings or {
```
numberZero: `local settings =`
Don’t alter MT namespace.

SmallJoker:
@DS-Minetest  Needs improvement and rebase.

numberZero:
Both #344 and #350 are merged, making this obsolete (except of maybe description texts?)

PR #397, opened 2018-01-09
Sticky block: Add texture, sounds and groups

Wuzzy2:
This PR does the following:

* Add a texture for the sticky block. It's a wooden box with glue
* Rename the node from “All-Faces Sticky Block” to “Sticky Block” because it is (currently) the only one. Can be reverted later, of course
* Add node sounds
* Add proper grouping (choppy, oddly_breakable_by_hand)

Preview of texture:
https://i.imgur.com/E29MeFO.png


C1ffisme:
The texture is a bit too yellow for my taste... Could you make it a bit more green?

VanessaE:
If it were up to me, it would be a plain-ish block, with glue painted onto just the middle, similar to a sticky piston.

Wuzzy2:
xD @VanessaE.
Actually, did exactly that. I basically just combined the glue texture (blurred) with the piston head (slight color change at the border).
But yeah, I can understand why you hate this texture. I'm not a huge fan either. :-(
@C1ffisme: “More green” doesn't make sense at all to me. The glue item is yellow, not green. Same for the sticky piston head.

VanessaE:
I dunno, to my eyes it looks like you spread the glue across the whole surface instead of just a glob in the center.

VanessaE:
better.

numberZero:
@Wuzzy2 

Also it’s fine that you don’t add crafting recipes as the block is still buggy.

Wuzzy2:
I added the new texture in my repo and squashed the commits.

PR #400, opened 2018-01-10
Limit digilines

numberZero:
Modified #379 
@Hawk777 @20kdc what’s your opinions on this variant?

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -292,6 +292,7 @@ local function safe_deep_copy_helper(msg, back_references, size_limit)
 			if k == nil or v == nil then
 				return nil, 0
 			end
+			ret[k] = v
 			cost = cost + k_cost + v_cost
 			if cost > size_limit then
 				return nil, 0
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -269,28 +269,60 @@ local function get_interrupt(pos)
 	return interrupt
 end
 
+local function safe_deep_copy_helper(msg, back_references, size_limit)
+	local t = type(msg)
+	if t == "string" then
+		return msg, #msg + 25
+	elseif t == "number" then
+		return msg, 8
+	elseif t == "boolean" then
+		return msg, 1
+	elseif t == "table" then
+		local bref = back_references[msg]
+		if bref then
+			return bref, 0
+		end
+		local cost = 8
+		local ret = {}
+		back_references[msg] = ret
+		for k, v in pairs(msg) do
+			local k_cost, v_cost
+			k, k_cost = safe_deep_copy_helper(k, back_references, size_limit)
+			v, v_cost = safe_deep_copy_helper(v, back_references, size_limit)
+			if k == nil or v == nil then
+				return nil, 0
+			end
+			cost = cost + k_cost + v_cost
+			if cost > size_limit then
+				return nil, 0
+			end
+		end
+		return ret, cost
+	else
+		return nil, 0
+	end
+end
+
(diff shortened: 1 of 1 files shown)


20kdc:
Something like this got committed to master by Jejia over a year ago.
See: https://github.com/minetest-mods/mesecons/commit/703e6fdadb5251b6f42e35f0f71f3094f5e15f75 and discussion there.
Then this happened... https://github.com/minetest-mods/mesecons/commit/54daee236e680fa79b2dd31e812f7e063887a0bf

20kdc:
...To clarify on the previous, I'm not saying this is a bad approach for this particular issue. I *am* saying I don't know if it'll stick around, given what happened last time this approach was chosen. That said, the previous approach used both serialize and something like this - perhaps this will be faster than serialize.

numberZero:
But I *replace* serialization test with a special function, so this should be actually even more performant than the original code.

numberZero:
Testing needed! Preferably with something as heavy as LTC-4000, does anyone have experience with it?

Hawk777:
At first glance this looks like a pure rewrite of #379, so what’s the difference?

20kdc:
I went and read through both, seems to me like it's based on PR 379 but with additions.

numberZero:
@Hawk777
1. It refuses to send anything containing disallowed data, instead of “cleaning up” an invalid message.
2. It returns early if the message appears to be too heavy.

mesecons_luacontroller/init.lua
```diff
@@ -269,28 +269,60 @@ local function get_interrupt(pos)
...
+		return msg, 8
+	elseif t == "boolean" then
+		return msg, 1
+	elseif t == "table" then
+		local bref = back_references[msg]
+		if bref then
+			return bref, 0
+		end
+		local cost = 8
+		local ret = {}
```
Hawk777: You never actually put anything in `ret`, do you?

mesecons_luacontroller/init.lua
```diff
@@ -269,28 +269,60 @@ local function get_interrupt(pos)
 	return interrupt
 end
 
+local function safe_deep_copy_helper(msg, back_references, size_limit)
+	local t = type(msg)
+	if t == "string" then
+		return msg, #msg + 25
```
Hawk777: A long string won’t be checked against `size_limit`.

Hawk777:
Meh. Seems like this is exactly identical to #379 if you were to say yes to my question at https://github.com/minetest-mods/mesecons/pull/379#discussion_r161133901, plus duplicating received tables. If you think it’s easier to make a whole new PR rather than just edit the existing one, go ahead.

mesecons_luacontroller/init.lua
numberZero: er, my mistake.

mesecons_luacontroller/init.lua
numberZero: it will. see `safe_deep_copy`

numberZero:
@Hawk777 Sorry for inconvenience, it’s easier to copy a PR and edit the copy, than to edit the PR branch directly.

mesecons_luacontroller/init.lua
numberZero: fixed.

Hawk777:
… but I was editing the branch all along, fixing your comments, without any trouble. Oh well.

mesecons_luacontroller/init.lua
Hawk777: Yes it will, I see now the size limit is checked in two places. Sorry about that.

numberZero:
Merged #379.

PR #418, opened 2018-07-18
Fix crash when error()ing an invalid value

### LINKED ISSUE #416: LuaController's error function called with unexpected argument type causes server crash.
```lua
error()
error(nil)
error(true)
error({})
```

beyondlimits:
Fixes #416


numberZero:
Thanks!

PR #419, opened 2018-07-18
Fix mesecon receiver duplication.

### LINKED ISSUE #414: Infinite mesecons glitch.
Applies to:
- MT 0.4.17 with MTG 0.4.17
- MT 0.5 with MTG 0.4.17
- MT 0.5 with MTG 0.5

1. Place a couple of mesecons in straight line and connect the lever or button to the end.
2. Dig the mesecon from the line.
3. Place the mesecon back.

At either step 2 or 3 (I don't remember), an itementity of mesecon will appear where there shouldn't be any.

MTG 0.4.17 seems to be the culprit, because after downgrading it to 0.4.16 I couldn't exploit the bug on MT 0.4.16, 0.4.17 and 0.5. What's odd, with MT 0.4.16 and MTG 0.4.17 I couldn't exploit the bug either.

numberZero:
Fix #414.
Dunno how could that code work in MT 0.4.
@beyondlimits

### DIFF
--- a/mesecons_receiver/init.lua
+++ b/mesecons_receiver/init.lua
@@ -209,7 +209,6 @@ function mesecon.receiver_place(rcpt_pos)
 	local param2 = minetest.dir_to_facedir(minetest.facedir_to_dir(node.param2))
 
 	if string.find(nn.name, "mesecons:wire_") ~= nil then
-		minetest.dig_node(pos)
 		minetest.set_node(pos, {name = rcvtype, param2 = param2})
 		mesecon.on_placenode(pos, nn)
 	end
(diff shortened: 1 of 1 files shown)


DS-Minetest:
https://github.com/minetest-mods/mesecons/commit/3c2d0ba325d4bcf6d56433cc33539eaea7dae86a#diff-2b7f5574880ed25568d13cf80aff893cR73
Those `dig_node`s were pretty old, what were they for? Should maybe dig callbacks be executed?

numberZero:
> Should maybe dig callbacks be executed?

No, I suppose, `update_autoconnect` is executed explicitly. You may experiment with various weird placement patterns, like wires adjacent to the button’s one but not connecting to it (as it has restricted rules).

PR #420, opened 2018-07-19
Microcontroller: Check protection access, fix globals assignment

SmallJoker:


SmallJoker:
I see but :man_shrugging: it's not worth splitting this commit for a few changed lines in the same file.

Issue #413, opened 2018-07-03
Movestone doesn’t drop falling nodes


Hawk777:
If a movestone has a falling node on top of it and moves out of the way, the falling node doesn’t fall. Note that if a sticky movestone pulls a node behind itself out of the way of a falling node, the falling node *does* fall, it’s just if *the movestone itself* used to be supporting the falling node, then it doesn’t fall. This is because, in `movestone_move`, `mvps_pull_all` calls `check_for_falling`, but the movestone itself is handled by an explicit `set_node` and `remove_node` and `check_for_falling` is not called.

This applies to both a horizontal movestone moving sideways out from under a falling node and a vertical movestone moving down with a falling node above it. It doesn’t apply to a vertical sticky movestone because that pulls its nodes rather than letting them fall.

numberZero:
@Hawk777 
Seems trivially fixable, can you make a PR, please?

PR #421, opened 2018-07-23
Let things fall behind movestones

Hawk777:
Fixes GH-413.

### DIFF
--- a/mesecons_movestones/init.lua
+++ b/mesecons_movestones/init.lua
@@ -60,13 +60,17 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 		minetest.get_node_timer(frontpos):start(timer_interval)
 
 		-- ### Step 3: If sticky, pull stack behind ###
-		if not is_sticky then
-			return
+		if is_sticky then
+			local backpos = vector.subtract(pos, direction)
+			success, stack, oldstack = mesecon.mvps_pull_all(backpos, direction, max_pull)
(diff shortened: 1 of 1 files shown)


numberZero:
Works for horizontal movestones, but doesn’t fix the case of a vertical one moving downwards.

mesecons_movestones/init.lua
```diff
@@ -60,13 +60,17 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
...
+				mesecon.mvps_move_objects(backpos, vector.multiply(direction, -1), oldstack, -1)
+			end
 		end
-		local backpos = vector.subtract(pos, direction)
-		success, stack, oldstack = mesecon.mvps_pull_all(backpos, direction, max_pull)
-		if success then
-			mesecon.mvps_move_objects(backpos, vector.multiply(direction, -1), oldstack, -1)
+
+		-- ### Step 4: Let things fall
+		if not is_vertical then
```
numberZero: The check is not necessary: if it moves upwards, or is sticky, there is nothing to fall.

numberZero:
Nvm, fixed myself.

DS-Minetest:
:thinking: Related: #371

DS-Minetest:
Sorry for confusing and for wasting others time.
I just thought if it would be better to do this check here for all mvps moves. But now I saw that the movestone is not moved per mvps.

Issue #424, opened 2018-07-23
Mod is not started


ne-vlezay80:
[OK] Mesecons
2018-07-23 22:00:54: ERROR[Main]: ModError: Failed to load and run script from /opt/minetest4/usr/local/bin/../share/minetest/games/heaes_revisited/mods/mesecons-master/mesecons_torch/init.lua:
2018-07-23 22:00:54: ERROR[Main]: ...visited/mods/mesecons-master/mesecons_torch/init.lua:83: attempt to perform arithmetic on field 'LIGHT_MAX' (a nil value)
2018-07-23 22:00:54: ERROR[Main]: stack traceback:
2018-07-23 22:00:54: ERROR[Main]: 	...visited/mods/mesecons-master/mesecons_torch/init.lua:83: in main chunk

sfan5:
Which Minetest version are you using?

ne-vlezay80:
0.4.17.1

PR #425, opened 2018-07-23
Fixed bug

ne-vlezay80:
For issue: #424


sfan5:
This is not the correct way to fix this bug, instead the code should attempt to read from `minetest.LIGHT_MAX`

PR #430, opened 2018-08-26
Fix vertical movestone textures

### LINKED ISSUE #423: Vertical sticky movestones have wrong textures
Actually, they use the same textures as horizontal ones, but that’s incorrect: glue should be on top and bottom sides.

numberZero:
Fix #423


numberZero:
Does the texture look okay?
screenshot_20180907_220528
It mixes existing textures so doesn’t require updating texture packs.

PR #437, opened 2018-09-17
Mark LuaController memory as private

luk3yx:
If LuaControllers handle sensitive information, hacked clients could get this information from the LuaController. Marking the memory as private fixes this and saves a small amount of bandwidth.


Billy-S:
I agree with this

numberZero:
:+1:

PR #441, opened 2018-11-29
Replace ABM functions with per-node timers

h-v-smacker:
Since mesecons are machines, and therefore much more rare than "natural" nodes, it makes sense to activate them using callback for node timer, as opposed to introducing wide-sweep ABM scans.
In total, there are 4 files with 9 registered ABMs: one for torch, two for the turbine and solar panel, and 4 for detectors. I replaced them with per-node timer events and an LBM to migrate existing nodes (activate their timers).

### DIFF
--- a/mesecons_torch/init.lua
+++ b/mesecons_torch/init.lua
@@ -60,12 +60,16 @@ local torch_update = function(pos)
 		end
 	end
 
-	if is_powered and node.name == "mesecons_torch:mesecon_torch_on" then
-		minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_off", param2 = node.param2})
-		mesecon.receptor_off(pos, torch_get_output_rules(node))
-	elseif node.name == "mesecons_torch:mesecon_torch_off" then
-		minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_on", param2 = node.param2})
-		mesecon.receptor_on(pos, torch_get_output_rules(node))
+	if is_powered then
+		if node.name == "mesecons_torch:mesecon_torch_on" then
+			minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_off", param2 = node.param2})
+			mesecon.receptor_off(pos, torch_get_output_rules(node))
+		end
+	else
+		if node.name == "mesecons_torch:mesecon_torch_off" then
+			minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_on", param2 = node.param2})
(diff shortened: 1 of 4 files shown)


mesecons_torch/init.lua
```diff
@@ -49,6 +49,28 @@ local torch_selectionbox =
...
+		end
+	end
+
+	if is_powered and node.name == "mesecons_torch:mesecon_torch_on" then
+		minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_off", param2 = node.param2})
+		mesecon.receptor_off(pos, torch_get_output_rules(node))
+	elseif node.name == "mesecons_torch:mesecon_torch_off" then
+		minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_on", param2 = node.param2})
+		mesecon.receptor_on(pos, torch_get_output_rules(node))
+	end
```
DS-Minetest: You made a mistake here:

`is_powered` | `node.name` | what happens | what should happen / happened before
---|---|---|---
`false` | `"mesecons_torch:mesecon_torch_on"` | nothing happens | nothing happens
`false` | `"mesecons_torch:mesecon_torch_off"` | torch is activated | torch is activated
`true`  | `"mesecons_torch:mesecon_torch_on"` | torch is deactivated | torch is deactivated
`true`  | `"mesecons_torch:mesecon_torch_off"` | torch is activated | nothing happens

The torch will probably always go off and on if it's powered.

h-v-smacker: Pardon me, but I fail to see how it's different from the previous condition structure which is used in abm:
```
if is_powered then
	if node.name == "mesecons_torch:mesecon_torch_on" then
		minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_off", param2 = node.param2})
		mesecon.receptor_off(pos, torch_get_output_rules(node))
	end
elseif node.name == "mesecons_torch:mesecon_torch_off" then
	minetest.swap_node(pos, {name = "mesecons_torch:mesecon_torch_on", param2 = node.param2})
	mesecon.receptor_on(pos, torch_get_output_rules(node))
end
```
the inner condition in the first part (if node.name == "mesecons_torch:mesecon_torch_on") was considered only when the outermost (if is_powered) was also true. In that respect, joining them with an "and" operator should change nothing.

I also tested the behavior of the nodes both on master and my branch, and they appear to be giving the same results. I did not check if they are giving the same results as theoretically expected, of course... Just moved the function from abm to 1 Hz node timer.

PR #448, opened 2018-12-10
Actionqueue timing

thomasrudin:
Adds optional actionqueue timing, useful for server owners, disabled by default:

Settings:
```
# enable timing logging
mesecon.actionqueue_timing_log = true

# threshold for whole actionqueue step, in microseconds, default: 100'000 / 100 ms
mesecon.actionqueue_timing_step_threshold = 100000

# threshold for single action timing, in microseconds, default: 10'000 / 10 ms
mesecon.actionqueue_timing_action_threshold = 10000
```

### DIFF
--- a/mesecons/actionqueue.lua
+++ b/mesecons/actionqueue.lua
@@ -82,8 +82,20 @@ minetest.register_globalstep(function (dtime)
 
 	while(#actions_now > 0) do -- execute highest priorities first, until all are executed
 		local hp = get_highest_priority(actions_now)
-		mesecon.queue:execute(actions_now[hp])
+		local action = actions_now[hp]
+
+		local ts0 = minetest.get_us_time()
+		mesecon.queue:execute(action)
 		table.remove(actions_now, hp)
+		local ts1 = minetest.get_us_time()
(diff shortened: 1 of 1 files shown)


mesecons/actionqueue.lua
```diff
@@ -56,13 +56,22 @@ local get_highest_priority = function (actions)
 	return highesti
 end
 
+-- enable timing log
+local enable_timing_log = mesecon.setting("actionqueue_timing_log", false)
+-- threshold for globalstep timing warning
+local step_log_threshold = mesecon.setting("actionqueue_timing_step_threshold", 100000) -- 100 ms
+-- threshold for action timing warning
+local action_log_threshold = mesecon.setting("actionqueue_timing_action_threshold", 10000) -- 10 ms
```
numberZero: Instead of putting comments, rename the variables (and settings keys) to something more meaningful, that won’t make them longer.

thomasrudin:
The longer i think about this PR the more i don't like it, closing this as the use-case is really very slim (mostly mesecon performance edge-cases)
I don't want to clutter this mod with (almost) useless timing-statements

PR #452, opened 2019-01-03
Improve recipes

### LINKED ISSUE #316: LuaControllers are way too cheap
3 sand blocks, 1 steel ingot, and 2/9 mese crystal... Well, 2 NAND gates cost more mese...
Suggestion 1: replace 1 or 2 silicon items with mese crystals in the LuaCtl recipe.
Suggestion 2: replace steel with mese in the silicon recipe itself.

numberZero:
Fix #316 
Note that the conductors are still relatively cheap, only the logic is more expensive.

### DIFF
--- a/mesecons_gates/init.lua
+++ b/mesecons_gates/init.lua
@@ -104,40 +104,45 @@ local function register_gate(name, inputnumber, assess, recipe, description)
 	minetest.register_craft({output = basename.."_off", recipe = recipe})
 end
 
+local _ = ""
+local m = "group:mesecon_conductor_craftable"
+local t = "mesecons_torch:mesecon_torch_on"
+local s = "mesecons_materials:silicon"
+
 register_gate("diode", 1, function (input) return input end,
-	{{"mesecons:mesecon", "mesecons_torch:mesecon_torch_on", "mesecons_torch:mesecon_torch_on"}},
+	{{m, t, t}},
 	"Diode")
 
 register_gate("not", 1, function (input) return not input end,
-	{{"mesecons:mesecon", "mesecons_torch:mesecon_torch_on", "mesecons:mesecon"}},
+	{{m, t, m}},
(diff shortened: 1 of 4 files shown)


DS-Minetest:
Why is there still an output of two luacontrollers at once? I suggest one.

I'm not sure if I like the change to silicium. Don't forget that other items like the solar panel use it, too.

numberZero:
Solar panels become expensive, just like IRL... Can’t imagine a situation where you need many solar panels; they are detectors and not power sources here. Can increase the recipe output, however, that won’t harm.

In my opinion, gates are more important; not that they become more expensive (that should be OK), but that the cost is different (2 silicon pieces for the XOR gate and none for NOR). What’s about such recipes?
```
Diode    NOT     AND    NAND     XOR     NOR      OR
. . .   . . .   m t .   m t .   m t .   m . .   m . .
m t t   m t m   . s t   . s m   t s t   . s t   . s m
. . .   . . .   m t .   m t .   m t .   m . .   m . .
```
(**m**esecon, **t**orch, **s**ilicon)
Note that a pair of diodes may be used instead of an OR gate, and a pair of NOT gates replaces NAND gate (actually that’s the easy way if you need more that 2 inputs).

numberZero:
Updated. The recipes are now:
```
Diode    NOT     AND    NAND     XOR     NOR      OR
. . .   . . .   m t .   m t .   m t .   m . .   m . .
m t t   m t m   . s t   . s m   s s t   . s t   . s m
. . .   . . .   m t .   m t .   m t .   m . .   m . .
```

SmallJoker:
@numberZero Is there something left to do, or can this PR be merged now?

sfan5:
The recipes in the documentation will need to be updated too:
https://github.com/minetest-mods/mesecons/blob/master/mesecons_gates/doc/and/recipe.png

PR #455, opened 2019-02-10
Allow Luacontroller programs to request the use of lightweight interrupts

cheapie:
This adds a third parameter to the LuaC's interrupt() function, allowing the program running in the LuaC to request that lightweight interrupts be used even if they're not forced on that server.

If this value is set to true:
* Lightweight interrupts are used, regardless of the server setting

If this value is set to false:
* On servers set to use lightweight interrupts, this gives the error "Non-lightweight interrupts are disabled on this server"
* On servers not set to use lightweight interrupts, traditional "non-lightweight" interrupts are used.

If this value is set to nil or otherwise not specified:
* The interrupt type follows the server's mesecon.luacontroller_lightweight_interrupts setting.

Note that since this is a per-interrupt setting, this allows, for instance, some interrupts to be lightweight and some not if desired by the person programming the Luacontroller and allowed by the server's configuration. This can be used, for example, for machines that perform both important and unimportant tasks.

Examples:
```
interrupt(5)
```
Sets an interrupt for 5 seconds, with the type depending on the server's configuration.

```
interrupt(5,"blah")
```
Sets an interrupt for 5 seconds, with the type depending on the server's configuration. If the server is set to use lightweight interrupts, then the usual warning about IIDs not being supported will be issued.

```
interrupt(5,"blah",false)
```
On servers that use lightweight interrupts, an error will be shown. On servers that do not, a non-lightweight interrupt will be set for 5 seconds.

```
interrupt(5,nil,true)
```
A lightweight interrupt will be set for 5 seconds, regardless of whether the server is set to use lightweight interrupts or not.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -271,8 +271,9 @@ local get_interrupt
 if mesecon.setting("luacontroller_lightweight_interrupts", false) then
 	-- use node timer
 	get_interrupt = function(pos, itbl, send_warning)
-		return (function(time, iid)
+		return (function(time, iid, lightweight)
 			if type(time) ~= "number" then error("Delay must be a number") end
+			if lightweight == false then error("Non-lightweight interrupts are disabled on this server") end
 			if iid ~= nil then send_warning("Interrupt IDs are disabled on this server") end
 			table.insert(itbl, function() minetest.get_node_timer(pos):start(time) end)
 		end)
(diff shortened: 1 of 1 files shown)


mesecons_luacontroller/init.lua
```diff
@@ -282,24 +283,29 @@ else
...
-				iid = remove_functions(iid)
-				local msg_ser = minetest.serialize(iid)
-				if #msg_ser <= mesecon.setting("luacontroller_interruptid_maxlen", 256) then
-					mesecon.queue:add_action(pos, "lc_interrupt", {luac_id, iid}, time, iid, 1)
-				else
-					send_warning("An interrupt ID was too large!")
-				end
-			end)
+			if lightweight then
+				if iid ~= nil then send_warning("Interrupt IDs cannot be used with lightweight interrupts") end
```
numberZero: This should probably be an error here, as the lightweight mode is requested explicitly.

mesecons_luacontroller/init.lua
```diff
@@ -271,8 +271,9 @@ local get_interrupt
 if mesecon.setting("luacontroller_lightweight_interrupts", false) then
```
numberZero: This condition should probably be moved inside the function then. The value may be cached (`local force_lightweight_interrupts = ...`), though.

thomasrudin:
any progress on this?

cheapie:
I ended up just making an alternative (digilines timer in digistuff) instead of trying to make the mesecons devs happy. I forgot to close this.

thomasrudin:
Love yor digistuff mod btw :+1:

PR #458, opened 2019-03-08
Revert "Update init.lua" (getpos->get_pos)

DS-Minetest:
Remove an annoying error message.

I think, it's time to think about compatibility to older minetest versions.
As minetest 5.0 is released, sooner or later everyone will use the new version to be able to connect to the majority of the servers. This means that there won't be anyone who would benefit from the compatibility with older versions.
I suggest:
1. For the next few months keep compatibility with 0.4.17.1, not to older versions.
2. Make a release or a tag or something the like.
3. Only keep compatibility with 5.0+.
4. **Profit** from easier to maintain and less warnings screaming code.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -217,7 +217,7 @@ function mesecon.mvps_move_objects(pos, dir, nodestack, movefactor)
 	movefactor = movefactor or 1
 	dir = vector.multiply(dir, movefactor)
 	for id, obj in pairs(minetest.object_refs) do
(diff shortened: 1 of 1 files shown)


VanessaE:
Or just dispense with 0.4.x compatibility altogether, as needed.  As 5.0.0 clients can't communicate with 0.4.x servers, or vice versa, people will be forced to update much faster than with previous 0.4.x -> 0.4.(x+1) changes. 

So to my mind, there's no reason for backward compatibility.

PR #465, opened 2019-06-21
[WIP] add luacontroller environment customization function

thomasrudin:
the luacontroller environment can be customized in dependent mods
by overriding the mesecon.luacontroller.customize_environment(env, pos, mem) function
providing dependent mods the ability to add more functions to
the luacontroller enviromment (use-case, for example: mob-control, chat, etc)

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -438,9 +438,14 @@ local safe_globals = {
 }
 
 -- luacontroller environment customization
--- can be overridden in dependent mods to add more functions
-function mesecon.luacontroller.customize_environment(env, pos, mem)
-	-- no-op
+
+-- list of customization callbacks
+-- table with function(env, pos, name) entries
+local customization_callbacks = {}
+
+-- register a luacontroller customization callback in the form of function(env, pos, name)
+function mesecon.luacontroller.register_customization(callback)
+	table.insert(customization_callbacks, callback)
 end
 
 local function create_environment(pos, mem, event, itbl, send_warning)
@@ -525,8 +530,10 @@ local function create_environment(pos, mem, event, itbl, send_warning)
 		env[name] = _G[name]
 	end
 
-	-- customization hook
-	mesecon.luacontroller.customize_environment(env, pos, mem)
+	-- run customization hooks
+	for customization_callback in ipairs(customization_callbacks) do
+		customization_callback(env, pos, mem)
+	end
 
 	return env
 end
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -24,8 +24,7 @@
 -- a very restricted environment.
 -- Actually the only way to damage the server is to
 -- use too much memory from the sandbox.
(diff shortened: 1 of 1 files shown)


numberZero:
Nice idea. Wouldn’t callback list be better? (like `minetest.register_on...`)
Also some documentation, reminding that all the code mods provide runs under sandboxing, may be interrupted at any time, must propagate errors, etc.

thomasrudin:
Will do, this is just an initial proposal.
Maybe a `mesecons.luacontroller.register_environment()` would be nicer than
just overriding the existing function.

mesecons_luacontroller/init.lua
```diff
@@ -24,8 +24,7 @@
 -- a very restricted environment.
 -- Actually the only way to damage the server is to
 -- use too much memory from the sandbox.
--- You can add more functions to the environment
--- (see where local env is defined)
+-- You can add more functions to the environment if you override mesecon.luacontroller.customize_environment
```
numberZero: no longer true

mesecons_luacontroller/init.lua
```diff
@@ -433,6 +437,17 @@ local safe_globals = {
 	"tonumber", "tostring", "type", "unpack", "_VERSION"
 }
 
+-- luacontroller environment customization
+
+-- list of customization callbacks
+-- table with function(env, pos, name) entries
+local customization_callbacks = {}
+
+-- register a luacontroller customization callback in the form of function(env, pos, name)
```
numberZero: Description of sandbox peculiarities is needed. Also `itbl` may be very useful. (And, again, that’s not `name` but `mem`. Is it really necessary? The callback could access it from `env`)

mesecons_luacontroller/init.lua
```diff
@@ -515,6 +530,11 @@ local function create_environment(pos, mem, event, itbl, send_warning)
 		env[name] = _G[name]
 	end
 
+	-- run customization hooks
+	for customization_callback in ipairs(customization_callbacks) do
```
numberZero: `_, customization_callback` as `ipairs` yields pairs, not values

mesecons_luacontroller/init.lua
```diff
@@ -433,6 +437,17 @@ local safe_globals = {
 	"tonumber", "tostring", "type", "unpack", "_VERSION"
 }
 
+-- luacontroller environment customization
+
+-- list of customization callbacks
+-- table with function(env, pos, name) entries
+local customization_callbacks = {}
```
numberZero: `name`?

DS-Minetest:
Imo, overriding a function isn't that bad, and it would be much simpler, but idk.

If you do register_... stuff, you should use the builtin run_callbacks api and set the mod origin.

thomasrudin:
> I probably need to write that myself as the sandboxing machinery is rather obscure.

This was an idea of mine and a work in progress, but if you (@numberZero) feel you can do it better i have no objections if you implement it differently...

numberZero:
@thomasrudin I meant the docs, not the code =)

PR #466, opened 2019-06-27
Respect protection in MVPS

### LINKED ISSUE #281: Pistons move nodes into protected areas

numberZero:
Fix #281 
Seems to work well, testing needed. @LoneWolfHT @DS-Minetest

### DIFF
--- a/mesecons_movestones/init.lua
+++ b/mesecons_movestones/init.lua
@@ -56,6 +56,8 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 
 		-- ### Step 2: Move the movestone ###
 		minetest.set_node(frontpos, node)
+		local meta2 = minetest.get_meta(frontpos)
+		meta2:from_table(meta:to_table())
 		minetest.remove_node(pos)
 		mesecon.on_dignode(pos, node)
 		mesecon.on_placenode(frontpos, node)
--- a/mesecons_movestones/init.lua
+++ b/mesecons_movestones/init.lua
@@ -49,6 +49,10 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 		-- ### Step 1: Push nodes in front ###
 		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, max_push, owner)
 		if not success then
+			if stack == "protected" then
+				meta:set_string("infotext", "Can't move: protected area on the way")
+				return
+			end
 			minetest.get_node_timer(pos):start(timer_interval)
 			return
 		end
@@ -57,7 +61,7 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 		-- ### Step 2: Move the movestone ###
 		minetest.set_node(frontpos, node)
 		local meta2 = minetest.get_meta(frontpos)
-		meta2:from_table(meta:to_table())
+		meta2:set_string("owner", owner)
 		minetest.remove_node(pos)
 		mesecon.on_dignode(pos, node)
 		mesecon.on_placenode(frontpos, node)
@@ -87,7 +91,15 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 		rules = mesecon.rules.default,
 	}}
 
-	def.after_place_node = mesecon.mvps_mark_owner
+	def.after_place_node = mesecon.mvps_set_owner
+
+	def.on_punch = function(pos, node, player)
+		local player_name = player and player.get_player_name and player:get_player_name()
+		if mesecon.mvps_claim(pos, player_name) then
+			minetest.get_node_timer(pos):start(timer_interval)
+			minetest.chat_send_player(player_name, "Reclaimed movestone")
+		end
+	end
 
 	def.on_timer = function(pos, elapsed)
 		local sourcepos = mesecon.is_powered(pos)
--- a/mesecons_movestones/init.lua
+++ b/mesecons_movestones/init.lua
@@ -43,9 +43,11 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 	local function movestone_move(pos, node, rulename)
 		local direction = get_movestone_direction(rulename, is_vertical)
 		local frontpos = vector.add(pos, direction)
+		local meta = minetest.get_meta(pos)
+		local owner = meta:get_string("owner")
 
 		-- ### Step 1: Push nodes in front ###
-		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, max_push)
+		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, max_push, owner)
 		if not success then
 			minetest.get_node_timer(pos):start(timer_interval)
 			return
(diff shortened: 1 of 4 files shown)


mesecons_mvps/init.lua
```diff
@@ -133,31 +133,89 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
...
 end
 
-function mesecon.mvps_push(pos, dir, maximum)
-	return mesecon.mvps_push_or_pull(pos, dir, dir, maximum)
+function mesecon.mvps_mark_owner(pos, placer)
+	local meta = minetest.get_meta(pos)
+	local owner = placer and placer.get_player_name and placer:get_player_name()
+	if owner and owner ~= "" then
+		meta:set_string("owner", owner)
+		meta:set_string("infotext", "Owned by " .. owner)
```
DS-Minetest: This might be confusing because some players might thinks that the ownership protects pistons.

Also I'm against an infotext for pistons.

mesecons_mvps/init.lua
```diff
@@ -133,31 +133,89 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
...
+		elseif mode == "restrict" then
+			return true
+		else
+			error("Invalid protection mode")
+		end
+	end
+	local is_protected = minetest.is_protected
+	for _, pos in pairs(positions) do
+		if is_protected(pos, name) then
+			minetest.record_protection_violation(pos, player_name)
```
DS-Minetest: This might happen very often and automatically. And a player can use the piston of someone else. I'm not sure whether it makes sense to record protection violation here.

DS-Minetest:
This concept probably works nicely.
And code looks mostly good.

Imo the default value for `mesecon.mvps_protection_mode` should be `compat`. New servers won't have a problem with this. And old servers probably care more about old builds suddenly not working anymore. And if a player wants to use this to attack a protected structure, they would have to somehow get a mvps mover to the protected structure first which is takes much effort.

mesecons_mvps/init.lua
numberZero: I don’t think that’s the best solution, but the owner should be mentioned IMO. The problem is that if e.g. Alice helps Bob building some piston machine, and then Bob tries to protect it, it will stop working. When there is an infotext, Bob can see that pistons aren’t his, and replace them easily.

Instead of replacing, reclaiming (by left- or right-click) may be supported; that will also make abusing unowned pistons incredibly hard, as one accidental click will make the piston owned by the clicker (if not protected ofc). But I see some problems here, like placing a piston just outside the protected area, waiting for that accidental click...

HybridDog: It is unlikely that a player protects a piston which somebody else placed. You could set the infotext to an error message when the protection check fails so that Bob can see the problem when it appears.
This way you can do the reclaiming only when the error infotext occurs and send a chat message to the player who claimed the piston.

numberZero: Thanks

> send a chat message to the player who claimed the piston.

Sorry, don’t understand. When and whom should the message be sent?

HybridDog: > When and whom should the message be sent?

When the error message is set in the infotext, the player who owns the area can punch the piston to claim it. If he/she claims the piston, the infotext is removed and a chat message is sent to him/her so that this player is informed about the change. The previous "owner" does not receive a chat message.

For the situation where many pistons get protected by somebody, a tool or chatcommand could be added which allows claiming the pistons when they do not have the error message in the infotext; with this the player does not need to send a mesecons signal to all the pistons before claiming.

numberZero:
Updated.

mesecons_movestones/init.lua
```diff
@@ -43,17 +43,25 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
...
 		local frontpos = vector.add(pos, direction)
+		local meta = minetest.get_meta(pos)
+		local owner = meta:get_string("owner")
 
 		-- ### Step 1: Push nodes in front ###
-		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, max_push)
+		local success, stack, oldstack = mesecon.mvps_push(frontpos, direction, max_push, owner)
 		if not success then
+			if stack == "protected" then
+				meta:set_string("infotext", "Can't move: protected area on the way")
```
Panquesito7: Add a dot on the end.
I think it'll look better.

mesecons_movestones/init.lua
```diff
@@ -83,6 +91,16 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 		rules = mesecon.rules.default,
 	}}
 
+	def.after_place_node = mesecon.mvps_set_owner
+
+	def.on_punch = function(pos, node, player)
+		local player_name = player and player.get_player_name and player:get_player_name()
+		if mesecon.mvps_claim(pos, player_name) then
+			minetest.get_node_timer(pos):start(timer_interval)
+			minetest.chat_send_player(player_name, "Reclaimed movestone")
```
Panquesito7: Same here as above.

mesecons_mvps/init.lua
```diff
@@ -133,31 +133,100 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
...
 
-function mesecon.mvps_pull_all(pos, dir, maximum)
-	return mesecon.mvps_push_or_pull(pos, vector.multiply(dir, -1), dir, maximum, true)
+function mesecon.mvps_claim(pos, player_name)
+	if not player_name or player_name == "" then return end
+	local meta = minetest.get_meta(pos)
+	if meta:get_string("infotext") == "" then return end
+	if meta:get_string("owner") == player_name then return end -- already owned
+	if minetest.is_protected(pos, player_name) then
+		minetest.chat_send_player(player_name, "Can't reclaim: protected")
```
Panquesito7: Same here.

mesecons_pistons/init.lua
```diff
@@ -237,6 +243,13 @@ local function piston_rotate_pusher(pos, node, player, mode)
 	return piston_rotate_on(piston_pos, piston_node, player, mode)
 end
 
+local function piston_punch(pos, node, player)
+	local player_name = player and player.get_player_name and player:get_player_name()
+	if mesecon.mvps_claim(pos, player_name) then
+		minetest.chat_send_player(player_name, "Reclaimed piston")
```
Panquesito7: Same here.

settingtypes.txt
```diff
@@ -28,6 +28,17 @@ mesecon.luacontroller_memsize (Controller memory limit) int 100000 10000 1000000
...
+
+# In pre-existing world, MVPS may not be labelled with the owner.
+# Protection handling for them is configurable.
+# - normal: allow legacy MVPS to work in unprotected areas only
+# - compat: allow legacy MVPS to work everywhere
+# - ignore: allow all MVPS to work everywhere
+# - restrict: disallow legacy MVPS
+# Note that new unowned (e.g. machine-placed) MVPS are always
+# handled as in `normal` mode.
+mesecon.mvps_protection_mode (MVPS [movestones, pistons] protection handling) enum compat normal,compat,ignore,restrict
```
Panquesito7:
```suggestion
mesecon.mvps_protection_mode (MVPS [movestones, pistons] protection handling) enum compat normal, compat, ignore, restrict
```

numberZero: nice joke)

PR #472, opened 2019-08-12
Add digiline command to set node detector distance

coil0:
This adds a table-based digiline API to the node detector:

```lua
{
    command = "distance",
    distance = 5
}
```

to set distance to 5 (strings also work),

```lua
{
    command = "get"
}
```

to get node name, and

```lua
{
    command = "scanname",
    name = "default:acacia_wood"
}
```

to scan for acacia wood planks. Sending "GET" is identical to `{command = "get"}` and sending "default:acacia_wood" is identical to `{command = "scanname", name = "default:acacia_wood"}`, so this is compatible with existing machines. The novelty is the command to set distance.

### DIFF
--- a/mesecons_detector/init.lua
+++ b/mesecons_detector/init.lua
@@ -189,28 +189,49 @@ local function node_detector_scan(pos)
 		(frontname ~= "air" and frontname ~= "ignore" and scanname == "")
 end
 
+local function node_detector_send_node_name(pos, node, channel, meta)
+	local distance = meta:get_int("distance")
+	local distance_max = mesecon.setting("node_detector_distance_max", 10)
+	if distance < 0 then distance = 0 end
+	if distance > distance_max then distance = distance_max end
+	local nodename = minetest.get_node(
+		vector.subtract(pos, vector.multiply(minetest.facedir_to_dir(node.param2), distance + 1))
+	).name
+
+	digiline:receptor_send(pos, digiline.rules.default, channel, nodename)
+end
+
 -- set player name when receiving a digiline signal on a specific channel
 local node_detector_digiline = {
 	effector = {
 		action = function(pos, node, channel, msg)
 			local meta = minetest.get_meta(pos)
 
-			local distance = meta:get_int("distance")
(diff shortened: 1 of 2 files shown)


mesecons_detector/init.lua
```diff
@@ -189,28 +189,43 @@ local function node_detector_scan(pos)
 		(frontname ~= "air" and frontname ~= "ignore" and scanname == "")
 end
 
+local function send_node_name(pos, node, channel, meta)
```
DS-Minetest: I suggest renaming the function to `node_detector_send_node_name`.

DS-Minetest:
Please also update https://github.com/minetest-mods/mesecons/blob/master/mesecons_detector/doc/nodedetector/description.html. (It is used here: http://mesecons.net/items.html)

DS-Minetest:
(It might be useful to be able to scan for `param2`, too. But that's for another PR.)

I wonder if something like `{ command = "get", distance = 3}` should be possible, too.

numberZero:
Alternatively, it could just read table fields, so e.g. `{distance = 5}` would set the distance, and `{scanname="default:stone", distance=0}` would set both parameters. In this scheme, `get` might be passed the same way, or just be assumed either for `{}` or for any table.

coil0:
I prefer explicit commands. That leaves room for adding more later, and it should be possible both to set parameters without getting a response and to set parameters and get a response with a single message.
```lua
{
    distance = 5,
    scanname = "default:acacia_wood"
}
```
sets distance to 5 and scans for acacia wood,
```lua
{
    command = "get",
    distance = 5
}
```
sets distance to 5 and, using the new distance, responds with the node name,
```lua
{
    command = "scan"
}
```
responds with a boolean identical to the mesecon signal. Setting parameters is separate from any command specified, but happens before the command.

DS-Minetest:
Code looks good. :+1:

numberZero:
@coil0 Next time, please don’t amend commits you pushed already. Force-push is a last resort for things like rebase, while squashing is done at merge, if necessary (MT core had different rules IIRC, but they are relaxed now. Linear history is for SVN).

PR #476, opened 2019-08-23
Add translation support and translate into German

Wuzzy2:
This PR makes Mesecons translatable and translates all mods into German.

Special changes:
- Interface of Luacontroller was changed, the baked-in text was stripped from the image and replaced by real text so it can be translated
- Remove all “you hacker you” strings. These are not helpful and just annoying if you somehow get one of these items due to a bug

I did some rudimentary testing, but I advice you do some tests yourselves.


sfan5:
Looks good from a quick look, maybe split this into 2 PRs (translation support, DE translation) for easier reviewability.

Wuzzy2:
Replaced by #503.

PR #477, opened 2019-08-27
Allow corners and t-junctions to be crafted anywhere in the grid

mbartlett21:


DS-Minetest:
I've tested current master. The corners and t-junctions can already be crafted anywhere in the grid.
Nonetheless, this is a **trivial** code style improving little PR.

Issue #462, opened 2019-04-25
Piston drags a player


za267:
When pulling blocks back to 'open a door' for example, sticky pistons seem to drag a player when it retracts, bringing the player inside a wall, or in some cases even behind it.  Is that normal behaviour?

Thanks!

DS-Minetest:
That the player is pulled is intended but he shouldn't end up in a node.

za267:
There are two situations where that effect is rather unintended.  On my server I have a space port where I open up a 'door' where basically blocks are pulled out of the way by sticky pistons.  If the player is near the blocks they get dragged into the wall. Sometimes even behind the wall and fall into the 'mesecons' wiring in the back.  

Another situation is where we tried to make a "trap" that uses 8 pistons to pull the floor out from underneath a player when it is detected.  The player simply gets caught in the machinery instead.

DS-Minetest:
Before I write anything else, I should probably ask if you are really sure, that you have the newest mesecons version installed. So, have you?
(The version from contentdb might be a bit outdated.)

za267:
I'll get back to you on that one.  It's recent but I better double check.

DS-Minetest:
Then test again, and if it didn't help, try replacing
https://github.com/minetest-mods/mesecons/blob/b0158f5674433e4303219205d866789c1e5d8282/mesecons_mvps/init.lua#L250
with
```lua
local obj_offset = dir_l * movefactor * (obj_pos[dir_k] - pos[dir_k])
```
.

DS-Minetest:
Ahhh, now I see the problem.

DS-Minetest:
\>_< mvps pulling has multiple problems.
- Here is not checked, whether the player is really in the stack or just next to it: https://github.com/minetest-mods/mesecons/blob/b0158f5674433e4303219205d866789c1e5d8282/mesecons_mvps/init.lua#L252-L253
- If the player is in front of a sticky piston, he will only be pulled if it's westwards (-X).
- …

benrob0329:
Pushing/pulling has been a little buggy for a long time (there is a reason I have to use a minecart on fast elevators)

za267:
Thanks both of you for the update.  I might give a shot at making a small change.

HybridDog:
The new `add_player_velocity` method could be used to push and pull a player. 
https://github.com/minetest/minetest/commit/cf64054390970f3cf974afb2b174340a3e1da382

PR #478, opened 2019-09-01
Fix players being dragged along the side of stacks

cheapie:
This fixes bug #462, where players alongside a stack of nodes being moved via MVPS would be dragged along with it, ending up inside a node if one was present there.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -225,8 +225,8 @@ function mesecon.mvps_move_objects(pos, dir, nodestack, movefactor)
 		for k, v in pairs(pos) do
 			local edge1, edge2
 			if k ~= dir_k then
-				edge1 = v - 0.51 -- More than 0.5 to move objects near to the stack.
-				edge2 = v + 0.51
+				edge1 = v - 0.499 -- Slightly less than 0.5 so that players next to the stack don't get dragged with it
+				edge2 = v + 0.499
 			else
 				edge1 = v - 0.5 * dir_l
 				edge2 = v + (#nodestack + 0.5 * movefactor) * dir_l


benrob0329:
This breaks mechanics and designs that I use in many places.

cheapie:
It fixes a bug you relied upon where pistons push things that aren't even in front of them, yes.

DS-Minetest:
This was done to keep a feature: https://github.com/minetest-mods/mesecons/pull/367#discussion_r139297935

benrob0329:
That "bug" is what makes practical piston elevators and piston-bolts possible in Minetest, as well as reasonably moving carts around.

cheapie:
> This was done to keep a feature: [#367 (comment)](https://github.com/minetest-mods/mesecons/pull/367#discussion_r139297935)

I can certainly attempt to re-add it in the +Y direction (so that objects on top slide with while objects that are just next to the stack don't) if you think there is interest in that. I'm not sure this is all that realistic given the instant nature of the movement - inertia would keep anything resting on top in place - but I suppose realism isn't always the goal here.

Another possibility would be to use a smaller radius for players than for other objects - this would still fix the original bug as reported, where players get pulled with a stack (and into a node if present), without removing the buggy behavior for other objects if this is what is desired. If there is any support for this, I'll either modify this PR or close it and create a new one as appropriate.

> That "bug" is what makes practical piston elevators and piston-bolts possible with mesecons, as well as reasonably moving carts around.

I have no idea what a "piston-bolt" is, but there are certainly piston elevator designs that do not rely on bugs. I've even shown you one of those on VE-Creative (and you seem to like nocontroller challenges, how about a nocontroller version of that one?)
I also fail to see how moving carts around in the way you have shown is anything but a bug - carts are supposed to be on rails, not shoved off to the side at random.

PR #479, opened 2019-09-09
Add sandstone pressure plate

frukto:
Just add another pressure plate. The textures are based on the original default sandstone.


DS-Minetest:
I doubt that it's justified to add a sandstone pressure plate while there are many more nodes that could be made to a pressure plate, too (eg. desert stone).
Note: You can also add new pressure plates in an additional mod.

>The textures are based on the original default sandstone.

It is possible that you have to add a note about this in the license file.

The new item should be documented, see https://github.com/minetest-mods/mesecons/tree/master/mesecons_pressureplates/doc and http://mesecons.net/items.html.

Why should this new pressure plate be crafted with sandstone blocks and not with sandstone?

The texture names should start with `"mesecons_pressureplates_"`.

frukto:
There are already some pressure plates. Wouldn't it be useful to add all the remaining pressure plates (e.g. dessert stone, ...) as well? I would rather see a complete package than an additional mod.

I'll check the licences. Add documentation, also I am not sure how to produce the preview image. Also sandstone makes more sense then sandstone block. No idea why i have chosen sandstone block :D. But despite all this, are willing to merge at all. Or do you really want to see additional pressure plates in a separate mod?

DS-Minetest:
>Add documentation, also I am not sure how to produce the preview image.

@sfan5 might know, they have made the fpga preview image, I think.

>are willing to merge at all.

I'm just a contributor, no member. ^_^

frukto:
I addressed all the issues. Well, i have no high res preview. 
The textures from minetest are all CC BY-SA 3.0 as are the textures from mesecons. So everything is fine here.

numberZero:
I’m neutral.

> The textures from minetest are all CC BY-SA 3.0 as are the textures from mesecons. So everything is fine here.

Strictly speaking, that’s *not* OK as the BY part means, texture author (copyright holder) must be mentioned as such (that’s what the `COPYING` file is for here). But in the real word, that hardly ever matters; enforcement capabilities is the only important thing in copyright (as well as in any other lawsuit, to a great extent).

leucome:
What about making a nice thin frame with an empty middle.  To make  already existing blocks looks like a pressure plate. If well done it can probably even look better than the current default plate and it would work with anything like literally.

frukto:
@leucome Can you be more specific. I think i don't get your idea.

ssieb:
My understanding of his suggestion is that the pressure plate is a very thin empty frame so you can see the node beneath it.  That way any node can (appear to) be a pressure plate.

leucome:
@frukto   Here you can see a demo.
https://www.youtube.com/watch?v=5ZoeMNcLDU4&feature=youtu.be

frukto:
@leucome That's cool. Did you already implement it? I think it a great addition to mesecons-mod. In particular because it lessens the need/discussion for additional plates.
If you have nothing ready, I'll think I create an additional PR, superseding this one.

leucome:
Yes it's ready. I  just went to upload and create a PR.

PR #480, opened 2019-09-14
Added static vertical mesecons (for more compact circuits)

FranchuFranchu:
This PR adds two blocks
- Static vertical wire bottom `mesecons_extrawires:static_vertical_bottom`
- Static vertical wire `mesecons_extrawires:static_vertical`

These two blocks (no crafting yet) allow you to make compact circuits by replacing this
bad one
with this
good one
They don't have a separate texture yet, but since I am bad at doing textures, I used the normal vertical mesecon texture

### DIFF
--- a/mesecons_extrawires/vertical.lua
+++ b/mesecons_extrawires/vertical.lua
@@ -12,9 +12,10 @@ local bottom_box = {
 	type = "fixed",
 	fixed = {
 		{-8/16, -8/16, -8/16, 8/16, -7/16, 8/16},
-		{-1/16, -7/16, -1/16, 1/16, 8/16, 1/16},
+		{-1/16, -7/16, -1/16, 1/16, 8/16, 1/16}
 	}
 }
+
 
 local vertical_rules = {
 	{x=0, y=1, z=0},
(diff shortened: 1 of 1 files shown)


VanessaE:
Can't you do this with mese blocks?

benrob0329:
You can do similar things with mese blocks, but not this.

DS-Minetest:
I'd like something like this. It would be very useful.

Mese blocks are very expensive.

Two (or three) things are actually needed:
- Plates that can be placed in the middle.
- Rods on top (to conduct to nodes from below).
- Maybe also rods on the bottom to conduct from below if that's not already possible with normal vertical wire. (If the current vertical wire doesn't connect to the bottom, it's weird. It should then have insulation on the bottom per texture.)

benrob0329:
What if we separate the rods from the plates, rather than adding 3-4 new nodes?

DS-Minetest:
Automatic placement is convenient for the user but PITA for the modder.

Having 4 nodes to place manually would be easy and powerful:
- A rod, connecting to top and bottom.
- A plate, connecting to north, south, east, west (=nesw) and bottom.
- A plate with insulated bottom combined with a rod, connecting to nesw, and top.
- A plate combined with a rod, connecting to all 6 sides.

DS-Minetest:
If we add a 5th item, converted from the current vertical mesecon wire item per alias, as craftitem, we could use this to automatically place.

benrob0329:
I mostly agree, but I'd simplify it a bit.

* A Rod, connecting top and bottom
* A Plate, which connects to rods on the top and bottom, and to anything else on nsew

DS-Minetest:
For backwards compatibility the plate that doesn't connect to the top is needed.

benrob0329:
That is true, although that is mostly a corner case that I wonder if we could cover with something less..janky. I have used it, but not often. All perfect-world clean mechanic wishes aside though, the bare minimum currently sits at:

* Rods: Top and Bottom Connecting
* "Center" Plates: Top and Bottom for Rods, NSEW For Everything Else
* Bottom Plates: Top, NSEW Connecting

The more I think about it though, the more I feel like we might be better just to add the center plate and keep everything else as-is. With the current state of mesecons I don't think we can get a cleaner setup _and_ maintain backwards compatibility (although I wouldn't mind a Mesecons 2.0 which cleans up and normalizes the mechanics).

DS-Minetest:
I've opened a second PR: #481
(Sorry, FranchuFranchu.)

Getsuga88:
I usually use ghoststone blocks as a workaround for that as they behave kinda like mese blocks (transmit mesecon signal, also way less expensive).

PR #481, opened 2019-09-17
Add static middle vertical wire

DS-Minetest:
This is highly related to and closes #480. (Sorry that I'm opening a second PR, FranchuFranchu.)

What this PR adds / changes / does and other notes:
- Everything is tested and works fine.
- The vertical bottom is visually insulated at the bottom.
- One new conductor: the static middle vertical wire
- It can be crafted.
- It can be rotated (wallmounted).
- Vertical mesecons connects to it if it's rotated up or down.
- It can receive signals from buttons and co.


DS-Minetest:
I've done some small improvements and fixes.

DS-Minetest:
Would someone like to review or test please btw.?

leucome:
I use it since the beginning... It work great and it's really useful.  I'm not so sure why nobody merged it already. Do it need more review or somebody with some administration privilege or something else?

Digiline also provide a block with the same feature would be nice if Mesecon had it by default too.

DS-Minetest:
>I use it since the beginning... It work great and it's really useful.

That's nice to hear!

>Do it need more review or somebody with some administration privilege or something else?

Somebody would have to review and someone would then have to merge.

leucome:
> What are you referring to? It's probably a digiline expansion, isn't it?
You are right it come with an other mod called digistuff.

numberZero:
Rotation is broken: the nodes aren’t updated when they should, and even when they are, some combinations are still problematic.

Conductive Air | Power Mesecon
---|---
screenshot_20200628_011651 | screenshot_20200628_012143

Desour:
Thanks for testing!

The left thing is actually intended. Otherwise it would be weird if there was a mese block and it didn't connect. I know, this is ugly, but player might be used to such connections anyway (eg. from the pipeworks connecting tubes).

I've tried to fix the right thing, but I couldn't get it working correctly. Therefor I've just disabled rotation of active wire.

(I've also rebased.)

PR #482, opened 2019-09-21
Use table.copy in mesecons.tablecopy

### LINKED ISSUE #473: Luacontroller recursive table interrupt iid crash
Server crash POC:
```lua
hax = {}
hax.a = hax
interrupt(1, hax)
```
After browsing the source I've found that the iid is passed to add_action without serialization ([init.lua](https://github.com/minetest-mods/mesecons/blob/master/mesecons_luacontroller/init.lua#L298))
```lua
local msg_ser = minetest.serialize(iid)
if #msg_ser <= mesecon.setting("luacontroller_interruptid_maxlen", 256) then
	mesecon.queue:add_action(pos, "lc_interrupt", {luac_id, iid}, time, iid, 1)
else
```

numberZero:
`mesecons.tablecopy` didn’t support recursive tables, while Minetest `table.copy` works well for them.

~~Fixes~~ #473

### DIFF
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -186,19 +186,11 @@ function mesecon.invertRule(r)
 	return vector.multiply(r, -1)
 end
 
-function mesecon.tablecopy(table) -- deep table copy
-	if type(table) ~= "table" then return table end -- no need to copy
-	local newtable = {}
-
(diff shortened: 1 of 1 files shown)


DS-Minetest:
Code looks good. :+1:

PR #484, opened 2019-09-22
set mvps_protection_mode default to the documented value

fluxionary:
The default value of mvps_protection_mode differs in the settingtypes file and in the code. According to the PR for that feature, the default should be "compat", so I've changed the code so that it is "compat".


DS-Minetest:
Related: #466
@numberZero

numberZero:
Thanks. I forgot there are 2 places where the default value is stated: one is `settingtypes.txt`, while another one is the `mesecon.setting` call.

PR #485, opened 2019-09-27
Do not print the mesecons OK message

HybridDog:
I don't think it's important to print that the mod loading succeeded. Other mods also do not print a message when loading.

### DIFF
--- a/mesecons/init.lua
+++ b/mesecons/init.lua
@@ -118,7 +118,7 @@ function mesecon.receptor_off(pos, rules)
 end
 
 
-print("[OK] Mesecons")
+minetest.log("info", "[OK] Mesecons")
 
 -- Deprecated stuff
 -- To be removed in future releases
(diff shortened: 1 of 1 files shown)


sfan5:
I don't know where I saw this first, but some mods put this inside a `if minetest.settings:get_bool("log_mods") then`.

louisroyer:
Also, maybe the message could be changed to something more descriptive like `[Mesecon] Successfully loaded` rather than `[OK] Mesecon`…

HybridDog:
In newest Minetest this message is actually redundant.
minetest --info shows the loaded mod and how long the loading took, i.e. something like this:
```
2020-02-24 20:41:51: INFO[Main]:   [wield3d     ] ["[…]/wield3d/init.lua"]
2020-02-24 20:41:51: INFO[Main]: Mod "wield3d" loaded after 0 seconds
```

HybridDog:
The message is now completely removed.

PR #487, opened 2019-10-10
Servo motor sound for movestone

leucome:
It's a slightly muffed  and relatively low pitch sound, similar to the sound of a large cnc machine. 
There is the link for the license. It's is also included in the files meta. 
[CC0 creativecommons for movestone.ogg ](https://creativecommons.org/choose/zero/results?license-class=zero&name=Leucome&actor_href=https%3A%2F%2Fgithub.com%2Fleucome%2Fmesecons%2Fblob%2Fmaster%2Fmesecons_movestones%2Fsounds%2Fmovestone.ogg&work_title=movestone.ogg&work_jurisdiction=CA&confirm=confirm&understand=confirm&lang=fr_CA&field1=continuez&waiver-affirm=affirm)


numberZero:
cool... I didn’t like the sound first time, but caught myself building a noise machine just a moment after =)

screenshot_20200628_003116

I like it. I really like when I can *hear* the machine operating...

mesecons_movestones/init.lua
```diff
@@ -66,6 +66,7 @@ function mesecon.register_movestone(name, def, is_sticky, is_vertical)
 		mesecon.on_dignode(pos, node)
 		mesecon.on_placenode(frontpos, node)
 		minetest.get_node_timer(frontpos):start(timer_interval)
+		minetest.sound_play("movestone", {pos = pos, max_hear_distance = 20, gain = 0.5,})
```
Desour, line 69: [`sound_play`](https://github.com/minetest/minetest/blob/7be082f9a8bb6bd46c226d7ef4c42f0fd9fe7314/doc/lua_api.txt#L5177) now has an `ephemeral` parameter. I think, it should be set to true here.

numberZero: It should be set to true everywhere in Mesecons, to not leak handles at least. That’s a change for a whole PR, thanks for spotting!

numberZero: #521

PR #489, opened 2019-10-12
New sticky and slippy blocks

leucome:
Linear sticky block,   sticky on  opposite side of one axis.  
Flat sticky block,  sticky on opposite side of two axis.
These block can switch to any x,y,z orientation we can chose the next axis by punching it with anything sticky.  

The slippy block  do not stick to stickypiston stickyblock and movestone. 

A little demo video, https://www.youtube.com/watch?v=DumoGyWlGJg&feature=youtu.be


DS-Minetest:
Rotation should probably better be done with a [screwdriver](https://github.com/minetest/minetest_game/blob/6042e963f78a4559c3eded56a764f1a20b5219f9/game_api.txt#L552-L568).

In some places the indentation is a bit wrong.

Could you explain how the slippy block works? I guess it uses `buildable_to` in some hacky way. There might be better ways to do this.

mesecons_stickyblocks_flat/init.lua
```diff
@@ -0,0 +1,157 @@
...
+	{x =  0, y =  1, z =  0},
+	{x =  0, y = -1, z =  0},
+}
+
+mesecon.rules.yz = {
+	{x =  0, y =  1, z =  0},
+	{x =  0, y = -1, z =  0},
+	{x =  0, y =  0, z =  1},
+	{x =  0, y =  0, z = -1},
+}
```
DS-Minetest: Make those tables in this file: https://github.com/minetest-mods/mesecons/blob/master/mesecons/presets.lua

mesecons_slippyblocks/init.lua
```diff
@@ -0,0 +1,76 @@
...
+	
+	minetest.register_node("mesecons_slippyblocks:slippy_block", {
+	description = "Slippy block)",
+	drawtype = "nodebox",
+	tiles = {"mesecons_slippyblocks.png"},
+	walkable     = true,
+	pointable    = true,
+	diggable     = true,
+	buildable_to = true,
+	groups = {cracky=3,oddly_breakable_by_hand=3},
```
DS-Minetest: This would look nicer:
```suggestion
	groups = {cracky = 3, oddly_breakable_by_hand = 3},
```
(Not only here.)

mesecons_slippyblocks/init.lua
```diff
@@ -0,0 +1,76 @@
+-- Stippy blocks can be used together with pistons or movestones
+	
+	minetest.register_node("mesecons_slippyblocks:slippy_block", {
+	description = "Slippy block)",
+	drawtype = "nodebox",
+	tiles = {"mesecons_slippyblocks.png"},
+	walkable     = true,
+	pointable    = true,
+	diggable     = true,
```
DS-Minetest: I suggest not to explicitly set all these properties to their default values.

mesecons_stickyblocks_linear/init.lua
```diff
@@ -0,0 +1,144 @@
...
+			minetest.swap_node(pos, {name = "mesecons_stickyblocks_linear:sticky_block_x" })
+		end
+	end,
+})
+	
+mesecon.rules.y = {
+	{x =  0, y =  1, z =  0},
+	{x =  0, y = -1, z =  0},
+	{x =  0, y =  0, z =  0},
+	{x =  0, y =  0, z =  0},
```
DS-Minetest: Why the (0,0,0)? Should it connect to itself?

mesecons_stickyblocks_flat/init.lua
leucome: If they want to merge the change it's indeed a good idea to move them back in preset. I placed them there originally. But I moved them so that it can work as stand alone with older/unmodified version of mesecon.

leucome:
> Rotation should probably better be done with a [screwdriver](https://github.com/minetest/minetest_game/blob/6042e963f78a4559c3eded56a764f1a20b5219f9/game_api.txt#L552-L568).
> 
> In some places the indentation is a bit wrong.
> 
> Could you explain how the slippy block works? I guess it uses `buildable_to` in some hacky way. There might be better ways to do this.

I noticed that the sticky system already ignore air and water or any build_to block.  So I made a built_to block that swap place with a common block when we try  interact with it. 

I am sure there is a better way to do this.  But it will require change to be done to other mesecon files. If these block are approved as a concept and included to mesecon it will be a good idea to look into these improvement.

PR #491, opened 2019-10-23
Add a pressure sensor

leucome:
It's like the pressure plate but with an empty middle. So it can be combined with other block to achieve different looks.


leucome:
A video demonstration. 
https://www.youtube.com/watch?v=5ZoeMNcLDU4&feature=youtu.be

frukto:
This supersedes #479 and provides a better solution. Great work @leucome

thomasrudin:
oh, this looks awesome... :+1:

thomasrudin:
@leucome have you considered doing this as a separate mod? Getting PR's merged in this repo is neither fast nor very rewarding.. :disappointed:

DS-Minetest:
Sorry for not having looked into this earlier. :slightly_smiling_face: 

It looks like you have copied everything from `mesecons_pressureplates` and modified some things.
I suggest you add the new pressure sensor plate in `mesecons_pressureplates` instead and use the `mesecon.register_pressure_plate` function from there (modify the function to allow your needs).
As far as I can see, everything what makes this different from normal pressure plates is basically just the nodebox.
(In the current changes, you just copy the descriptions for mesecons.net and also override `mesecon.register_pressure_plate`, which is obviously bad.)
The specified nodebox has a hole in the middle, right? In this special case you can make the textures transparent in the middle and just use two (or three, if the frame has two steps) boxes. The 2nd (/3rd) box is then in the middle and inverted, so that there are faces in the inside. (I wonder if anybody got me right here. :sweat_smile: )

leucome:
The nodebox hole in the middle can be useful to replace the block under because we can dig right trough it. If I remember correctly using a texture hole prevent us to dig in the middle. 

It was originally built as part of mesecons_pressureplates..  I moved it out so that it can be used alone or disabled/enabled ect...  
If it's moved back with the original pressure plate it will be required to rebuild all the plate because they stop to works when the mod name change. 

Anyway first I'll fix the description and the mesecon.register_pressure_plate function.

numberZero:
Nice. But there are problems.

* Preview is obviously wrong:
preview
* Nodebox doesn’t correspond to the texture:
screenshot_20200628_004329-edited
If you need that complex shape, use a mesh instead.
* Inventory image is highly unclear. Wield image is okay, MT won’t let you do better.
* Stone becoming yellow is strange:
screenshot_20200628_010723

Didn’t read the code yet.

numberZero:
Stale: “improvement needed” for 3 years.

PR #494, opened 2019-11-02
FPGA: Various code changes

SmallJoker:
1) Unified operands table to remove duplicated code
2) Detach the formspec to use non-node-bound formspecs
    * The formspec is quite big: 56 dropdown elements
    * Gives the FPGA program serialization more importance - almost no other metadata left over

### Changes behaviour
1) The formspec only closes when it was programmed successfully
    *  Is an exit button wanted?
2) Protection checks

### DIFF
--- a/mesecons_fpga/init.lua
+++ b/mesecons_fpga/init.lua
@@ -3,7 +3,6 @@ plg.rules = {}
 -- per-player formspec positions
 plg.open_formspecs = {}
 
-
 local lcore = dofile(minetest.get_modpath(minetest.get_current_modname()) .. "/logic.lua")
 dofile(minetest.get_modpath(minetest.get_current_modname()) .. "/tool.lua")(plg)
 
--- a/mesecons_fpga/init.lua
+++ b/mesecons_fpga/init.lua
@@ -192,7 +192,7 @@ plg.to_formspec_string = function(is, err)
 	local function dropdown_action(x, y, name, val)
 		local selected = 0
 		local titles = { " " }
-		for i, data in ipairs(lcore.get_operands()) do
+		for i, data in ipairs(lcore.get_operations()) do
 			titles[i + 1] = data.fs_name
 			if val == data.gate then
 				selected = i + 1
@@ -249,7 +249,7 @@ plg.from_formspec_fields = function(fields)
 		end
 	end
 	local function read_action(s)
-		for i, data in ipairs(lcore.get_operands()) do
+		for i, data in ipairs(lcore.get_operations()) do
 			if data.fs_name == s then
 				return data.gate
 			end
--- a/mesecons_fpga/init.lua
+++ b/mesecons_fpga/init.lua
@@ -1,9 +1,9 @@
 local plg = {}
 plg.rules = {}
 
+
 local lcore = dofile(minetest.get_modpath(minetest.get_current_modname()) .. "/logic.lua")
 dofile(minetest.get_modpath(minetest.get_current_modname()) .. "/tool.lua")(plg)
-
 
 plg.register_nodes = function(template)
 	-- each loop is for one of the 4 IO ports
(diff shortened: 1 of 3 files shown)


SmallJoker:
Replaces #493

mesecons_fpga/init.lua
```diff
@@ -1,10 +1,23 @@
...
+
+plg.operands = {
+	-- index: Index in formspec
+	{ gate = "and",  short = "&", fs_name = " AND", func = function(a, b) return a and b end },
+	{ gate = "or",   short = "|", fs_name = "  OR", func = function(a, b) return a or b end },
+	{ gate = "not",  short = "~", fs_name = " NOT", func = function(a, b) return not b end },
+	{ gate = "xor",  short = "^", fs_name = " XOR", func = function(a, b) return a ~= b end },
+	{ gate = "nand", short = "?", fs_name = "NAND", func = function(a, b) return not (a and b) end },
+	{ gate = "buf",  short = "_", fs_name = "   =", func = function(a, b) return b end },
+	{ gate = "xnor", short = "=", fs_name = "XNOR", func = function(a, b) return a == b end },
```
sfan5: The duplication was intentional.
`logic.lua` was supposed to handle everything of the logic gate itself and provide and API, while the rest would create the logic gate node inside MT without touching the internals more than necessary.

mesecons_fpga/logic.lua
```diff
@@ -11,20 +13,14 @@ lg.serialize = function(t)
 			return tostring(t.n)
 		end
 	end
-	local function _action(s)
-		if s == nil then
-			return " "
+	-- Compress actions (gates) from eg. "and" to "&"
```
sfan5: While it does indeed "compress" every operation into just 1 character, this would more correctly be called serialization.
(same with decompress -> deserialize)

mesecons_fpga/init.lua
SmallJoker: Alright, I thought unifying these definitions would reduce the risk of out-of-sync definitions in future changes.
Would it be okay if I move this definition to `logic.lua` and expose it using API functions? Otherwise the first commit is totally unnecessary effort.

sfan5: Sure, in the end there's very little point in me opposing refactoring just because it doesn't match my original design decisions.

mesecons_fpga/logic.lua
```diff
@@ -1,5 +1,22 @@
...
+	-- index: Index in formspec
+	{ gate = "and",  short = "&", fs_name = " AND", func = function(a, b) return a and b end },
+	{ gate = "or",   short = "|", fs_name = "  OR", func = function(a, b) return a or b end },
+	{ gate = "not",  short = "~", fs_name = " NOT", func = function(a, b) return not b end },
+	{ gate = "xor",  short = "^", fs_name = " XOR", func = function(a, b) return a ~= b end },
+	{ gate = "nand", short = "?", fs_name = "NAND", func = function(a, b) return not (a and b) end },
+	{ gate = "buf",  short = "_", fs_name = "   =", func = function(a, b) return b end },
+	{ gate = "xnor", short = "=", fs_name = "XNOR", func = function(a, b) return a == b end },
+	{ gate = "nor",  short = "!", fs_name = " NOR", func = function(a, b) return not (a or b) end },
+}
```
sfan5: The "does this operation need two operands?" (used below) should also be moved into this table:
https://github.com/minetest-mods/mesecons/blob/4b7e2d065c350a30ccf064fd9806e333714a8d8a/mesecons_fpga/logic.lua#L122

mesecons_fpga/logic.lua
```diff
@@ -1,5 +1,22 @@
+
 local lg = {}
 
+local operands = {
```
sfan5: now that I think about it this should be called `operations` or `gates`

mesecons_fpga/logic.lua
```diff
@@ -1,5 +1,22 @@
...
+	-- index: Index in formspec
```
sfan5: some documentation would be nice:
<pre>
-- gate: Internal name
-- short: Serialized form, single character
-- fs_name: Display name, padded to 4 characters
-- func: Function that applies the operation
</pre>

SmallJoker:
Updated.

Issue #470, opened 2019-07-22
[Feature Request] Allow use of string.split in luacontroller sandbox environment


neoh4x0r:
Referencing minetest forums post: https://forum.minetest.net/viewtopic.php?p=341908#p341908

----------
> thetoolman:
> I'm working with the lua controller and digilines and I want to make the lua controller count something that is sent in the digilines string.
>
> To be a little more clear here, the digilines chest spits out a string like this "default:cobblestone 9" where the first part is the item, and the last part is the number.
>
> I don't have a github account so I wont be making a merge request, but if someone wants to take a look at this and possibly add it or explain why it's not there, that'd be nice.
>
> So the code I added in mesecons_luacontroller/init.lua looks like this
>
> ```
> local env = {
>   ...
>   string = {
>     ...
>     split = string.split,
>     ...
>   }
>   ....
> }
> ```
----------
Are there any technical issues with adding string.split ... it seems like it would be a useful addition.   I know it also possible to use a table with named members as the message parameter, but that might not always be practical in all cases (especially when the output format is generated by something that the user has no control over).

benrob0329:
Lua doesn't have a `string.split` function.

neoh4x0r:
@benrob0329 
What about this:
```
local env = {
  ...
  string = {
    ...
    split = safe_string_split,
    ...
  }
  ....
}


-- string.split splits a string by a delimieter
local function safe_string_split(s,c)
	local list = {}
	if s then
		if c and string.match(s, c) then
			-- split s by delimieter c and append to list
			for word in string.gmatch(s, '([^' .. c .. ']+)') do
			    table.insert(list, word)
			end
		else
			-- s does not contain delimieter c so add s as the first element
			table.insert(list, s)
		end
	end
	return list
end
```

DS-Minetest:
If something like `string.split` is added, it might make sense to think about other helpers, too.

neoh4x0r:
@DS-Minetest 
Yes, that would probably be a good idea.

However, that raises another question:

What should the scope of helpers be limited to ?

(some helpers might be by-nature unwanted for the purposes of the sandbox)

A few that I can think of are:
- string.match (string contains),
- string.gmatch (regx matching), 
- ability to call members of vector class builtin/common/vector.lua
  https://github.com/minetest/minetest/blob/master/builtin/common/vector.lua

I was just looking and mintest already defines a lua function called string.split
[...]

PR #500, opened 2019-12-15
Add string.split function

Lejo1:
Adds the default minetest string.split function.
I know it's just one helper function but it were also asked in #470 and should be pretty safe.
A player could anyway add it to his controller:
``` Lua
--  Copied from mt code: https://github.com/minetest/minetest/blob/0df646e0689b5324c7bdb13205c2a63e42ebe8e8/builtin/common/misc_helpers.lua#L169
function string.split(str, delim, include_empty, max_splits, sep_is_pattern)
	delim = delim or ","
	max_splits = max_splits or -2
	local items = {}
	local pos, len = 1, #str
	local plain = not sep_is_pattern
	max_splits = max_splits + 1
	repeat
		local np, npe = string.find(str, delim, pos, plain)
		np, npe = (np or (len+1)), (npe or (len+1))
		if (not np) or (max_splits == 1) then
			np = len + 1
			npe = np
		end
		local s = string.sub(str, pos, np - 1)
		if include_empty or (s ~= "") then
			max_splits = max_splits - 1
			items[#items + 1] = s
		end
		pos = npe + 1
	until (max_splits == 0) or (pos > (len + 1))
	return items
end
```
We could also set the max_splits to a specific value to be 100 % sure.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -231,6 +231,16 @@ local function safe_string_find(...)
 	return string.find(...)
 end
 
+-- do not allow pattern matching in string.split (see string.find for details)
+local function safe_string_split(...)
+	if select(5, ...) then
+		debug.sethook() -- Clear hook
+		error("string.split: 'sep_is_pattern' (fifth parameter) may not be used in a Luacontroller")
+	end
+
+	return string.split(...)
+end
+
 local function remove_functions(x)
 	local tp = type(x)
 	if tp == "function" then
@@ -463,7 +473,7 @@ local function create_environment(pos, mem, event, itbl, send_warning)
 			reverse = string.reverse,
 			sub = string.sub,
 			find = safe_string_find,
-			split = string.split,
+			split = safe_string_split,
 		},
 		math = {
 			abs = math.abs,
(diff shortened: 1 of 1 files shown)


Lejo1:
Just one side note:
We have a secure string.find function, but string.split uses the normal one. Is that insecure?

I’m pretty unsure about this cause I don’t know why we need a secure find function

SmallJoker:
Right. `string.split` requires a custom safe implementation according to
https://github.com/minetest-mods/mesecons/blob/f1ff09277011dd43d253184e3c7c660625a509ca/mesecons_luacontroller/init.lua#L223-L228

DS-Minetest:
Just do the same as for `string.find`:
```lua
-- do not allow pattern matching in string.split (see string.find for details)
local function safe_string_split(...)
	if select(5, ...) then
		debug.sethook() -- Clear hook
		error("string.split: 'sep_is_pattern' (fifth parameter) may not be used in a Luacontroller")
	end

	return string.split(...)
end
```
I wonder what other [helper functions](
https://github.com/minetest/minetest/blob/44458896d9097c15c722c5bffb3d72e14962b7ce/doc/lua_api.txt#L2839-L2907) it might be worth adding to the luacontroller.

Lejo1:
> Just do the same as for `string.find`

Thanks, did that!

Lejo1:
Included in #507

PR #501, opened 2019-12-28
Refactor actionqueue.lua

DS-Minetest:
The actionqueue does some pretty bad stuff at the moment.
Eg. it takes finds the action with the highest priority (O(n)) for each action in `action_now` (=\> O(n²)) in every globalstep. This is replaced by a table.sort (with a comparator that should make the sort stable).

The goal of this PR is to increase the performance of the actionqueue, to modernize the code and to increase the readability while at the same time not changing the behaviour of anything.

It's very possible that there are other improvements left to do, but such changes might threaten the compatibility.

For testing, my [meseconometer](https://github.com/DS-Minetest/meseconometer) mod might be useful.

Also note that split view might make reviewing easier as some of the diffs are just renaming, movement of code or indentation changes.

### DIFF
--- a/mesecons/actionqueue.lua
+++ b/mesecons/actionqueue.lua
@@ -1,3 +1,28 @@
+--[[
+Mesecons uses something it calls an ActionQueue.
+
+The ActionQueue holds functions and actions.
+Functions are added on load time with a specified name.
+Actions are preserved over server restarts.
+
+Each action consists of a position, the name of an added function to be called,
+the params that should be used in this function call (additionally to the pos),
+the time after which it should be executed, an optional overwritecheck and a
+priority.
+
+If time = 0, the action will be executed in the next globalstep, otherwise the
+earliest globalstep when it will be executed is the after next globalstep.
+
+It is guaranteed, that for two actions ac1, ac2 where ac1 ~= ac2,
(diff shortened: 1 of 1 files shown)


DS-Minetest:
This is no longer WIP, please review. :slightly_smiling_face:

leucome:
I tried many Mesecon build in my laboratory world, they all behave as expected. 
So the compatibility is pretty good. 

Have you already did some benchmarks? If so what kind performance gain are to be expected?

DS-Minetest:
>Have you already did some benchmarks? If so what kind performance gain are to be expected?

No I haven't yet. I'm neither experienced in benchmarking nor in unittesting.

thomasrudin:
> Theoretically, function and action adding have no speed gain, but the globalstep should be faster.

I tried various things in that direction a few months ago, the performance gain was barely noticeable though... :disappointed:

Still: better than nothing :+1:

DS-Minetest:
Oh, btw. I've made some ugly benchmarks (So ugly that publishing hurts.): See https://github.com/DS-Minetest/mesecons_actionquque_test (liceses go partly back to mesecons as the actionqueue file is included, everything by me is CC0). There was a measureable effect.
Note that this PR does not only improve the performance but also improves code style and adds documentation.

PR #503, opened 2020-02-23
Add MT 5.0 translation support (2nd try)

Wuzzy2:
This PR makes Mesecons translatable (replaces #476).

Special changes:

* Interface of Luacontroller was changed, the baked-in text was stripped from the image and replaced by real text so it can be translated
* Remove all “you hacker you” strings. These are not helpful and just annoying if you somehow get one of these items due to a bug

Differences from #476:
* Everything is squashed
* No German translation

I did some rudimentary testing, but I advice you do some tests yourselves.


Wuzzy2:
Mesecons now has translation support. This PR has become obsolete.

PR #504, opened 2020-03-13
Solarpanel: cleanup

MoNTE48:
Sometimes I do small PRs with bugfixes and code style improvements.
Can I clean and fix `mesecons` code style a bit?

Random mod from the list:
Solarpanel. Significant reduction in code size.


mesecons_solarpanel/init.lua
```diff
@@ -1,101 +1,63 @@
...
-minetest.register_node("mesecons_solarpanel:solar_panel_on", {
+mesecon.register_node("mesecons_solarpanel:solar_panel", {
+	description = "Solar Panel",
 	drawtype = "nodebox",
-	tiles = { "jeija_solar_panel.png", },
-	inventory_image = "jeija_solar_panel.png",
-	wield_image = "jeija_solar_panel.png",
+	tiles = {"mesecons_solarpanel.png"},
+	inventory_image = "mesecons_solarpanel.png",
+	wield_image = "mesecons_solarpanel.png",
```
Desour: Renaming the textures might break texture packs and co. and is inconsistent with the rest of mesecons. However, this naming convention is better and there are probably not that many texture packs for mesecons and the few ones can be broken, pff.

Desour:
:+1: The code changes look good.
I wonder if combining the two abms is faster or slower.

MoNTE48:
@Desour 
1) it definitely uses less memory
2) checking node.name is not expensive

mesecons_solarpanel/init.lua
MoNTE48: This uses the Minetest modding recommendations. 

> Mods should generally prefix their textures with `modname_`, e.g. given
> the mod name `foomod`, a texture could be called:
> 
>     foomod_foothing.png

The use of these names in the future will facilitate the creation and support of texture packs. Although it may break some now. However, is there so many texture packs for Mesecons?

SmallJoker:
Please don't remove the `doc/` files. Those are used on the `mesecons.net` website.

MoNTE48:
@SmallJoker  I did not do it. Why do you think so?

SmallJoker:
@MoNTE48 Oh, they were optimized. I saw the removed bytes part and assumed they were deleted.

mesecons_solarpanel/init.lua
```diff
@@ -1,101 +1,63 @@
...
+	}}
 })
 
 minetest.register_craft({
-	output = "mesecons_solarpanel:solar_panel_off 1",
+	output = "mesecons_solarpanel:solar_panel_off",
 	recipe = {
 		{"mesecons_materials:silicon", "mesecons_materials:silicon"},
-		{"mesecons_materials:silicon", "mesecons_materials:silicon"},
+		{"mesecons_materials:silicon", "mesecons_materials:silicon"}
```
sfan5: There is no advantage to removing these commas.

MoNTE48: I know :) There is no clear information in the code style guide how to correctly.

MoNTE48:
After 1.5 months, is anyone interested in this?

PR #506, opened 2020-03-27
Make more nodes trigger special noteblock sounds

Wuzzy2:
This PR enables more nodes to trigger the special noteblock sounds:

* Hi-hat: Also accepts obsidian glass instead of only glass
* Snare: Also accepts locked chest instead of only chest
* Kick: Now accepts `group:stone` instead of only `default:stone`
* Crash: Now accepts `group:tree` instead of only `default:tree`
* Lite crash: Now accepts `group:wood` instead of only `default:wood`


Wuzzy2:
FYI: I have a couple of other noteblock-related PRs in the pipeline (the branches are already online), but I wait with submitting them for now, as they (kinda sorta) depend on this PR.

* Add pitch support to most noteblock (`noteplusplus`)
* More noteblock sounds (`noteplusplus`)
* Fade out noteblock's fire sound (because it sounds bad when the fire sounds ends so apruptly ...) (`noteblock_fade_out_fire`)
* Muffle noteblock sound when a block is placed above it, with 5 volume levels (`noteblock_muffle`)

PR #507, opened 2020-03-30
Add all safe helper function from lua_api.txt

Lejo1:
Includes #500 
Just need to be checked if there are any needing an safe implementation(Didn't found any).

PS: Can I do for math and table also just like for vector math = math? Do I need to list them all separately?

Discussion: Shall I implement get_us_time? There are extra blocks in mesecons + diglines to get the time so they would get useless...

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -231,6 +231,16 @@ local function safe_string_find(...)
 	return string.find(...)
 end
 
+-- do not allow pattern matching in string.split (see string.find for details)
+local function safe_string_split(...)
+	if select(5, ...) then
+		debug.sethook() -- Clear hook
+		error("string.split: 'sep_is_pattern' (fifth parameter) may not be used in a Luacontroller")
+	end
+
+	return string.split(...)
+end
+
 local function remove_functions(x)
 	local tp = type(x)
 	if tp == "function" then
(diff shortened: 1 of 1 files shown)


mesecons_luacontroller/init.lua
```diff
@@ -490,16 +503,22 @@ local function create_environment(pos, mem, event, itbl, send_warning)
 			rad = math.rad,
 			random = math.random,
 			sin = math.sin,
+			sing = math.sign,
```
Desour: The luacontroller can't sing.

mesecons_luacontroller/init.lua
```diff
@@ -508,6 +527,24 @@ local function create_environment(pos, mem, event, itbl, send_warning)
...
+			is_yes = minetest.is_yes,
+			is_nan = minetest.is_nan,
+			get_us_time = minetest.get_us_time,
+			parse_json = minetest.parse_json,
+			write_json = minetest.write_json,
+			rgba = minetest.rgba,
+			encode_base64 = minetest.encode_base64,
+			decode_base64 = minetest.decode_base64,
+		},
+		vector = vector,
```
Desour: >Can I do for math and table also just like for vector math = math? Do I need to list them all separately?

If you do this, the player can modify the `vector` table. This here is not a deep copy, you just copy the reference.

Lejo1:
Would vector = table.copy(vector) work?

mesecons_luacontroller/init.lua
Lejo1: Did using table.copy

Lejo1:
Is it possible to allow nonplain string.format for the dump and wrap_text function?

PR #509, opened 2020-03-30
Add droppers

Wuzzy2:
This adds a dropper.

The dropper is an effector that has an inventory with 9 slots. When powered, it will drop a random item in its inventory, provided the node in front of it is not solid (=`walkable`).

The dropper can face upwards, downwards or sideways.

I think it works fine (it's based on the droppers in MCL2), but you might dislike the textures (maybe someone more skilled than me can draw better ones).


mesecons_dropper/depends.txt
```diff
@@ -0,0 +1,2 @@
+mesecons
+screwdriver?
```
Panquesito7, line 2: Please add these to `mod.conf`.

numberZero:
Works, but:
1. Is such complicated texturing necessary?
2. Is this any different from pipeworks dispenser?

Wuzzy2:
1. I don't know what you mean. If 6 textures are too complicated to you, then I can't help you. :(
2. I don't know. I don't use pipeworks.

numberZero:
> If 6 textures are too complicated to you, then I can't help you. :(

I mean this:

> 3 node definitions are needed because of the way the textures are defined.

Also that prevents rotating droppers vertically. That may be a problem, it’s not always easy to place a node in the correct orientation.

numberZero:
> I don't use pipeworks.

So it may be the time to try. The dispenser does the same thing albeit a bit differently:

* Its inventory is 1×1 (so no random drops).
* It also acts on punch.
* It uses the `on_drop` callback (if any).

The difference is minor IMO, so is it reasonable to include that similar machine in Mesecons (provided that currently Mesecons never operate on items)?

Wuzzy2:
Oh well. This block I made is basically just a clone from the dropper in Minecraft. :P

johalun:
+1 for this! It's nice to have as randomizer.

PR #510, opened 2020-03-30
Add sound effects to FPGA programmer

Wuzzy2:
Just some beeps and bleeps for the FPGA programmer, played when:

* FPGA programmer copies something
* FPGA programmer writes something
* FPGA programmer has an error


SmallJoker:
Whereas not realistic, I like the audio feedback. IMO the sound volume could be lowered a bit, since it's only a feedback, not intended to broadcast music.

PR #512, opened 2020-04-24
Speed up search for objects to be moved in MVPS

cheapie:
This causes the object position checking to only have to check a subset of objects that are near the nodes being moved instead of iterating over every single object on the entire server.

Speedup seems to be most dramatic on servers with large numbers of entities - with about 3000 of them (this is rather common on servers using mods like signs_lib) a single piston push was taking up to 150ms before and is closer to 1-5ms at most now. The effects of this are quite noticeable when building machines using large numbers of pistons.

As far as I can tell, the behavior regarding when objects are moved remains exactly the same, including the bug that @benrob0329 loves so much. If anyone can find a situation where this PR *does* change behavior (aside from speed) then I'd love to hear about it, but I'm not currently aware of any.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -285,7 +285,7 @@ function mesecon.mvps_move_objects(pos, dir, nodestack, movefactor)
 	end
 	movefactor = movefactor or 1
 	dir = vector.multiply(dir, movefactor)
-	for id, obj in pairs(minetest.object_refs) do
+	for id, obj in pairs(minetest.get_objects_inside_radius(pos, #nodestack + 1)) do
 		local obj_pos = obj:get_pos()
 		local cbox = obj:get_properties().collisionbox
 		local min_pos = vector.add(obj_pos, vector.new(cbox[1], cbox[2], cbox[3]))


sfan5:
Related IRC discussion: http://irc.minetest.net/minetest/2020-04-24#i_5674187

Desour:
FYI, the thought behind this was that `get_objects_inside_radius` would iterate over every object anyways. The code was designed to work even with objects that have huge collision boxes. There might be objects, eg. vehicles, that have larger box radii than 2 nodes (actually less than 2 nodes, you have to subtract the distance from the mvps stack center (about `sqrt(1/2)`)).

sfan5:
> the thought behind this was that `get_objects_inside_radius` would iterate over every object anyways

That's still bad for two reasons, C++ code iterating over all objects and throwing most out is still much faster and if `get_objects_inside_radius` is ever optimized in the future you miss out on any speedup.

cheapie:
> FYI, the thought behind this was that `get_objects_inside_radius` would iterate over every object anyways. The code was designed to work even with objects that have huge collision boxes.
> There might be objects, eg. vehicles, that have larger box radii than 2 nodes (actually less than 2 nodes, you have to subtract the distance from the mvps stack center (about `sqrt(1/2)`)).

If it's desired for pistons to be able to move huge objects like that, I suppose it could be possible for the radius to be increased - should be trivial if you want to submit a PR for that.

PR #513, opened 2020-05-02
Add safe version of pcall and error function to luacontroller environment

rdococ:
Currently propagates all errors generated by the luacontroller code itself (e.g. string length overflow), which might not be the desired behavior, but can be fixed if necessary.

I'm not sure if anyone actually wants this feature, but it could be useful for certain forms of non-local control flow, and judging by the comment I saw by the safe_globals definition, it seems like there was some interest in it.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -225,10 +225,23 @@ end
 local function safe_string_find(...)
 	if (select(4, ...)) ~= true then
 		debug.sethook() -- Clear hook
-		error("string.find: 'plain' (fourth parameter) must always be true in a Luacontroller")
+		error("string.find: 'plain' (fourth parameter) must always be true in a Luacontroller", 2)
 	end
 
 	return string.find(...)
+end
+
+-- pcall can capture certain errors after the debug hook is removed.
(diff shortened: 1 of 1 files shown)


mesecons_luacontroller/init.lua
```diff
@@ -430,7 +443,7 @@ end
 local safe_globals = {
 	-- Don't add pcall/xpcall unless willing to deal with the consequences (unless very careful, incredibly likely to allow killing server indirectly)
```
TurkeyMcMac, line 444:
```suggestion
```

TurkeyMcMac:
I suggest you write `safe_pcall` something like this (I haven't tested the code):

```lua
local function finish_safe_pcall(...)
	local ok, err = ...

	if not ok and not debug.gethook() then
		error(err, 0)
	end

	return ...
end

local function safe_pcall(f, ...)
	return finish_safe_pcall(pcall(f, ...))
end
```

This handles `nil` return values correctly.

Calling `error` with 0 as its second argument throws the given message without modifying it.

Bituvo:
I support this! I would not like to manually check inputs for errors.

PR #515, opened 2020-05-19
Gates: Modify appearance

Awkor:
I elevated the center part of the gate to follow (more or less) the style of other blocks in the mod (like the microcontroller). With this change, the underside doesn't show the gate icon anymore. If this is unwanted, let me know and I'll revert this part.

I also added a small 'cable connection' that shows the output direction and the status of the gate.

I wanted to make the same for the inputs, but (if I understood correctly) one would have to register a node for each combination of inputs and outputs. That sounds pretty unwieldy. I'll maybe get around to it when [this](https://github.com/minetest/minetest/issues/9193) issue gets resolved.

# Screenshots

Annotation 2020-05-19 030919
Annotation 2020-05-19 031057
Annotation 2020-05-19 031347
Annotation 2020-05-19 031326
Annotation 2020-05-19 031750
Annotation 2020-05-19 031811


sfan5:
I liked the gates being flat, but this visual change is nice too.
Have you compressed the textures with a tool like [optipng](http://optipng.sourceforge.net/)?

> one would have to register a node for each combination of inputs and outputs

This is correct.

Awkor:
Textures are now compressed using `optipng -o7 -zm1-9 -strip all *.png`

PR #524, opened 2020-08-06
Insulated double corner

numberZero:
Fixes #408. Also makes the regular corner curved.
screenshot_20200816_123338

PR #528, opened 2020-08-31
add `.luacheckrc` / gh action and fix reported errors

BuckarooBanzay:
# Overview

This PR adds a `.luacheckrc` and the corresponding github action to check on every PR and Push.

# TODO

* [ ] ~~enable github actions on the current PR for testing~~
* [x] Fix remaining errors
* [x] Resolve more severe code-issues below

## silenced luacheck errors

The following error-types are silenced for this PR:
```lua
ignore = {
	"631", -- line too long
	"621", -- inconsistent indendation
	"542", -- empty if branch
}
```

# Unresolved code-issues

## microcontroller

`code` in this context is global and should be always nil:
https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons_microcontroller/init.lua#L472

## luacontroller

`newstate` is a global variable, `L[port]` should always be `0`
https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons_luacontroller/init.lua#L63

## extrawires

`corner_nodebox` is global and not defined within the mod:
https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons_extrawires/corner.lua#L33
@numberZero this was your addition in #524 is there a file or something missing?

## base "mesecons" mod

`offstate` and `onstate` aren't defined anywhere:
https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons/internal.lua#L319
https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons/internal.lua#L335

### DIFF
--- a/.github/workflows/luacheck.yml
+++ b/.github/workflows/luacheck.yml
@@ -1,0 +1,17 @@
+name: luacheck
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: apt
+      run: sudo apt-get install -y luarocks
+    - name: luacheck install
+      run: luarocks install --local luacheck
+    - name: luacheck run
+      run: $HOME/.luarocks/bin/luacheck ./
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,30 @@
+
+globals = {
+	"mesecon",
+	"minetest"
+}
+
+-- ignore unused vars
+unused = false
+
+read_globals = {
+	-- Stdlib
+	string = {fields = {"split"}},
+	table = {fields = {"copy", "getn"}},
+
+	-- Minetest
+	"vector", "ItemStack",
+	"dump", "VoxelArea",
+
+	-- system
+	"DIR_DELIM",
+
+	-- deps
+	"default", "screwdriver"
+}
+
+ignore = {
+	"631", -- line too long
+	"621", -- inconsistent indendation
+	"542", -- empty if branch
+}
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -16,9 +16,6 @@ read_globals = {
 	"vector", "ItemStack",
 	"dump", "VoxelArea",
 
-	-- system
-	"DIR_DELIM",
-
 	-- deps
 	"default", "screwdriver",
 	"digiline", "doors"
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -20,7 +20,8 @@ read_globals = {
 	"DIR_DELIM",
 
 	-- deps
-	"default", "screwdriver"
+	"default", "screwdriver",
+	"digiline", "doors"
 }
 
 ignore = {
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -574,15 +574,13 @@ end
 
 -- Returns success (boolean), errmsg (string)
 -- run (as opposed to run_inner) is responsible for setting up meta according to this output
-local function run_inner(pos, code, event)
-	local meta = minetest.get_meta(pos)
+local function run_inner(pos, meta, event)
 	-- Note: These return success, presumably to avoid changing LC ID.
 	if overheat(pos) then return true, "" end
 	if ignore_event(event, meta) then return true, "" end
 
 	-- Load code & mem from meta
 	local mem  = load_memory(meta)
-	code = meta:get_string("code")
 
 	-- 'Last warning' label.
 	local warning = ""
@@ -595,7 +593,7 @@ local function run_inner(pos, code, event)
 	local env = create_environment(pos, mem, event, itbl, send_warning)
 
 	-- Create the sandbox and execute code
-	local f, msg = create_sandbox(code, env)
+	local f, msg = create_sandbox(meta:get_string("code"), env)
 	if not f then return false, msg end
 	-- Start string true sandboxing
 	local onetruestring = getmetatable("")
@@ -651,7 +649,7 @@ end
 local function run(pos, event)
 	local meta = minetest.get_meta(pos)
 	local code = meta:get_string("code")
-	local ok, errmsg = run_inner(pos, code, event)
+	local ok, errmsg = run_inner(pos, meta, event)
 	if not ok then
 		reset_meta(pos, code, errmsg)
 	else
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -231,8 +231,8 @@ local function safe_string_find(...)
 	return string.find(...)
 end
 
-local function remove_functions(x)
-	local tp = type(x)
+local function remove_functions(obj)
+	local tp = type(obj)
 	if tp == "function" then
 		return nil
 	end
@@ -261,9 +261,9 @@ local function remove_functions(x)
 		end
 	end
 
-	rfuncs(x)
+	rfuncs(obj)
 
-	return x
+	return obj
 end
 
 -- The setting affects API so is not intended to be changeable at runtime
@@ -582,7 +582,7 @@ local function run_inner(pos, code, event)
 
 	-- Load code & mem from meta
 	local mem  = load_memory(meta)
-	local code = meta:get_string("code")
+	code = meta:get_string("code")
 
 	-- 'Last warning' label.
 	local warning = ""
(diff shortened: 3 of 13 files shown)


numberZero:
Sorry but PR content differs from the description significantly. It may be better not to mix tooling and code changes.

> newstate is a global variable, L[port] should always be 0

Might that be the reason for spurious “on” events?

> corner_nodebox is global and not defined within the mod:

It’s not needed anymore, the corner is a mesh node now

BuckarooBanzay:
> Sorry but PR content differs from the description significantly. It may be better not to mix tooling and code changes.

title changed, tooling and code changes should be introduced together

.luacheckrc
```diff
@@ -0,0 +1,31 @@
+
+globals = {
+	"mesecon",
+	"minetest"
```
Desour, line 4: Why does mesecons need to write `minetest`?

.luacheckrc
```diff
@@ -0,0 +1,31 @@
...
+	-- Stdlib
+	string = {fields = {"split"}},
+	table = {fields = {"copy", "getn"}},
+
+	-- Minetest
+	"vector", "ItemStack",
+	"dump", "VoxelArea",
+
+	-- system
+	"DIR_DELIM",
```
Desour, line 20: Afaik, one should rather use `"/"` than `DIR_DELIM`. `DIR_DELIM` is something from builtin, so minetest, not system.

.luacheckrc
```diff
@@ -0,0 +1,31 @@
...
+
+	-- deps
+	"default", "screwdriver",
+	"digiline", "doors"
+}
+
+ignore = {
+	"631", -- line too long
+	"621", -- inconsistent indendation
+	"542", -- empty if branch
```
Desour, line 30: Why do we want to keep empty if branches?

Btw. builtin also has a `.luacheckrc`, which might be interesting for this PR:
https://github.com/minetest/minetest/blob/master/.luacheckrc

mesecons_luacontroller/init.lua
```diff
@@ -582,7 +582,7 @@ local function run_inner(pos, code, event)
 
 	-- Load code & mem from meta
 	local mem  = load_memory(meta)
-	local code = meta:get_string("code")
+	code = meta:get_string("code")
```
Desour, line 585: This line could simply be removed.
`run_inner` is only called at one place (in function `run`).
I suggest to replace the `code` parameter with `meta` (which is already gotten and used in `run`).

mesecons_pistons/init.lua
```diff
@@ -462,8 +462,8 @@ end
 
 local function piston_get_stopper(node, dir, stack, stackid)
 	local pistonspec = get_pistonspec(node.name, "onname")
-	local dir = vector.multiply(minetest.facedir_to_dir(node.param2), -1)
-	local pusherpos = vector.add(stack[stackid].pos, dir)
+	local inverted_dir = vector.multiply(minetest.facedir_to_dir(node.param2), -1)
+	local pusherpos = vector.add(stack[stackid].pos, inverted_dir)
```
Desour, line 466: `inverted_dir` suggests that it is `-dir`, which it isn't. (`dir` is the direction of the mvps movement and `inverted_dir` is the direction in which the pusher of the moved piston is.)
I suggest replacing the `dir` parameter with `_` and keeping the name of the local `dir`.

Desour (commented):
:+1: For using luacheck. (It's incredible how many mistakes this already captured.)

I don't know how the github workflows thing works.

Desour:
> ## microcontroller
> 
> `code` in this context is global and should be always nil:
> 
> https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons_microcontroller/init.lua#L472

I guess it should be `params.code` there.

> ## luacontroller
> 
> `newstate` is a global variable, `L[port]` should always be `0`
> 
> https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons_luacontroller/init.lua#L63

Rename to `new_state`.

> > newstate is a global variable, L[port] should always be 0
> 
> Might that be the reason for spurious “on” events?

It looks like this is only used for `pin`.
I do not think that I've ever experienced those spurious “on” events.
This code branch only seems to be active if the luacontroller is `changesignal`ed with a table of rules. `event` will then also have a table of rules. I've never seen this happening, I think.

> ## base "mesecons" mod
> 
> `offstate` and `onstate` aren't defined anywhere:
> 
> https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons/internal.lua#L319
> 
> https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons/internal.lua#L335

This part of the function is never executed if the function is called on a conductor.
Suggestion: Just rise an error if it's not a conductor:
```lua
function mesecon.get_conductor_on(node_off, rulename)
	local conductor = mesecon.get_conductor(node_off.name)
	if not conductor then
		error("mesecon.get_conductor_on called for non-conductor", 2)
	end
	if conductor.onstate then
	-- ...
		return conductor.states[tonumber(binstate,2)+1]
	end
	error("mesecon.get_conductor_on called for conductor with bad definition", 2)
end 
```
Alternatively, you could just return `nil`.

numberZero:
> This part of the function is never executed if the function is called
on a conductor.

On conductor *in the expected state.*

> Suggestion: Just rise an error if it's not a conductor:

That would be reasonable for a new function but could break
compatibility with that-and-that mod. So are there any actual problems
with the current behavior of returning `nil` in that case?

Desour:
You are right, returning `nil` would be better.

It would also be possible to recreate the old intended behaviour: return `node_off` instead of `offstate`
See https://github.com/minetest-mods/mesecons/commit/885899cbaf6ccb2d22a3495131204ea0d01131ed#diff-aa0bed8d00c0792d0861e298584246e8R272.

.luacheckrc
BuckarooBanzay: it overwrites its own globalstep to delay it in some weird way:
https://github.com/minetest-mods/mesecons/blob/9fda51b650c83aea82616055ae8814a722353bf4/mesecons/actionqueue.lua#L128-L129

the global can be set read-only in another PR, if that issue is addressed and fixed

.luacheckrc
BuckarooBanzay: done

.luacheckrc
numberZero: Probably because of
```lua
                if not node then
                        -- Area does not exist; do nothing
                elseif mesecon.is_conductor_off(node, f.link) then
```
in a loop, and Lua lacks `continue`.

BuckarooBanzay: i don't intend to fix all bugs in that PR, i just like to get the basic infrastructure ready for continuous testing. The *existing* issues should be addressed in another PR (it worked like this for years)

numberZero:
So, what are the actual bugs discovered so far? I’d prefer to fix them before the release.

.luacheckrc
Desour: It might be better to allow this only for that one file.

BuckarooBanzay:
> So, what are the actual bugs discovered so far? I’d prefer to fix them before the release.

Just the above code-snippets under "Unresolved code-issues" everything else looks good so far :+1: 
Do you have a dead-line or something?

BuckarooBanzay:
The remaining issues are fixed and `luacheck` now runs without issues.

The next step after merging would be to enable stricter luacheck settings and fix those errors:
```lua
ignore = {
	"631", -- line too long
	"621", -- inconsistent indendation
	"542", -- empty if branch
}
```

Thanks for the reviews and suggestions, if there is anything else feel free to comment :+1:

SmallJoker:
@numberZero Could you please have a look at this PR again?

.github/workflows/luacheck.yml
```diff
@@ -0,0 +1,17 @@
+name: luacheck
```
Panquesito7, line 1: Maybe you could use [Factorio LuaCheck](https://github.com/marketplace/actions/factorio-mod-luacheck) here...it's faster and easier to use (see [ChaosWormz/teleport-request](https://github.com/ChaosWormz/teleport-request/blob/master/.github/workflows/build.yml) as an example). 🙂

.luacheckrc
```diff
@@ -0,0 +1,28 @@
+
```
Panquesito7, line 1:
```suggestion
allow_defined_top = true

```

BuckarooBanzay:
@numberZero can you take a look at this before this needs to be rebased please...?

PR #532, opened 2020-09-10
Add piston overheating

### LINKED ISSUE #499: Lag Non-contained Issues (Piston Driven Clocks)

numberZero:
Fix #499. Alternative to #526.
Pistons work at full speed but overheat if overloaded, just like most other items. Unlike those, a piston will cool down automatically, as requested by @Desour. The cooldown can be disabled if desired.
BTW should overheated pistons have different textures? Overheating isn’t immediately visible on many real-world devices (until you look inside) though on others it’s pretty obvious.

### DIFF
--- a/mesecons_pistons/init.lua
+++ b/mesecons_pistons/init.lua
@@ -2,11 +2,13 @@ local specs = {
 	normal = {
 		offname = "mesecons_pistons:piston_normal_off",
 		onname = "mesecons_pistons:piston_normal_on",
+		hotname = "mesecons_pistons:piston_normal_overheated",
 		pusher = "mesecons_pistons:piston_pusher_normal",
 	},
 	sticky = {
 		offname = "mesecons_pistons:piston_sticky_off",
 		onname = "mesecons_pistons:piston_sticky_on",
+		hotname = "mesecons_pistons:piston_sticky_overheated",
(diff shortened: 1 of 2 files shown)


SmallJoker:
> Overheating isn’t immediately visible on many real-world devices

This information could also be put into the infotext. It's less obvious than changed textures, but IMO good enough.

Desour:
Changes look fine btw. :+1:

TurkeyMcMac:
Perhaps pistons should build up heat both when extending and when retracting. This would seem to be more consistent with other components such as logic gates.

PR #534, opened 2020-09-12
Fix crash on recursive interrupt ID

### LINKED ISSUE #473: Luacontroller recursive table interrupt iid crash

numberZero:
Fix #473.
This ensures the interrupt ID, if it is a table, is a proper tree. A DAG wouldn’t crash the server either but checking for that is tricky.
This also deprecates non-string IIDs to allow future simplifications. Interrupt ID isn’t a place to store tables.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -266,6 +266,46 @@ local function remove_functions(x)
 	return x
 end
 
+local function validate_iid(iid)
+	if not iid then return true end -- nil is OK
+
+	local limit = mesecon.setting("luacontroller_interruptid_maxlen", 256)
+	if type(iid) == "string" then
+		if #iid <= limit then return true end -- string is OK unless too long
+		return false, "An interrupt ID was too large!"
+	end
+	if type(iid) == "number" or type(iid) == "boolean" then return true, "Non-string interrupt IDs are deprecated" end
+
+	local warn
+	local seen = {}
+	local function check(t)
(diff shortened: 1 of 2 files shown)


mesecons_luacontroller/init.lua
```diff
@@ -266,6 +266,46 @@ local function remove_functions(x)
...
+		if #iid <= limit then return true end -- string is OK unless too long
+		return false, "An interrupt ID was too large!"
+	end
+	if type(iid) == "number" or type(iid) == "boolean" then return true, "Non-string interrupt IDs are deprecated" end
+
+	local warn
+	local seen = {}
+	local function check(t)
+		if type(t) == "function" then
+			warn = "Functions can't be used in interrupt IDs"
```
sfan5, line 283:
```suggestion
			warn = "Functions cannot be used in interrupt IDs"
```

sfan5 (approved):
Looks good otherwise

PR #535, opened 2020-09-18
Add pitch variations for most noteblock sounds

Wuzzy2:
This PR adds pitch variations for almost all noteblock sounds with all semitones from C to B (like the piano).
Pitch is calculated with this formula:

```
pitch = 2^((val-6)/12)
```

Where `val` is a value from 0 (=C) to 11 (=B).

Piano is still using the custom piano sounds, but the other sounds use the Minetest `pitch` feature.

The fire sound (Lava Source) and explosion sound (Coal Block) still ignore pitch because these sound don't work well when pitched.
I changed the litecrash sound because it was the same as the crash sound but at a higher pitch. Since pitch is now

Finally, I made sure the noteblock is initialized at the C note instead C# when placed (using `place_param2`, so this change will be backwards-compatible).


Wuzzy2:
Bump.

mesecons_noteblock/init.lua
```diff
@@ -61,6 +62,18 @@ mesecon.noteblock_play = function(pos, param2)
 	pos.y = pos.y-1
 	local nodeunder = minetest.get_node(pos).name
 	local soundname = node_sounds[nodeunder]
+	local pitch
```
numberZero, line 65: `local use_pitch = true`

mesecons_noteblock/init.lua
```diff
@@ -61,6 +62,18 @@ mesecon.noteblock_play = function(pos, param2)
...
+	-- Special sounds
+	if not soundname then
+		for k,v in pairs(node_sounds_group) do
+			local g = minetest.get_item_group(nodeunder, k)
+			if g ~= 0 then
+				soundname = v
+				break
+			end
+		end
+	end
```
numberZero, line 75: Looks identical to the following block. Where’s the difference?

Desour, line 75: Exactly the same happens below.
(I guess you only wanted to add the comments, but copied by accident.)

mesecons_noteblock/init.lua
```diff
@@ -79,13 +92,26 @@ mesecon.noteblock_play = function(pos, param2)
...
+	-- Disable pitch for fire and explode because they'd sound too odd
+	if soundname == "fire_fire" or soundname == "tnt_explode" then
+		pitch = false
+	end
+	if pitch == false then
+		pitch = nil
+	else
+		-- Calculate pitch
+		local val = (param2+1)%12
```
numberZero, line 105: why?

mesecons_noteblock/init.lua
Desour, line 65: Maybe a more descriptive name:
```suggestion
	local use_pitch
```

mesecons_noteblock/init.lua
```diff
@@ -79,13 +92,26 @@ mesecon.noteblock_play = function(pos, param2)
 		if nodeunder == "default:steelblock" then
 			soundname = soundname.. 2
 		end
+		pitch = false
+	end
+	-- Disable pitch for fire and explode because they'd sound too odd
+	if soundname == "fire_fire" or soundname == "tnt_explode" then
+		pitch = false
+	end
+	if pitch == false then
```
Desour, line 101: Why not initialize `pitch` with `true` instead?

mesecons_noteblock/init.lua
```diff
@@ -79,13 +92,26 @@ mesecon.noteblock_play = function(pos, param2)
...
+		pitch = nil
+	else
+		-- Calculate pitch
+		local val = (param2+1)%12
+		-- All semitones from C to B (analog to piano mode)
+		pitch = 2^((val-6)/12)
```
numberZero, line 107: I doubt it’s from C to B. That would mean the original sound is F#, but is it?

mesecons_noteblock/init.lua
Desour, line 105: Comment would be good on why `+1` here. (Because `param2==11` is the lowest pitched sound.)

Desour (commented):
Pitch calculation is correct. (According to OpenAL docs, a factor of 2 means +12 semitones.)

You can use `mesecons_noteblock_fsharp` and `mesecons_noteblock_fsharp2`, and vary those by pitch, and drop all the other piano sounds (tested, it sounds exactly the same).
(Idk if other mods depend on the piano sounds, but I don't know of any. You can also just keep the files in the media folders (and not use them) for backwards compatibility.)

numberZero:
> > I doubt it’s from C to B. That would mean the original sound is F#, but is it?
> 
> I'm not a music expert. How about testing it? The note order is not changed, it is the same as for the already existing piano sounds.

Thing is, while octaves and semitones are relative things (octave = 12 semitones = 2× frequency) and are applicable to all sounds, notes aren’t. They are absolute¹. So what the code does is not pitching the sound **to** C, C#, D, ..., B (which wouldn’t even make sense for some sounds), it is pitching it **by** 6 semitones down, 5 semitones down, 4 semitones down, ..., 5 semitones up.

¹ Relative to A1 technically, but it is almost always 440 Hz nowadays.

Wuzzy2:
Thankfully, none of this is exposed to the player and is completely hidden behind the code.

So what do I have to do now to get this accepted? I don't know music theory and don't know what I must do next.

numberZero:
Remove this comment:
`-- All semitones from C to B (analog to piano mode)`
It’s likely possible to write a good one instead but that’s optional.

That aside, LGTM now.

Wuzzy2:
Done.

Wuzzy2:
Nice! Here's a follow-up PR, adding many new noteblock sounds: #635

PR #536, opened 2020-09-18
Add pulsar and flip-flop

### LINKED ISSUE #474: Add Something Akin To Redstone's Observer
In almost every circuit I make nowadays I find that I need pulses. Pulse when something comes on, pulse when something turns off, pulse to grab that block, or pulse to push that one back out. I find that for complex setups (mostly when making piston doors) that I can end up with multiple monostable circuits for pulsing at different times and this bloats up the size of things quite a bit as I then have to route wires around things and it all becomes quite a mess.

My idea is to add a block called the "Pulser" (or similar) which takes a mesecon input (as block updates are hard and slow to detect in Minetest) and outputs a pulse the length of a gate's delay (or the fastest possible monostable circuit as of now). This would help immensely with projects that need to pulse a piston, movestone, light, or shorten a signal at all. It would also help with routing signals as you can use them to pass a pulse through a chain of "Pulsers" and even split them in ways you can't easily without using mese block stacks (which then interact with nodes next to them).

numberZero:
Fixes #474. Also adds a simple memory cell.

The Pulsar detects a rising edge and outputs a pulse 2 Mesecon ticks long. Sharper (1-tick) pulse is possible but while e.g. gates and movestones work well with that, lamps and pistons don’t:
screenshot_20200918_224810

The Flip-Flop simply toggles its state on a rising edge.

Everything seems to work well, testing is welcome. Crafting and documentation are pending)


numberZero:
Any ideas for crafting recipes?

Also if there should be 3 kinds of pulsars, what could their images look like?

Desour:
:+1: Code looks fine. Also tested, it works well.
(You might want to use my meseconomenter mod for testing the pulsar: https://github.com/Desour/meseconometer (Just give a mesecons signal to the activation port (default A), then give some signals to the device and rightclick to see the changes logged based on globalstep time.))

Crafting recipes could include what one would normally use to build the devices, eg. nor gates and insulated crossover for the flip flop; NOT gate, AND gate and wire for the pulsar.
For 3 kinds of pulsars, you could add recipes to craft one to the other: rising-edge-detector -> falling-edge-detector -> any-edge-detector -> rising-edge-detector (And additionally eg. NOT gate, NOR gate and wire for the falling-edge-detector.)
For the images you could write a 1 and a 0 (or o and I) onto the pulsars.

PR #539, opened 2020-10-06
Use monospaced font for Lua controller

johalun:
Monospaced font option for textareas, etc is now available in Minetest master branch. Make use of it for the Lua controller's code code area and error label.

I also tested this with MT 5.1.1 and it didn't not complain about the added style_type so I take it is silently ignored if not supported.


SmallJoker (approved):
<3 Thank you!

auouymous:
Why did you close this before it was merged?

johalun:
Oops sorry thought it was merged...

PR #545, opened 2020-10-22
Mvps: make mvps stoppers non-replaceable even if they are buildable_to

### LINKED ISSUE #544: mvps: you can push things out of the edge of the world and pull them back again
`ignore` nodes seem to be mvps stoppers already, that doesn't help.

Desour:
Fixes #544.

For testing please note that if you've already pushed out of world, there's air now, so you'll have to move your test piston.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -55,16 +55,16 @@ end
 
 -- tests if the node can be pushed into, e.g. air, water, grass
 local function node_replaceable(name)
-	if name == "ignore" then
-		-- ignore is buildable_to, but we do not want to push into it
+	local nodedef = minetest.registered_nodes[name]
+
+	-- everything that can be an mvps stopper (unknown nodes and nodes in the
+	-- mvps_stoppers table) must not be replacable
+	-- Note: ignore (a stopper) is buildable_to, but we do not want to push into it
+	if not nodedef or mesecon.mvps_stoppers[name] then
 		return false
 	end
 
-	if minetest.registered_nodes[name] then
-		return minetest.registered_nodes[name].buildable_to or false
-	end
-
-	return false
+	return nodedef.buildable_to or false
 end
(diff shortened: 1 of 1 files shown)


fluxionary:
Unless you can check against all mvps stopper nodes and not just `ignore` (which looks to be a little tricky), it's probably best to log a warning if you register a node w/ `buildable_to = true` as an mvps stopper.

Desour:
You are right. All mvps stoppers should not be replacable, will do.

fluxionary:
The update has a problem, because the value of `mesecon.mvps_stoppers[name]` can be a function that needs to be evaluated to determine if the node is actually a stopper, and this fix will e.g. unconditionally make extend pistons into stoppers, when that should only be a problem when they're extended in a perpendicular direction to the push.

Desour:
> The update has a problem, because the value of `mesecon.mvps_stoppers[name]` can be a function that needs to be evaluated to determine if the node is actually a stopper, and this fix will e.g. unconditionally make extend pistons into stoppers, when that should only be a problem when they're extended in a perpendicular direction to the push.

I know about this. But it should be no problem as `node_replaceable` should return `false` for piston heads anyways. The only difference is in nodes that are `buildable_to` and stoppers. I don't know if nodes where you can push into from one side, but where you can't move from the other side are a thing.

fluxionary:
Hm. I suppose this would only be a problem for `buildable_to` nodes that are conditionally stoppers, and I've got a feeling that doesn't actually exist in the wild... but it could?

numberZero:
Alternative: #546. The difference is that it handles `ignore` unconditionally, and doesn’t handle `buildable_to` stoppers at all (but these shouldn’t exist in the first place IMO so it warns on attempt to register one).

Desour:
I think, the old code would just have pushed into such nodes, so it maybe even couldn't.

PR #548, opened 2020-12-04
Prevent unauthorized players from toggling blinky plant.

auouymous:


Panquesito7 (approved):

mesecons_blinkyplant/init.lua
```diff
@@ -32,7 +32,13 @@ mesecon.register_node("mesecons_blinkyplant:blinky_plant", {
 		fixed = {-0.3, -0.5, -0.3, 0.3, -0.5+0.7, 0.3},
 	},
 	on_timer = on_timer,
-	on_rightclick = toggle_timer,
+	on_rightclick = function(pos, node, clicker)
+		if minetest.is_protected(pos, clicker and clicker:get_player_name()) then
```
MoNTE48, line 36: `clicker and clicker:get_player_name() or ""`

auouymous: Can clicker even be nil, other than buggy mods that directly call it? Mesecons is the only mod I have that even checks for an entity before calling get_player_name, and none of them (including mesecons) append `or ""`.

MoNTE48: Since 5.3 - sometimes it can.

auouymous: Fixed.

sfan5 (approved):
Looks good.

SmallJoker:
:+1: makes sense. Players should not influence protected circuits.

PR #556, opened 2021-03-09
Improve performance of mesecon.turnon and mesecon.turnoff

Desour:
### Improvements:
- Use a better fifo queue for `frontiers`. (Old one is still included as separate implementation for testing, remove before merging!)
  (Related: https://github.com/minetest/minetest/pull/7683 @HybridDog)
- Nodes that we certainly don't need to visit again (on-/off-turned conductors and nodes that are ignored (eg. air)) are marked as such via `pos_can_be_skipped` (a hashset using the hashed node pos), and are not visited again.
  This affects the `depth`, but what is that anyway?
- `mesecon.turnoff` had called `mesecon.is_receptor_on` in a loop for no reason. It was probably a copy-error.

(There's probably much more that could be done.)

### How to test:
Use worldedit to make a big conductor block. And turn it on and off.
For time measuring I've used globalstep dtime:
```lua
local function printc(msg)
	print(msg)
	minetest.chat_send_all(msg)
end

minetest.register_globalstep(function(dt)
	if dt < 0.1 then
		return
	end
	printc("globalstep dt: "..dt)
end)
```

The conductor cuboid I've used `has a volume of 45000 nodes (30*30*50)` and consists of mese and `wire_11111111_off`:

screenshot_20210309_233911

### Test results:
(My cpu was scaled to 1600MHz, I'm using luajit.)
branch | turnon | turnoff
---|---|---
this PR | ca. 1.3 s | ca. 1.6 s
master | ca. 12 s | ca. 14 s


mesecons/internal.lua
```diff
@@ -419,44 +429,49 @@ end
 --	depth = indicates order in which signals wire fired, higher is later
 -- }
 function mesecon.turnoff(pos, link)
-	local frontiers = {{pos = pos, link = link}}
+	local frontiers = fifo_queue.new()
+	frontiers:add({pos = pos, link = link})
 	local signals = {}
+	local pos_can_be_skipped = {}
```
SmallJoker, line 435: Wouldn't it make sense to include this duplicate check functionality into `fifo_queue`? I think that would be more reliable than adding the cached positions manually.

Desour: That's not what a fifo queue usually does.
And some positions need to be added multiple times: eg. effectors, think of a gate that needs to be activated from two sides.

SmallJoker (approved):

PR #560, opened 2021-03-25
Fix on-state receptors (eg. gates) recepting to everywhere

### LINKED ISSUE #559: Wires going into FPGAs never turn off
Following setup:
`[FPGA] [wire]`
Gate config:
`  = C -> A`

Put a power plant next to the wire, remove the plant again and observe how the wire is still on?!

Broken by #556 @Desour I'm pretty sure I tested this before but I guess not .

Desour:
Fixes #559.

In #556 I've removed a loop in `mesecon.turnoff` and totally forgot that it is very possible that there is no rule to the pos at all (when the receptor doesn't recept to that position) which results in the loop not getting executed even once. This resulted in on-state receptors recepting to any rules.

## How to test:
```
  P
  D
XWW

or

>
B
W
X

P = powerplant
D = OR gate
W = wire
X = see below
> = NOT gate
B = mese block
```
Place a powerplant at X and dig it, and see if the wire goes off again.

### DIFF
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -449,8 +449,10 @@ function mesecon.turnoff(pos, link)
 					-- Check if an onstate receptor is connected. If that is the case,
 					-- abort this turnoff process by returning false. `receptor_off` will
 					-- discard all the changes that we made in the voxelmanip:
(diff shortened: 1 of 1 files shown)


sfan5 (approved):
Works.

Issue #561, opened 2021-03-28
Logic gates don't overheat


VanessaE:
All you have to do is connect the output of a NOT gate back to its input, and the circuit will sit there cycling until you break it.

That is, unless you set `dedicated_server_step` to something ridiculously low, like 0.025 seconds (that is, a quarter of its usual setting of 0.1s).  Then, such a circuit will overheat in a few seconds and shut down.

cheapie:
From what I've seen so far, this problem only happens on servers - in singleplayer, logic gates seem to run much faster and overheat like they should.

Desour:
A NOT gate where the input is connected should get one heat every 2 steps.
Assuming default settings, and a `dedicated_server_step` of 0.1s:
The cooldown time is 20 heat / 2 s = 10 h/s.
Heat generation is 1 heat / (2 steps * 0.1 second / step) = 5 h/s.

Server owners should probably change their settings, eg. increase `cooldown_time` to 8.5 seconds.

Changing the default settings would probably result in overheating circuits, especially in singleplayer.

Imo it would be better if overheat was in server steps rather than in seconds. It would be much more reliable, and one could build circuits that are guaranteed not to overheat.

numberZero:
> Imo it would be better if cooldown was in server steps rather than in seconds. It would be much more reliable, and one could build circuits that are guaranteed not to overheat.

LuaC interrupts would also need to be in steps for full reliability.

> ridiculously low, like 0.025 seconds

That's only 40 UPS. For comparison, Factorio runs at 60 UPS, with all its complexity. Something... just doesn't feel right.

PR #562, opened 2021-03-28
Add setting to have overheat cooldown in steps

Desour:
## Problems to be solved:

Currently the cooldown is done in steps of seconds and the time it takes a device to fully cooldown is also in seconds.
Gates however have a delay in server steps (always exactly 2 steps). This makes it nearly impossible to build circuits that can never overheat, in any circumstances.
Also, servers have a much longer server step dtime, which results in inconsistencies and bad defaults (see also #561).

## Changes:

- A new setting `cooldown_mode` (enum: seconds(default), steps) says whether to use the old or the new method.
- Two new settings `cooldown_time_steps` (float, default=125.0) and `cooldown_granularity_steps` (int) can be used to configure the new mode, they work like their non-`_steps` counterparts.
- Rationale for not reusing the old settings: The values for `*_steps` are typically much higher. If you use an older mesecons version, the `cooldown_mode` setting is ignored and circuits will overheat. Also it's much more convenient to have separate settings if you want to play around with them. Furthermore, the new settings allow to define suitable defaults.
- Choice of values:
  125 steps is 2 seconds for a step time of 0.016 seconds. I was not able to measure a lower step time in singleplayer.
  This is a pretty strict value compared to the old settings. But it isn't actually _that_ strict, it allows one heat every 6.25 steps, which is enough for a shortcut NOT gate plus two diode gates for delay.
  The granularity of 20 steps if the same as 0.5 seconds with a dtime of 0.025 seconds.
  For servers the seconds-based cooldown would be ca. every 5 steps. So this value is more performance-friendly to servers.
  I currently don't know exactly if this value is good, but it's probably good enough.

## Other changes:

- Added documentation of settings in settingtypes.txt.
- Removed arbitrary limits in settingtypes.txt.
- Check lower bounds for settings in lua.
- Use float type for `overheat_max` in settingtypes.txt.

### DIFF
--- a/mesecons/services.lua
+++ b/mesecons/services.lua
@@ -74,9 +74,16 @@ minetest.register_on_placenode(mesecon.on_placenode)
 minetest.register_on_dignode(mesecon.on_dignode)
 
 -- Overheating service for fast circuits
-local OVERHEAT_MAX = mesecon.setting("overheat_max", 20)
-local COOLDOWN_TIME = mesecon.setting("cooldown_time", 2.0)
-local COOLDOWN_STEP = mesecon.setting("cooldown_granularity", 0.5)
+local OVERHEAT_MAX = math.max(1.0, mesecon.setting("overheat_max", 20.0))
+local COOLDOWN_MODE = mesecon.setting("cooldown_mode", "seconds")
+local COOLDOWN_IN_STEPS = COOLDOWN_MODE == "steps"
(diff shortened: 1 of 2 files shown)


sfan5:
Shouldn't this just be the default?

Desour:
Yes sure. I was just scared by the possibility of people complaining about a change.

steps is now the default.

PR #563, opened 2021-04-08
Fix use_texture_alpha warnings

Desour:
Use everywhere `"opaque"`, but `"clip"` for `hydro_turbine_on` (tested, of course).


sfan5 (approved):
Tested.

PR #564, opened 2021-04-08
Add transistors

Desour:
Adds new nodes are conducting or not depending on whether a special effector next to them is activated.
The player always places two nodes at once with the node, and digs both at once. (TNT can blast away one part, but the sinlge part causes no crashes or other problems.)
This works pretty much exactly like a piston that moves a conductor, the timing is also exactly the same (tested with my meseconometer mod). But these new nodes are smaller, make no noise, overheat and are less expensive for the server (no mvps movement).
There are the normally opened (N/O) and the normally closed (N/C) conductor. Closed means that the node is conducting. The N/O conductor only conducts if it's on, the N/c only if it's off.

Screenshot:
screenshot_20210408_233344

Please feel free to make suggestions for crafting recipes!

If this is not considered suitable for mesecons, I will release this as separate mod, or ask for inclusion in moremesecons.


ssieb:
Why not do it with a single node?  I don't see a reason for the extra one.

Desour:
Ah, I forgot to explain that.
A node can not be a conductor and an effector at the same time. Mesecons doesn't support this.

ssieb:
I think I understand now what you're trying to do, but I don't really see the use of it.  This is not a transistor, it's more like a solid-state relay.  (The piston would be a better match for a mechanical relay.)  You want bi-directional transmission.  What's the use-case for that?

auouymous:
A piston-wire combo can conduct around a corner. This would be better if it conducted through all three faces instead of only the two sides.

w = wire, C = conductor node, B = base node
```
w
CwCw
B B
```

ssieb:
The other gates also have instant transmission.  I just tested it out.  I put two AND gates in series and it updates at the same time as a piston controlled switch.  Have you had a problem with that in the past?

ssieb:
Just for fun, I implemented two bits of a full adder.  I don't see why there's a problem.
Screenshot from 2021-04-08 19-35-30

Desour:
>A piston-wire combo can conduct around a corner. This would be better if it conducted through all three faces instead of only the two sides.

I have thought about making it conduct to 3 sides. But you can achieve more conductor sides easily by using multiple conditional conductors / transistors / relays / whatever (see the construction in the back of my screenshot). I've decided to only have two conductor sides for simplicity. And you can build smaller circuits if you can have uninsulated conductors (like mese blocks or vertical mescons) next to it. But I don't have a strong opinion about this.

Desour:
>The other gates also have instant transmission. I just tested it out. I put two AND gates in series and it updates at the same time as a piston controlled switch. Have you had a problem with that in the past?

Well, the delay is short. But it is very noticeable if you put many gates in succession.

I have some screenshots for you:

If don't believe that gates have a delay, try something like this:
screenshot_20210409_175010

If you only use AND gates and no pistons, you will build things like this:
screenshot_20210409_175001
There will always be a delay from one side, eg. from the left switch. But with pistons, you can have 0 delay when you activate the lower switches.

The adder I was referring to looks like this:
screenshot_20210409_175411

But as it's too hard to see anything there, here's an image single bit adder cell:
screenshot_20210409_175251
Input comes from down, carry comes from right and goes to left, output is up.
With the help of the piston you can have the carry bit available at every cell in constant time.

----------------------------------------

Btw. another difference of conditional conductors is that the conductor doesn't overheat, which can be useful too.

numberZero:
Looks nice. Not the diff stats though.
How does this work, makes the second node a conductor/non-conductor depending on the input?

> This is not a transistor, it's more like a solid-state relay

Sounds enough like a FET for me. Totally not a BJT of course.

> A node can not be a conductor and an effector at the same time. Mesecons doesn't support this.

File a bug.

PR #572, opened 2021-07-17
Fix conductor turnon link direction within on_placenode

TurkeyMcMac:
When testing this, I put this line at the beginning of `mesecon.turnon` to track calls to it:

```lua
minetest.chat_send_all(minetest.pos_to_string(pos) .. " " .. minetest.pos_to_string(link))
```

Then I placed a switch, turned it on, placed a regular wire adjacent to it, and turned the switch off and on again. With this PR, both calls to `turnon` output the same debugging information. Without it, the values of `link` are opposite to each other.

In other words, this PR makes the behavior when placing a conductor consistent with the behavior when turning on a receptor, with respect to the turned-on conductor.

I don't think the bug occurs with any built-in conductors, but it affects a mod I'm working on.


SmallJoker:
:+1: LGTM. Will merge in a few days unless there are objections.

PR #573, opened 2021-07-28
Only skip conductors with flat rules in turnon/off

TurkeyMcMac:
This is a possible solution to https://github.com/minetest-mods/mesecons/issues/571. It implements what I suggested in the description of that issue. These changes seem not to break anything judging by my testing of some of the circuits from the game Mesecons Lab. If some kind of test suite is available, it would be good to run these changes by that.

This patch may slow things down due to reducing the caching, but I haven't done a rigorous benchmark. I didn't notice any big performance changes. Again, is there some benchmark suite available?

### DIFF
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -384,21 +384,27 @@ function mesecon.turnon(pos, link)
 		if not node then
 			-- Area does not exist; do nothing
 			pos_can_be_skipped[minetest.hash_node_position(f.pos)] = true
-		elseif mesecon.is_conductor_off(node, f.link) then
+		elseif mesecon.is_conductor(node.name) then
 			local rules = mesecon.conductor_get_rules(node)
 
-			-- Call turnon on neighbors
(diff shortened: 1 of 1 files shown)


TurkeyMcMac:
Looking at past pull requests, they don't seem to use specific tests or benchmarks, so maybe I should do some more testing and benchmarking of my own. I'm not currently in a position where I can do so, but I will be next week probably.

Desour:
You can look at  #556 to get some ideas on how this could be benchmarked. (Though, realistic circuits are probably much more linear.)

Actually, I think, I had not even known that the `states` thing (as opposed to `state`) was a thing (documentation of mesecons could be much better).


Change works. :+1:

TurkeyMcMac:
Ok, great! I think you've done enough benchmarking, so I'm done with this PR unless someone wants more benchmarking/testing done.

(I agree that some more documentation would be nice.)

TurkeyMcMac:
Is there anything else I should do for this PR before it can be merged?

PR #578, opened 2021-11-16
Optimize light updates when turning conductors on and off

TurkeyMcMac:
## Overview

Most calls to `VoxelManip:write_to_map` made from `turnon` and `turnoff` do not need to include a light update, since most conductors do not change their lighting characteristics depending on their state. There are exceptions, including mese blocks and ghoststone blocks. Such exceptions are automatically handled in this PR. Most unnecessary light updates have been removed.

## Benchmarks

My changes create a measurable performance improvement. To measure them, I made this change to `mesecons/actionqueue.lua`:

```patch
diff --git a/mesecons/actionqueue.lua b/mesecons/actionqueue.lua
index 5508095..1ccddd5 100644
--- a/mesecons/actionqueue.lua
+++ b/mesecons/actionqueue.lua
@@ -70,6 +70,8 @@ end
 -- However, even that does not work in some cases, that's why we delay the time the globalsteps
 -- start to be execute by 4 seconds
 
+local sum = 0
+
 local function globalstep_func(dtime)
 	local actions = queue.actions
 	-- split into two categories:
@@ -106,10 +108,23 @@ local function globalstep_func(dtime)
 		end
 	end)
 
+	local has_actions = actions_now[1] ~= nil
+
+	local before
+	if has_actions then before = minetest.get_us_time() end
+
 	-- execute highest priorities first, until all are executed
 	for _, ac in ipairs(actions_now) do
 		queue:execute(ac)
 	end
+
+	local after
+	if has_actions then after = minetest.get_us_time() end
+
+	if has_actions then
+		sum = sum + (after - before)
+		minetest.chat_send_all(sum / 1000000)
+	end
 end
 
 -- delay the time the globalsteps start to be execute by 4 seconds
```

This prints the total action queue runtime after any action executes. You have to restart the server to reset the sum.

I ran these with and without the PR applied. Your benchmark results will probably differ from mine.

### Long line of diodes

WorldEdit schematic: [example-diodes.we.gz](https://github.com/minetest-mods/mesecons/files/7550372/example-diodes.we.gz)

To test this, turn the switch off then on again.

Runtime without PR: 0.53s
Runtime with PR: 0.29s

### Block of wires

WorldEdit schematic: [example-wireblock.we.gz](https://github.com/minetest-mods/mesecons/files/7550394/example-wireblock.we.gz)

To test this, turn the switch on and off again six times.

Runtime without PR: 5.717s
Runtime with PR: 5.70s

### 256 byte memory bank

WorldEdit schematic: [example-memory.we.gz](https://github.com/minetest-mods/mesecons/files/7550403/example-memory.we.gz)

Unlike the other benchmarks, this one is of a device with an actual purpose. The controls are at the northwest corner of the device. Before testing, turn on the leftmost switch and the switch at the bottom of the rightmost stack of switches.

To test this, turn the switch at the top of the rightmost stack on and off again six times. For consistency, make sure the circuitry settles down before each flip of the switch.

Runtime without PR: 5.21s
Runtime with PR: 3.64s

## Tests

These test functionality rather than performance.

### Mese block

WorldEdit schematic: [example-meseblock.we.gz](https://github.com/minetest-mods/mesecons/files/7550506/example-meseblock.we.gz)

The light emitted by the mese block changes depending on its conduction state. This test confirms that the light update occurs.

To run this test, enter the dark chamber and flip the switch.

### Ghoststone

WorldEdit schematic: [example-ghoststone.we.gz](https://github.com/minetest-mods/mesecons/files/7550509/example-ghoststone.we.gz)

Whether or not ghoststone lets light through depends on its conduction state. This test confirms that the light update occurs.

To run this test, enter the dark chamber and flip the switch.

### DIFF
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -368,46 +368,59 @@ function mesecon.is_power_off(pos, rulename)
 	return false
 end
 
--- This is the callback for swap_node_force in turnon and turnoff. It determines
--- whether a conductor node necessitates a lighting update.
-local cached_update_light = {}
-local function get_update_light_conductor(pos, name)
-	local update_light = cached_update_light[name]
+-- The set of conductor states which require light updates when they change.
+local light_update_conductors
 
-	if update_light == nil then
-		-- Calculate and cache whether the conductor needs light updates.
+-- Calculate the contents of the above set if they have not been calculated.
+-- This must be called before get_update_light_conductor.
+local function find_light_update_conductors()
+	-- The expensive calculation is only done the first time.
+	if light_update_conductors then return end
 
-		local def = minetest.registered_nodes[name]
+	light_update_conductors = {}
 
+	-- Find conductors whose lighting characteristics change depending on their state.
+	local checked = {}
+	for name, def in pairs(minetest.registered_nodes) do
 		local conductor = mesecon.get_conductor(name)
-		local other_states
-		if conductor.onstate then
-			other_states = {conductor.onstate}
-		elseif conductor.offstate then
-			other_states = {conductor.offstate}
-		else
-			other_states = conductor.states
-		end
+		if conductor then
+			if not checked[name] then
+				-- Find the other states of the conductor besides the current one.
+				local other_states
+				if conductor.onstate then
+					other_states = {conductor.onstate}
+				elseif conductor.offstate then
+					other_states = {conductor.offstate}
+				else
+					other_states = conductor.states
+				end
 
-		update_light = false
-		for _, other_state in ipairs(other_states) do
-			local other_def = minetest.registered_nodes[other_state]
-			if (def.paramtype == "light") ~= (other_def.paramtype == "light")
-			or def.sunlight_propagates ~= other_def.sunlight_propagates
-			or def.light_source ~= other_def.light_source then
-				-- The light characteristics change depending on the state.
-				update_light = true
-				break
+				-- Check the conductor. Other states are marked as checked.
+				for _, other_state in ipairs(other_states) do
+					local other_def = minetest.registered_nodes[other_state]
+					if (def.paramtype == "light") ~= (other_def.paramtype == "light")
+					or def.sunlight_propagates ~= other_def.sunlight_propagates
+					or def.light_source ~= other_def.light_source then
+						-- The light characteristics change depending on the state.
+						-- The states are added to the set.
+						light_update_conductors[name] = true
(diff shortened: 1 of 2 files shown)


mesecons/internal.lua
```diff
@@ -368,6 +368,48 @@ function mesecon.is_power_off(pos, rulename)
...
+				update_light = true
+				break
+			end
+		end
+
+		cached_update_light[name] = update_light
+		for _, other_state in ipairs(other_states) do
+			cached_update_light[other_state] = update_light
+		end
+	end
```
sfan5, line 408: Have you considered caching the value for all conductors once at startup?
The table could also then be smaller with `nil` meaning false.

TurkeyMcMac: Good idea.

mesecons/internal.lua
```diff
@@ -368,11 +368,68 @@ function mesecon.is_power_off(pos, rulename)
...
+	if light_update_conductors then return end
+
+	light_update_conductors = {}
+
+	-- Find conductors whose lighting characteristics change depending on their state.
+	local checked = {}
+	for name, def in pairs(minetest.registered_nodes) do
+		local conductor = mesecon.get_conductor(name)
+		if conductor then
+			if not checked[name] then
```
sfan5, line 387: You can merge these two conditions

mesecons/internal.lua
```diff
@@ -434,6 +491,8 @@ end
 --	depth = indicates order in which signals wire fired, higher is later
 -- }
 function mesecon.turnoff(pos, link)
+	find_light_update_conductors()
```
sfan5, line 494: I suggest just doing this in `minetest.register_on_mods_loaded` instead of adding overhead to turnoff/turnon

TurkeyMcMac: I chose not to use `minetest.register_on_mods_loaded`, the reason being that (unless I'm mistaken) other mods can change node properties in `minetest.registered_on_mods_loaded` callbacks using `minetest.override_item`. I want to be sure the node property values read are the final values.

The overhead of the current approach is very small--just one extra nil check per call to `turnon`/`turnoff`--because the set of conductors requiring light updates is only built the first time.

sfan5: Hmm, that's unfortunate.

sfan5 (approved):
Tested a bit. Works.

TurkeyMcMac:
Can this be merged now?

sfan5:
Sure.

PR #581, opened 2022-01-18
Fix torch rules

### LINKED ISSUE #580: Torch logic directional bug
The torch only deactivates if powered from below on the -X axis
image

Otherwise it does nothing
image

The bug happens with any activation source be it other torches, mesecon lines or insulated lines
image
image
 
I'm using Minetest 5.4.1, Singleplayer mode.

Desour:
Fixes #580.

The extra rule is probably there to allow using conductors on top of the node that the torch is placed to:
screenshot_20220119_002941

But this rule should not be used for vertical torches.

Test by placing torches.

### DIFF
--- a/mesecons_torch/init.lua
+++ b/mesecons_torch/init.lua
@@ -27,9 +27,13 @@ local torch_get_output_rules = function(node)
 	return rotate_torch_rules(rules, node.param2)
 end
 
+local torch_input_rules_unrotated_horizontal = {vector.new(-2, 0, 0), vector.new(-1, 1, 0)}
+local torch_input_rules_unrotated_vertical   = {vector.new(-2, 0, 0)}
+
 local torch_get_input_rules = function(node)
-	local rules = 	{{x = -2, y = 0, z = 0},
-				 {x = -1, y = 1, z = 0}}
+	local rules = (node.param2 == 0 or node.param2 == 1)
+		and torch_input_rules_unrotated_vertical
+		or  torch_input_rules_unrotated_horizontal
 
 	return rotate_torch_rules(rules, node.param2)
 end


sfan5 (approved):

VorTechnix:
Just tested this change and it made the problem worse. Now torches aren't turned off by torches 1 block below them no matter which side they're on. This change would move the mesecon torch further away from the redstone torch's behavior in Minecraft which is the opposite of what I was trying to achieve by opening issue #580.

Desour:
@VorTechnix Just to make sure, 2 nodes below the torch does work for you, right?

I don't know how exactly the redstone torches in minecraft work, but mesecons doesn't exactly replicate redstone anyways.
The problem with changing rules is always that existing builds can be broken, so this PR's intention was never to add 1-node-below rules but to remove a very weird asymmetric behaviour.
The commit that added the "behind above" rules (https://github.com/minetest-mods/mesecons/commit/ac0fb9113997e87c724ed5ece9e9914936ed6a3b) called it "behind above", and this name makes not much sense for vertical torches. So removing it for vertical torches can be considered as bugfix and will likely not break any build.

VorTechnix:
@Desour No you misunderstood the problem. It was incorrect interactions between 2 vertical torches that was glitched. The behind above rule is for horizontal torches only (or if not it should be).

Here is a picture showing the expected interaction between vertical torches offset 1 block horizontally and one block vertically:
image

I achieved the result using the following code:
```lua
local torch_input_rules_unrotated_vertical   = {
	-- Somehow X and Y have been switched so global Y -1 is X -1 to mesecon torches
	vector.new(-1, 0, 1),
	vector.new(-1, 0, -1),
	vector.new(-1, 1, 0),
	vector.new(-1, -1, 0)
}
```

VorTechnix:
However, upon further testing, the above code breaks the interaction between vertical torches and horizontal ones. This is most frustrating.

Desour:
> @Desour No you misunderstood the problem. It was incorrect interactions between 2 vertical torches.

Torches don't interact with other torches differently than with any other mesecons things. Could you please elaborate?

> The behind above rule is for horizontal torches only (or if not it should be).

I agree, and this PR does this.

> Here is a picture showing the expected interaction between vertical torches offset 1 block horizontally and one block vertically: [img]

I disagree about that being expected behaviour.
As already said:
>this PR's intention was never to add 1-node-below rules

Because of the bugginess of the current 1-node-below rules (and because it only works in 1 of 4 directions), it is very unlikely that builds make use of it. It is more likely that there are builds where your suggested rules cause unintentional torch deactivation.

VorTechnix:
> Because of the bugginess of the current 1-node-below rules (and because it only works in 1 of 4 directions), it is very unlikely that builds make use of it. It is more likely that there are builds where your suggested rules cause unintentional torch deactivation.

I agree to an extent but then what about builds like this (torch based AND gate) that require side and below activation of vertical torches?
image

VorTechnix:
Anyway for now the changes you've made in this PR remove inconsistency which I guess technically solves the issue I opened. That said I think there should be some discussion about some sort of implementation of node below logic (maybe as a different torch).

Issue #517, opened 2020-06-28
Gates do not update when rotated


Desour:
If you rotate a gate with mtg's screwdriver, the connected wires do not update correctly.

Example:
![screenshot_20200628_141721](https://user-images.githubusercontent.com/7613443/85948044-b5636880-b94e-11ea-818c-8aa6f7605b67.png)

PR #583, opened 2022-01-25
Store gate inputs in param2 instead of metadata

TurkeyMcMac:
All logic gates have paramtype2 = "facedir". The upper three bits of param2 are therefore unused. However, instead of using these bits to store inputs, binary logic gates store their inputs in metadata, using more than 14 extra bytes of storage per logic gate node. This PR aims to change that in a backward-compatible way.

When a binary logic gate is updated, bit 5 of its param2 is checked. Bit 5 being set to 1 indicates that the logic gate is using the new implementation. In this case, bit 6 is the value of input1 and bit 7 is the value of input2. If bit 5 is 0, then the logic gate is either old or has just been placed. In this case, the binary logic gate gets migrated to the new implementation. Bit 5 is then set and the metadata is cleared if it is present.

I decided not to use LBMs to do the migration. Since LBMs run when a block gets activated, they are not suitable for circuit components. While Mesecons is executing, it may load mapblocks and make use of the circuit components inside them without activating the mapblocks. This means that LBMs may not run when they are needed.

Here's a benchmark: [ors.we.gz](https://github.com/minetest-mods/mesecons/files/7936581/ors.we.gz). The line of OR gates whose switch is marked "old" by the sign on top should only be used without the patch applied. The line marked "new" should only be used with the patch applied. The new version seems to be about the same speed, but I didn't benchmark too carefully.

---

I also fixed https://github.com/minetest-mods/mesecons/issues/569 with regards to logic gates. Binary gates no longer overheat just by being pushed.

### DIFF
--- a/mesecons_gates/init.lua
+++ b/mesecons_gates/init.lua
@@ -23,16 +23,18 @@ local gate_get_input_rules_twoinputs = mesecon.horiz_rules_getter({
 local function set_gate(pos, node, state)
 	local gate = minetest.registered_nodes[node.name]
 
-	if mesecon.do_overheat(pos) then
-		minetest.remove_node(pos)
-		mesecon.receptor_off(pos, gate_get_output_rules(node))
-		minetest.add_item(pos, gate.drop)
-	elseif state then
-		minetest.swap_node(pos, {name = gate.onstate, param2=node.param2})
-		mesecon.receptor_on(pos, gate_get_output_rules(node))
-	else
-		minetest.swap_node(pos, {name = gate.offstate, param2=node.param2})
-		mesecon.receptor_off(pos, gate_get_output_rules(node))
+	local new_nodename = state and gate.onstate or gate.offstate
+	minetest.swap_node(pos, {name = new_nodename, param2 = node.param2})
+	if new_nodename ~= node.name then
+		if mesecon.do_overheat(pos) then
+			minetest.remove_node(pos)
+			mesecon.receptor_off(pos, gate_get_output_rules(node))
+			minetest.add_item(pos, gate.drop)
+		elseif state then
+			mesecon.receptor_on(pos, gate_get_output_rules(node))
+		else
+			mesecon.receptor_off(pos, gate_get_output_rules(node))
+		end
 	end
 end
 
--- a/mesecons_gates/init.lua
+++ b/mesecons_gates/init.lua
@@ -62,9 +62,9 @@ local function update_gate(pos, node, link, newstate)
 			local meta = minetest.get_meta(pos)
 			if link.name == "input1" then
 				val1 = newstate == "on"
-				val2 = meta:get("input2") == "1"
+				val2 = meta:get_int("input2") == 1
 			else
-				val1 = meta:get("input1") == "1"
+				val1 = meta:get_int("input1") == 1
 				val2 = newstate == "on"
 			end
 			-- Set bit 5 so this won't happen again.
--- a/mesecons_gates/init.lua
+++ b/mesecons_gates/init.lua
@@ -42,11 +42,36 @@ local function update_gate(pos, node, link, newstate)
 	if gate.inputnumber == 1 then
 		set_gate(pos, node, gate.assess(newstate == "on"))
 	elseif gate.inputnumber == 2 then
-		local meta = minetest.get_meta(pos)
-		meta:set_int(link.name, newstate == "on" and 1 or 0)
-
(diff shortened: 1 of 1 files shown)


Desour:
:+1: Code looks fine, concept too.
- (You could use the now available bitop functions, and provide fallbacks. But that would probably make the code more complicated.)
- (Adding a `on_construct` to set the 5th bit might make things cleaner, but also adds more code.)
- I'd suggest to test this with screwdrivers and add an `on_rotate` callback, but gates are broken in that regard anyways (#517).

TurkeyMcMac:
Thanks. I think I'll leave the code as it is for now with regards to `on_construct`. `on_construct` might run on non-updated gates which have just been moved, and the alternative, `after_place_node`, may not reliably handle changes to param2 made within the callback (the stale value of param2 is passed to global `on_placenode` callbacks.)

I've recently been trying to fix gate rotation, among other things. I'll make a PR for this if I'm successful.

sfan5:
> I decided not to use LBMs to do the migration. Since LBMs run when a block gets activated, they are not suitable for circuit components. While Mesecons is executing, it may load mapblocks and make use of the circuit components inside them without activating the mapblocks.

I think you're incorrect about this, surely LBMs do what their name implies (run on *load*) and lua_api.txt contradicts you too.

TurkeyMcMac:
I could be wrong, but `applyLBMs` is called in `ServerEnvironment::activateBlock`.

TurkeyMcMac:
@sfan5 Here is a procedure to test my claim:

1. Apply this patch:
   ```patch
   diff --git a/mesecons/init.lua b/mesecons/init.lua
   index 83e611b..7377979 100644
   --- a/mesecons/init.lua
   +++ b/mesecons/init.lua
   @@ -126,3 +126,13 @@ dofile(minetest.get_modpath("mesecons").."/legacy.lua");
    
    --Services like turnoff receptor on dignode and so on
    dofile(minetest.get_modpath("mesecons").."/services.lua");
   +
   +minetest.register_lbm({
   +	label = "Test LBM",
   +	name = "mesecons:test_lbm",
   +	nodenames = {"default:diamondblock"},
   +	run_at_every_load = true,
   +	action = function(pos, node)
   +		minetest.chat_send_all(minetest.pos_to_string(pos))
   +	end,
   +})
   ```
2. Enable my mod [area_containers](https://content.minetest.net/packages/jwmhjwmh/area_containers/).
3. Place an area container in the world.
4. Enter the container by right-clicking it.
5. Place a diamond block in the container.
6. Exit the container by right-clicking the exit node.
7. Close the server and reopen it.
8. Place a power plant horizontally adjacent to the area container.
9. Enter the container.

If LBMs were run on block loading, you would see the message in the chat after step 8. However, you will see the message in the chat only after step 9, indicating that LBMs are run on block activation.

TurkeyMcMac:
I decided to tack on the second commit because the code is intertwined.

sfan5:
> Here is a procedure to test my claim:

I checked the engine code today and this is indeed true.
I suppose it makes sense from a performance standpoint to only do it for block activations but this is surprising given the name and [the docs](https://github.com/minetest/minetest/blob/21323ef1ff02dd714330a8c9d8235f047e1dacf8/doc/lua_api.txt#L7633-L7635) are also straight up wrong about it.

mesecons_gates/init.lua
```diff
@@ -23,16 +23,18 @@ local gate_get_input_rules_twoinputs = mesecon.horiz_rules_getter({
...
-		minetest.add_item(pos, gate.drop)
-	elseif state then
-		minetest.swap_node(pos, {name = gate.onstate, param2=node.param2})
-		mesecon.receptor_on(pos, gate_get_output_rules(node))
-	else
-		minetest.swap_node(pos, {name = gate.offstate, param2=node.param2})
-		mesecon.receptor_off(pos, gate_get_output_rules(node))
+	local new_nodename = state and gate.onstate or gate.offstate
+	minetest.swap_node(pos, {name = new_nodename, param2 = node.param2})
+	if new_nodename ~= node.name then
```
sfan5, line 28: pretty sure you want to call the overheat thing even if the gate doesn't turn on or off, or not?

TurkeyMcMac: I made this change to have gates behave like luacontrollers do since https://github.com/minetest-mods/mesecons/pull/593.

sfan5: :shrug:

sfan5 (approved):

PR #584, opened 2022-01-27
Make some fixes to movement and rotation

### LINKED ISSUE #517: Gates do not update when rotated

TurkeyMcMac:
Fixes:
- Gates, conductors, and buttonlike components (e.g. buttons) now update things correctly when rotating (fixes #517.)
- Conductors that are turned on do not become stuck in this state when pushed by pistons (they update correctly.)

I did various tests, which you can download as WorldEdit schematics:
- [ex-recv-horiz.we.gz](https://github.com/minetest-mods/mesecons/files/7954363/ex-recv-horiz.we.gz): The wires should not get into the wrong state as you rotate the lever node around its axis.
- [ex-recv-vert.we.gz](https://github.com/minetest-mods/mesecons/files/7954364/ex-recv-vert.we.gz): The wires should not get into the wrong state as you rotate the button node around its axis.
- [ex-rot-and.we.gz](https://github.com/minetest-mods/mesecons/files/7954365/ex-rot-and.we.gz): The wires and the AND gate should not get into the wrong state as you rotate the AND gate node around its axis.
- [ex-rot-tee.we.gz](https://github.com/minetest-mods/mesecons/files/7954366/ex-rot-tee.we.gz): The wires and the piston should not get into the wrong state as you rotate the T-junction wire node around its axis.
- [ex-piston-strobe.we.gz](https://github.com/minetest-mods/mesecons/files/7954358/ex-piston-strobe.we.gz): When you place a wire in front of the piston, it should start going on and off really fast. It might not be in the right state when you break the wire, but that's a separate issue.
- [ex-push-and.we.gz](https://github.com/minetest-mods/mesecons/files/7954359/ex-push-and.we.gz): When you activate the piston, the AND gate should start flashing on and off then overheat.
- [ex-push-cond-dc.we.gz](https://github.com/minetest-mods/mesecons/files/7954360/ex-push-cond-dc.we.gz): After you activate the piston, the wires should be in the correct state.
- [ex-push-cond-loop.we.gz](https://github.com/minetest-mods/mesecons/files/7954361/ex-push-cond-loop.we.gz): After you activate the piston, the wires should be in the correct state.
- [ex-push-nor.we.gz](https://github.com/minetest-mods/mesecons/files/7954362/ex-push-nor.we.gz): When you activate the first piston, the second piston should activate then deactivate really fast.

### DIFF
--- a/mesecons/services.lua
+++ b/mesecons/services.lua
@@ -2,6 +2,7 @@
 
 mesecon.on_placenode = function(pos, node)
 	mesecon.execute_autoconnect_hooks_now(pos, node)
+	node = minetest.get_node(pos)
 
 	-- Receptors: Send on signal when active
 	if mesecon.is_receptor_on(node.name) then
@@ -11,18 +12,19 @@ mesecon.on_placenode = function(pos, node)
 	-- Conductors: Send turnon signal when powered or replace by respective offstate conductor
 	-- if placed conductor is an onstate one
 	if mesecon.is_conductor(node.name) then
+		local conductor = mesecon.get_conductor(node.name)
+		if conductor.state ~= mesecon.state.off then
+			node.name = conductor.offstate or conductor.states[1]
+			minetest.swap_node(pos, node)
+		end
 		local sources = mesecon.is_powered(pos)
 		if sources then
-			-- also call receptor_on if itself is powered already, so that neighboring
-			-- conductors will be activated (when pushing an on-conductor with a piston)
+			mesecon.vm_begin()
 			for _, s in ipairs(sources) do
 				local rule = vector.subtract(s, pos)
 				mesecon.turnon(pos, rule)
 			end
-			--mesecon.receptor_on (pos, mesecon.conductor_get_rules(node))
-		elseif mesecon.is_conductor_on(node) then
-			node.name = mesecon.get_conductor_off(node)
-			minetest.swap_node(pos, node)
+			mesecon.vm_commit()
 		end
 	end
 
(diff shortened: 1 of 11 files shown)


mesecons/services.lua
```diff
@@ -11,18 +12,20 @@ mesecon.on_placenode = function(pos, node)
 	-- Conductors: Send turnon signal when powered or replace by respective offstate conductor
 	-- if placed conductor is an onstate one
 	if mesecon.is_conductor(node.name) then
+		local conductor = mesecon.get_conductor(node.name)
+		if conductor.state ~= mesecon.state.off then
+			-- Turn the conductor off.
+			node.name = conductor.offstate or conductor.states[1]
```
sfan5, line 18: Why not `node.name = mesecon.get_conductor_off(node)`?

mesecons/services.lua
```diff
@@ -2,6 +2,7 @@
 
 mesecon.on_placenode = function(pos, node)
 	mesecon.execute_autoconnect_hooks_now(pos, node)
+	node = minetest.get_node(pos)
```
sfan5, line 5: I assume `execute_autoconnect_hooks_now` may change the node, a comment should point out why the get_node here is needed

mesecons/services.lua
TurkeyMcMac: `mesecon.get_conductor_off` seems to work differently for conductors with more than two states. I think it only sets one of the state bits to zero.

sfan5 (approved):
LGTM

sfan5:
Was that a rebase or any changes?

TurkeyMcMac:
Just a rebase.

PR #585, opened 2022-01-28
Move default dependency to individual mods that need it

TurkeyMcMac:
It was not needed. It is now an optional dependency, so that mods which expect to gain default as a transitive dependency by depending on mesecons won't break. However, default has been added as a direct dependency where appropriate within the modpack.


sfan5:
I think "Move default dependency to individual mods that need it" would be more accurate.

sfan5 (approved):

PR #586, opened 2022-01-28
Basically remove mvps_process_stack

TurkeyMcMac:
The other calls to `mvps_process_stack` in mesecons_pistons seem to have been removed before this.

I also removed the body of `mvps_process_stack` itself, as calling it is always redundant. The function still exists, but it does nothing. It is kept in case any external mods are using that part of the API.

This may improve the performance of piston pushing a bit.


sfan5 (approved):
Looks good.

PR #588, opened 2022-02-04
Protect server against Luacontroller memory usage

TurkeyMcMac:
If an OOM error occurs within controller code, it is caught and the server does not crash. However, if the controller uses up almost all the memory then puts a lot of data in the persistent storage, the server can crash while serializing the persistent storage. This PR introduces a limit on volatile memory usage. This can only be checked after the luacontroller finishes running, so it can't prevent server lag. However, it can prevent crashes. In addition, after an OOM condition is caught, the garbage collector is run in order to prevent an OOM error a moment later. The default memory limit is 2MB.

I don't run any public servers, so people who do should decide whether this PR is actually useful.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -590,8 +590,12 @@ local function load_memory(meta)
 end
 
 
-local function save_memory(pos, meta, mem)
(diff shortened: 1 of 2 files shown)


TurkeyMcMac:
This probably isn't needed.

PR #590, opened 2022-02-10
Fix VM light update issue

TurkeyMcMac:
Sorry, I realized my PR https://github.com/minetest-mods/mesecons/pull/578 had another bug. It can cause things like this:
screenshot_20220210_161644

You can see that the light cuts off sharply. To replicate this yourself, set the powered mese block light source to 14, then make a wire where regular conductors follow a mese block. The regular conductors should extend to the next mapblock. When the wire is powered such that the mese block is powered first, you will see an affect like that in the picture.

To fix this, I made it so that any conductor that needs a light update will make all VMs in the transaction use light updates. This PR should not cause any performance degradation on the fast path. I have not noticed any.

When testing, I recommend applying the following patch to make sure light updates are only used when necessary.

```patch
diff --git a/mesecons/util.lua b/mesecons/util.lua
index 55a0d63..3c4027b 100644
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -343,6 +343,7 @@ end
 -- Finishes a VoxelManipulator-based transaction, freeing the VMs and map data
 -- and writing back any modified areas.
 function mesecon.vm_commit()
+	if vm_update_light then minetest.chat_send_all("update light") end
 	for hash, tbl in pairs(vm_cache) do
 		if tbl.dirty then
 			local vm = tbl.vm
```

### DIFF
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -326,6 +326,9 @@ end
 -- Nil if no VM-based transaction is in progress.
 local vm_cache = nil
 
+-- Whether the current transaction will need a light update afterward.
+local vm_update_light = false
+
 -- Starts a VoxelManipulator-based transaction.
 --
 -- During a VM transaction, calls to vm_get_node and vm_swap_node operate on a
@@ -334,6 +337,7 @@ local vm_cache = nil
 -- vm_abort.
 function mesecon.vm_begin()
 	vm_cache = {}
+	vm_update_light = false
 end
 
 -- Finishes a VoxelManipulator-based transaction, freeing the VMs and map data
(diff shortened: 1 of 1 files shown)


mesecons/util.lua
```diff
@@ -391,8 +395,8 @@ end
 --
 -- The swap will necessitate a light update unless update_light equals false.
 function mesecon.vm_swap_node(pos, name, update_light)
+	vm_update_light = vm_update_light or update_light ~= false
```
sfan5, line 398: This could use a comment explaining pretty much the same as the PR body

sfan5 (approved):
looks good

PR #591, opened 2022-02-12
Fix luacheck warning

TurkeyMcMac:
Fixes a Luacheck warning.


TurkeyMcMac:
@SmallJoker Could you merge this real quick? It's trivial but important.

PR #592, opened 2022-02-12
Fix MTG doors not reacting correctly to Mesecon power

Niklp09:
title says all

### DIFF
--- a/mesecons_doors/init.lua
+++ b/mesecons_doors/init.lua
@@ -22,11 +22,11 @@ end
 local function meseconify_door(name)
 	if minetest.registered_items[name .. "_b_1"] then
 		-- old style double-node doors
-		local function toggle_state1 (pos)
+		local function toggle_state1 (pos, node)
 			on_rightclick(pos, 1, name.."_t_1", name.."_b_2", name.."_t_2", {1,2,3,0})
 		end
 
-		local function toggle_state2 (pos)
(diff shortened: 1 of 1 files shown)


mesecons_doors/init.lua
```diff
@@ -65,7 +65,9 @@ local function meseconify_door(name)
 			}}
 		}
 		minetest.override_item(name .. "_a", override)
+		minetest.override_item(name .. "_c", override)
 		minetest.override_item(name .. "_b", override)
+		minetest.override_item(name .. "_d", override)
```
sfan5, line 70: these need to be inside an `if minetest.registered_items[name .. "_c"] then` to not cause errors on older versions

mesecons_doors/init.lua
```diff
@@ -22,11 +22,11 @@ end
 local function meseconify_door(name)
 	if minetest.registered_items[name .. "_b_1"] then
 		-- old style double-node doors
-		local function toggle_state1 (pos)
+		local function toggle_state1 (pos, node)
 			on_rightclick(pos, 1, name.."_t_1", name.."_b_2", name.."_t_2", {1,2,3,0})
 		end
 
-		local function toggle_state2 (pos)
+		local function toggle_state2 (pos, node)
```
sfan5, line 29: Adding these back is not necessary

sfan5 (changes requested):

PR #593, opened 2022-02-16
Prevent extra pin events with luacontrollers, microcontrollers, and FPGAs

TurkeyMcMac:
Fixes https://github.com/minetest-mods/mesecons/issues/569.

These devices now only run their code when `action_change` actually changes the state of a port.

I have tested this with circuits from Mesecons Lab.

@Makefile-dot-in, does this fix your issue?

### DIFF
--- a/mesecons_fpga/init.lua
+++ b/mesecons_fpga/init.lua
@@ -114,8 +114,9 @@ plg.register_nodes({
 		effector = {
 			rules = {}, -- replaced later
 			action_change = function(pos, _, rule, newstate)
-				plg.ports_changed(pos, rule, newstate)
-				plg.update(pos)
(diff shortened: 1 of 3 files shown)


sfan5 (approved):
Tested a bunch, looks good.

Desour (commented):
:+1: Code looks good (you could document the new return value though). Also did some quick test, works fine.

Makefile-dot-in:
it seems to fix the issue for me too, although I would've preferred a solution that would prevent `action_change` from being fired in the first place if there is no change. That would probably require larger modifications to track mesecon states though, so I understand why this solution was used. (and overall the effect is rather minor in other effectors)

PR #595, opened 2022-02-21
Use metatable to cache light update conductors

TurkeyMcMac:
One problem with https://github.com/minetest-mods/mesecons/pull/578 was that the light update conductor cache was checked for every call to `turnon`/`turnoff`. While I'm sure this check had a negligible impact on performance, this PR removes said check using some metatable magic. The conductors are calculated the first time the cache is indexed, after which a nil metatable is set so that no more checks are needed.

### DIFF
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -368,17 +368,9 @@ function mesecon.is_power_off(pos, rulename)
 	return false
 end
 
--- The set of conductor states which require light updates when they change.
-local light_update_conductors
-
--- Calculate the contents of the above set if they have not been calculated.
-local function find_light_update_conductors()
-	-- The expensive calculation is only done the first time.
-	if light_update_conductors then return end
-
-	light_update_conductors = {}
-
-	-- Find conductors whose lighting characteristics change depending on their state.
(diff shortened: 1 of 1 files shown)


sfan5:
What? Checking a single value per call could not be more irrelevant for performance and also the "is there a metatable" check Lua does internally is not much different.

TurkeyMcMac:
@sfan5 I was going by this comment of yours: https://github.com/minetest-mods/mesecons/pull/578#discussion_r752685130

sfan5:
Ah, heh. The overhead is real, just very small and if it can be easily removed that's good but if not that's also okay.

TurkeyMcMac:
This way removes the repeated check by setting the metatable to `nil`. It also can save a tiny bit of memory by allowing `calculate_light_update_conductors` to be garbage collected. However, the new code may be harder to follow.

SmallJoker:
Whereas the implementation looks very clever, I do not think that it's a good practice. The update cycle might happen somewhere in between other calculations, hence it is no longer clear *when exactly* this function is called. The current code may "suffer" from one function call overhead, I don't think it's worth to sacrifice the predetermined code paths.

TurkeyMcMac:
Fair enough.

PR #599, opened 2022-03-01
Use FIFO queue for mvps

TurkeyMcMac:
The queue implementation is shared with `turnon` and `turnoff`.

The new implementation should be more efficient as it does not do a linear search through the list of nodes for every new node added in order to find duplicates.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -65,7 +65,7 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
 	local nodes = {}
 	local pos_set = {}
 	local frontiers = mesecon.fifo_queue.new()
-	frontiers:add(pos)
+	frontiers:add(vector.new(pos))
 
 	for np in frontiers:iter() do
 		local np_hash = minetest.hash_node_position(np)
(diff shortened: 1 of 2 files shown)


sfan5:
How do I test this? Pistons use mvps, right?

TurkeyMcMac:
Yes. I recommend testing with `mesecons_stickyblocks`, having pistons push various structures made of sticky blocks. Sorry I didn't specify in the PR description.

mesecons_mvps/init.lua
```diff
@@ -63,28 +63,31 @@ end
...
-	local frontiers = {pos}
-
-	while #frontiers > 0 do
-		local np = frontiers[1]
-		local nn = minetest.get_node(np)
-
-		if not node_replaceable(nn.name) then
+	local pos_set = {}
+	local frontiers = mesecon.fifo_queue.new()
+	frontiers:add(pos)
```
Desour, line 68:
```suggestion
	frontiers:add(vector.new(pos))
```
The returned `nodes` should not have a reference to an argument, in case `pos` gets altered after the call there might be weird errors.

Desour (commented):
:+1: Code looks equivalent and more efficient.
Haven't tested.

PR #600, opened 2022-03-06
Avoid duplicating VM data in Lua memory

TurkeyMcMac:
This is a change making the VM cache use `get_node_at` and `set_node_at` rather than `get_data` (et al.) and `set_data`. The former two methods were added some time in 2014. Hopefully earlier Minetest versions don't need to be supported.

Including a cache of individual nodes, the new code seems quite significantly faster. You can test using the benchmarks from https://github.com/minetest-mods/mesecons/pull/578. As before, the results will vary depending on one's computer and how the circuits align with mapblocks. Here are my numbers:

* **Line of diodes**: Old: 0.323s. New: 0.104s.
* **Block of wires**: Old: 1.752s. New: 1.396s.
* **Memory bank**: Old: 4.177s. New: 1.525s.

### DIFF
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -317,14 +317,13 @@ end
 --
 -- Contents of the table are:
 -- “vm” → the VoxelManipulator
--- “va” → the VoxelArea
--- “data” → the data array
--- “param1” → the param1 array
--- “param2” → the param2 array
 -- “dirty” → true if data has been modified
 --
 -- Nil if no VM-based transaction is in progress.
 local vm_cache = nil
+
+-- Cache from node position hashes to nodes (represented as tables).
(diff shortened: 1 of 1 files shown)


sfan5 (approved):

TurkeyMcMac:
By the way, it does seem like the new code can be slower sometimes. It may be slower when the entire turnon/turnoff occurs in one mapblock and changes many conductors of certain types (not the default wires, it seems.) For example, if you put a 14x14x14 block of `default:mese` in the middle of a mapblock and switch it on and off, the old code may execute faster.

The circuits I've tested from Mesecons Lab are faster with the new code. I'm pretty sure the cases where the new code is slower are rare enough that this is still a good performance improvement.

PR #602, opened 2022-03-09
Consolidate horizontal rules getters

TurkeyMcMac:
There are several different components that can be rotated around the Y axis. This PR consolidates all their rules getter functions using a new API function. This has the added benefit of making them all use lookup tables instead of repeated rotation, which is probably faster and creates less garbage. I think I have noticed a minor performance increase.

I have tested that the affected nodes still work, but it would be good if someone could verify this.

### DIFF
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -72,7 +72,28 @@ function mesecon.rotate_rules_up(rules)
 	end
 	return nr
 end
---
+
+-- Returns a rules getter function that returns different rules depending on the node's horizontal rotation.
+-- If param2 % 4 == 0, then the rules returned by the getter are a copy of base_rules.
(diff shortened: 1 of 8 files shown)


sfan5:
> I have tested that the affected nodes still work, but it would be good if someone could verify this.

I didn't get to doing this myself but what about exhaustively dumping all matching rules and then comparing it before and after this PR.

TurkeyMcMac:
You can test with [this patch](https://github.com/minetest-mods/mesecons/files/8541112/test-patch.txt). Most of the tests will print "OK" to the error log. mesecons_insulated will dump some rules to the error log that do not match each other. You can visually verify that these rules are actually equivalent.

sfan5 (approved):
Verified to work

PR #604, opened 2022-03-30
Fix error prone negation reported by luacheck

Desour:
Trivial.

If you installed luacheck using luarocks, you might have to upgrade it. Using `install` again worked for me.


sfan5:
Can you rename the commit to something (short) that hint at the actual problem?

Desour:
This is what luacheck calls it.

PR #605, opened 2022-04-02
Add automated tests for some mods

TurkeyMcMac:
The tests use [mineunit](https://github.com/S-S-X/mineunit), a nice Minetest testing framework based on Busted.

mesecons, mesecons_mvps, mesecons_fpga, and mesecons_luacontroller are tested. I don't want to add tests for every mod in the modpack, because that would be even more work.

A testing job now runs automatically just like luacheck.


S-S-X:
If you like you could drop PR for VM here: https://github.com/S-S-X/mineunit/blob/master/voxelmanip.lua replacing empty shell implementation there. Reason it is like that is simply because that happened to cover all functionality needed for thing I was working with... 🤷

TurkeyMcMac:
I can make a PR, but the functionality is quite limited. Not even `get_data` is supported.

.test_fixtures/mesecons.lua
```diff
@@ -0,0 +1,147 @@
...
+		offstate = "mesecons:test_conductor_rot_off",
+	}}
+	mesecon.register_node("mesecons:test_conductor_rot", {
+		description = "Rotatable Test Conductor",
+		on_rotate = mesecon.on_rotate_horiz,
+	}, {mesecons = off_spec}, {mesecons = on_spec})
+end
+
+do
+	local mesecons_spec = {conductor = {
```
SmallJoker, line 101: A comment would be helpful, such as
```
-- Utility node: this conductor is used to test the connectivity and state of adjacent wires
```

.test_fixtures/mesecons_fpga.lua
```diff
@@ -0,0 +1,58 @@
...
+	for i, instr in ipairs(program) do
+		local op1, act, op2, dst
+		if #instr == 3 then
+			act, op2, dst = unpack(instr)
+		else
+			assert.equal(4, #instr)
+			op1, act, op2, dst = unpack(instr)
+		end
+		fields[i .. "op1"] = op1
+		fields[i .. "act"] = (" "):rep(4 - #act) .. act
```
SmallJoker, line 33: Needs commenting. what do this do? what's expected?

mesecons/spec/action_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+		mesecon.swap_node_force(layout[1][1], "mesecons:test_receptor_off")
+		mesecon.receptor_off(layout[1][1], mesecon.rules.alldirs)
+		mineunit:execute_globalstep()
+		mineunit:execute_globalstep()
+		assert.equal(4, #mesecon._test_effector_events)
+		assert.same({"off", layout[4][1]}, mesecon._test_effector_events[3])
+		assert.same({"off", layout[5][1]}, mesecon._test_effector_events[4])
+	end)
+
+	it("ignores overwritten actions", function()
```
SmallJoker, line 41:
```suggestion
	it("discards outdated/overwritten node events", function()
```

mesecons/spec/action_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+		mesecon.swap_node_force(layout[1][1], "mesecons:test_receptor_off")
+		mesecon.receptor_off(layout[1][1], mesecon.rules.alldirs)
+		mineunit:execute_globalstep()
+		mineunit:execute_globalstep()
+		assert.equal(0, #mesecon._test_effector_events)
+	end)
+
+	it("delays actions", function()
+		mesecon.swap_node_force(layout[1][1], "mesecons:test_receptor_on")
+		mesecon.queue:add_action(layout[1][1], "receptor_on", {mesecon.rules.alldirs}, 1, nil)
```
SmallJoker, line 53: Mods use `minetest.set_node`, hence it would make sense to use an equivalent here instead of manual speed optimized code.

mesecons/spec/service_spec.lua
```diff
@@ -0,0 +1,186 @@
...
+		world.layout(layout)
+	end)
+
+	after_each(function()
+		mesecon._test_reset()
+		world.clear()
+	end)
+
+	it("updates components when a receptor changes", function()
+		mesecon._test_dig(layout[1][1])
```
SmallJoker, line 26:
```suggestion
		-- Remove the source (receptor) and check whether the adjacent conductor(s) turn off
		mesecon._test_dig(layout[1][1])
```

It would be good to have a brief comment for some of the more complicated setups. Otherwise it's a bit tricky to figure out what's going on.

mesecons_fpga/spec/logic_spec.lua
```diff
@@ -0,0 +1,235 @@
...
+			{" ", " ", " ", " "},
+			{"=", "A", "B"},
+			{" ", " ", " ", " "},
+			{" ", " ", " ", " "},
+			{"=", "A", "C"},
+		})
+	end)
+
+	it("transmits changes", function()
+		mesecon._test_program_fpga(pos, {
```
SmallJoker, line 165: Same here. Something like `-- Reprogram the FPGA and check whether nodes are updated` would already help

mesecons_fpga/spec/logic_spec.lua
```diff
@@ -0,0 +1,235 @@
...
+				return fmt:format(a[1], a[2].x, a[2].y, a[2].z) < fmt:format(b[1], b[2].x, b[2].y, b[2].z)
+			end)
+			return o
+		end
+
+		mesecon._test_place(pos_a, "mesecons:test_receptor_on")
+		mineunit:execute_globalstep()
+		mineunit:execute_globalstep()
+		mineunit:execute_globalstep()
+		mineunit:execute_globalstep()
```
SmallJoker, line 191: Why multiple cycles? This looks very fragile and unreliable.

mesecons_mvps/spec/node_spec.lua
```diff
@@ -0,0 +1,276 @@
...
+
+world.set_default_node("air")
+
+describe("node movement", function()
+	after_each(function()
+		mesecon._test_reset()
+		world.clear()
+	end)
+
+	it("works with no nodes", function()
```
SmallJoker, line 13: There's always a node, often enough `air` or `ignore`. Unclear description.

SmallJoker (commented):
Generally a welcome addition, however comments are somewhat lacking. To ease maintenance and additions, it would really help to have it commented better.

mesecons_fpga/spec/logic_spec.lua
S-S-X: I do not know mesecons internals that well but if it is because of queue timing then specifying dtime with `mineunit:execute_globalstep(dtime)` might be better? When not explicitly specified default dtime for mineunit:execute_globalstep() is 0.1

Mineunit does not have good way to control `core.get_use_time()` but can offer predictable output if you add `time_step = 1000` in mineunit.conf for 1000us step between each call or ofc override whole thing. Default is to use (very bad) wall clock implementation for core.get_us_time.

TurkeyMcMac: What matters is not the dtime but the number of globalsteps. This is why the function is called repeatedly. I think doing so is fine, since the number of globalsteps required should not change across Mesecons versions. The regression tests will continue to work.

S-S-X: Well makes sense then, maybe having additional assertion (and comment or something) after 3 globalsteps to make it clear?
Have to agree that it looks very hacky without knowing internals inside out and prob would be good to have inline explanation because of that.

mesecons_fpga/spec/logic_spec.lua
TurkeyMcMac: I have changed the test description.

mesecons/spec/service_spec.lua
TurkeyMcMac: Hopefully the confusing tests are well described now.

.test_fixtures/mesecons.lua
```diff
@@ -0,0 +1,147 @@
...
+			minetest.remove_node(pos)
+			return
+		end
+		local bit = tonumber(rule_name.name, 2)
+		local bits_above = node.param2 - node.param2 % (bit * 2)
+		local bits_below = node.param2 % bit
+		local bits_flipped = new_state == mesecon.state.on and bit or 0
+		node.param2 = bits_above + bits_flipped + bits_below
+		minetest.swap_node(pos, node)
+	end
```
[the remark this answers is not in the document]
SmallJoker: Good enough. Thank you,

SmallJoker (approved):
It's a lot of changes - luckily "only" unittests. I did browse through the code again and it looks good. The unittests pass, so :+1:. Will merge in a few days unless there's objections.

sfan5 (approved):
I would prefer real-engine testing but it's not like anyone's gonna work on it so this is fine too.

PR #606, opened 2022-04-11
Allow specifying a port for digilines messages to be sent from the lua_controller

oversword:


Desour:
This has been requested once, but was denied due to digilines being intended to be position independent:
https://github.com/minetest-mods/digilines/issues/37#issuecomment-270847381

eshattow:
It would be really nice to be able to connect digiline to the node below the luacontroller / ioexpander. Burning a mesecons i/o port to connect with digiline is too limiting. The way around that limit is mesecons+digiline conducting pipeworks tube (but that is not possible in many situations). Another take on this would be to have mesecons ports connect on the same Y-height as the controller but have digiline connect at Y-1. Already there's some trouble with digiline filter injector not accepting digiline conducting pipeworks tube from Y+1. Saying that digiline is intended to be position independent is not quite what happens right now, it is but only if everything is at the same Y-height.

PR #607, opened 2022-04-20
Make modpack mostly independent from MTG

TurkeyMcMac:
This PR makes all Mesecons mods except `mesecons_doors` work with any game, and none of the mods hard-depend on `default` any longer. I added a mod `mesecons_gamecompat` to handle most external dependencies. The addition of a mod to the pack may cause problems for people updating it, in which case I should put the compatibility code in the `mesecons` mod or something. In any case, almost the entire modpack will now work with any game, although it requires `default` and some other MTG mods to provide sounds etc.

`mesecons_detectors` depended on `default` to provide `default_steel_block.png`. To remedy this, I added a copy of the texture to the mod as a fallback (giving appropriate attribution.) I could make the mod use the fallback texture 100% of the time, but this might mess with existing texture packs.

If no `mesecons_gamecompat:mese` alias is registered, `mesecons_extrawires` provides its own placeholder `mesecons_extrawires:mese`. Otherwise, `mesecons_extrawires:mese` is an alias of the real mese.

Support for more games without the `default` mod can be added in the future by adding more compatibility code to `mesecons_gamecompat`.


sfan5:
Nice.
To me `mesecons_compat` sounds more like internal backwards compatibility maybe `mesecons_gamecompat` would be a better name or you have a different idea.
I'll add some more comments in a review below.
Also what about [the list](https://github.com/minetest-mods/mesecons/blob/master/mesecons_mvps/init.lua#L314) inside mesecons_mvps?

mesecons_blinkyplant/init.lua
```diff
@@ -26,7 +26,7 @@ mesecon.register_node("mesecons_blinkyplant:blinky_plant", {
 	paramtype = "light",
 	is_ground_content = false,
 	walkable = false,
-	sounds = default.node_sound_leaves_defaults(),
+	sounds = mesecon.node_sound_leaves_defaults,
```
sfan5, line 29: Since you have the freedom to rename then anyway what about `mesecon.node_sounds.leaves`?

mesecons_compat/init.lua
```diff
@@ -0,0 +1,42 @@
...
+	end
+end
+
+if minetest.get_modpath("fire") then
+	mesecon.sound_name_fire = "fire_fire"
+end
+
+if minetest.get_modpath("tnt") then
+	mesecon.sound_name_explode = "tnt_explode"
+end
```
sfan5, line 42: These should be inside the first if. When 'default' isn't found then there's no telling that the dye, fire or tnt mods actually do what is expected (that being "exactly what MTG does").

TurkeyMcMac:
I'm not sure about changing the list in mesecons_mvps, since mods might want to register the MVPS stoppers for themselves.

sfan5:
> I'm not sure about changing the list in mesecons_mvps, since mods might want to register the MVPS stoppers for themselves.

I don't understand the concern here. There is a hardcoded list of nodes in the mod, surely it should only apply when MTG is detected?

TurkeyMcMac:
Yes, my thought is that the list should apply only when MTG is used, and other games that Mesecons is made compatible with can register their stoppers themselves.

sfan5:
Then what about moving the hardcoded list to `mesecons_gamecompat`, inside the MTG part?

mesecons_gamecompat/init.lua
```diff
@@ -0,0 +1,14 @@
...
+
+mesecon.texture = {}
+
+mesecon.dye_colors = {
+	"red", "green", "blue", "grey", "dark_grey", "yellow",
+	"orange", "white", "pink", "magenta", "cyan", "violet",
+}
+
+if minetest.get_modpath("default") then
+	dofile(minetest.get_modpath("mesecons_gamecompat").."/compat_mtg.lua")
```
sfan5, line 13:
```suggestion
	minetest.log("info", "Mesecons: detected Minetest Game for game compatibility")
	dofile(minetest.get_modpath("mesecons_gamecompat").."/compat_mtg.lua")
```

sfan5 (approved):
Works, tested with devtest.

Tim79000:
What's wrong with default?

TurkeyMcMac:
@Tim79000 default contains a lot of unrelated features, many of them specific to Minetest Game and its spin-offs. If a mod depends on default, it basically depends on Minetest Game.

PR #615, opened 2022-05-21
Fix conductor aliases again

TurkeyMcMac:
I think they were broken by https://github.com/minetest-mods/mesecons/pull/600, since the name of a node from the cache may now be an alias. So as not to risk slowing down the oft-called `mesecon.vm_swap_node`, I decided to fix this by resolving any aliases beforehand. This has the added benefit of fixing conductors with more than two states which have aliases in their state lists; before they would crash. The downside is that mesecons specifications may be modified, which is not documented. On the other hand, it isn't documented that they _won't_ be modified, either. Sharing state lists between multiple conductors still works.

The tests from https://github.com/minetest-mods/mesecons/pull/605 pass.

### DIFF
--- a/mesecons/util.lua
+++ b/mesecons/util.lua
@@ -403,6 +403,7 @@ function mesecon.vm_get_node(pos)
 end
 
 -- Sets a node’s name during a VoxelManipulator-based transaction.
+-- Aliases may not work correctly.
 --
 -- Existing param1, param2, and metadata are left alone.
 --
@@ -453,6 +454,7 @@ end
 -- the server’s main map data cache and then accessed from there.
 --
 -- Inside a VM transaction, the transaction’s VM cache is used.
+-- Aliases may not work correctly.
 --
 -- This function can only be used to change the node’s name, not its parameters
 -- or metadata.
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -372,17 +372,33 @@ end
 -- The set of conductor states which require light updates when they change.
 local light_update_conductors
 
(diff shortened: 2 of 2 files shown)


mesecons/util.lua
```diff
@@ -403,6 +403,7 @@ function mesecon.vm_get_node(pos)
 end
 
 -- Sets a node’s name during a VoxelManipulator-based transaction.
+-- Aliases may not work correctly.
```
SmallJoker, line 406:
```suggestion
-- Aliases may not work correctly if mesecon.get_conductor(name).states was touched by a mod.
```

That's how I understood the problem, right? If you want to make sure they're not modified, consider metatables. At least the reason *why* it may not work correct should be stated here.

TurkeyMcMac:
Actually, I think I can fix this problem in conjunction with https://github.com/minetest-mods/mesecons/pull/614. `vm_get_node` can just do the resolving of the alias. `vm_get_node_nocopy` is a private function, so it's not an issue.

mesecons/util.lua
TurkeyMcMac: The issue can occur if a mod swaps in a node whose name is actually an alias.

SmallJoker: Then please be so nice to mention in which cases this function is expected to work, and when not.

TurkeyMcMac: I will make some changes after #614 is merged or closed. If it is merged, I can make some further changes to the code then remove the notice. This PR is basically blocked for now.

PR #621, opened 2022-08-22
add minimum interrupt time to luacontroller to optionally remove hyper-fast signals

chvostek:
I've been seeing luacontrollers being used in place of blinkies, because their signal frequency can made much faster. This change provides a reasonable default minimum interrupt time, but also provides a knob so people wishing to keep short times have the option to do so.

### DIFF
--- a/settingtypes.txt
+++ b/settingtypes.txt
@@ -23,6 +23,7 @@ mesecon.luacontroller_string_rep_max (string:rep result length limit) int 64000 
 mesecon.luacontroller_digiline_maxlen (Digiline message size limit) int 50000 1000 1000000
 mesecon.luacontroller_maxevents (Controller execution time limit) int 10000 1000 100000
 mesecon.luacontroller_memsize (Controller memory limit) int 100000 10000 1000000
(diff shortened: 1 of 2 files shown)


settingtypes.txt
```diff
@@ -23,6 +23,7 @@ mesecon.luacontroller_string_rep_max (string:rep result length limit) int 64000
 mesecon.luacontroller_digiline_maxlen (Digiline message size limit) int 50000 1000 1000000
 mesecon.luacontroller_maxevents (Controller execution time limit) int 10000 1000 100000
 mesecon.luacontroller_memsize (Controller memory limit) int 100000 10000 1000000
+mesecon.luacontroller_minimum_interrupt (Minimum interrupt time) fload 1 0.1 1000000
```
blaboing: pretty sure that should be `float` not `fload` 😉

chvostek: I was just in the process of deleting and resubmitting... :/

blaboing: ah you already found it 😄

Desour:
:-1: Useless. This doesn't prevent high frequency circuits at all.

PR #622, opened 2022-08-22
add minimum interrupt time to luacontroller to optionally remove hyper-fast signals

chvostek:
I've been seeing luacontrollers being used in place of blinkies, because their signal frequency can made much faster. This change provides a reasonable default minimum interrupt time, but also provides a knob so people wishing to keep short times have the option to do so.

Per calcul0n's suggestion, this change provides a warning and sets too-small timer values to whatever minimum is configured, rather than simply failing with an error.


Desour:
:-1: https://github.com/minetest-mods/mesecons/pull/621#issuecomment-1222357532

numberZero:
Dupe of #621.

PR #623, opened 2022-08-25
Handle invalid param2 in presets.lua

TurkeyMcMac:
Fixes https://github.com/minetest-mods/mesecons/issues/610.

To test, load up devtest with Mesecons and homedecor_lighting. Place a Mesecons lever and, separately, a `homedecor:glowlight_small_cube_14`. Use the param2 tool to set the lever's param2 to 25 and the light's param2 to 6. Try turning on the lever. Then try turning on the light using a switch adjacent to it. The game should not crash.

### DIFF
--- a/mesecons/presets.lua
+++ b/mesecons/presets.lua
@@ -64,6 +64,8 @@ local rules_buttonlike = {
 }
 
 local function rules_from_dir(ruleset, dir)
+	if not dir then return {} end
+
 	if dir.x ==  1 then return ruleset.xp end
 	if dir.y ==  1 then return ruleset.yp end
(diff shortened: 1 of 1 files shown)


sfan5:
Can't we place the check closer to the cause of the issue?

TurkeyMcMac:
We could validate the param2 values, but that would be more complicated. I see the current approach as similar to checking for NaN after doing a numeric calculation.

Really the root cause is some other mod changing param2 values willy-nilly, but Mesecons has no control of that.

SmallJoker (approved):
I would also expect an empty table of rules if there is none for the current mesecon direction.
Looks good :+1:

PR #625, opened 2022-11-11
Add Esperanto translation

jolesh:
Only missing the gates as they are registered in more complex function, but doesn't really matter in my opinion as they often are called AND Gates, etc. in other languages as well.


Niklp09:
#503

SmallJoker (approved):
LGTM. The other PR needs a lot of rebasing anyway, so this here is a good step into the right direction. Will merge in a few days unless there's objections.

PR #627, opened 2022-11-20
Fix mvps receptor_off issue

TurkeyMcMac:
This is a less disruptive alternative to https://github.com/minetest-mods/mesecons/pull/594. See that PR for how to reproduce the issue. This one just checks that effectors should actually be powered off when they are powered directly by a receptor. When a receptor turns off conductors instead, this check is effectively done already.

A call to `changesignal` was removed in `mesecons/init.lua`. I think this was superfluous, as it was first added in 841bc70b97e237477b0dd9ba36fd9cd307daf581 to fix a bug that does not exist in the current code.

### DIFF
--- a/mesecons/init.lua
+++ b/mesecons/init.lua
@@ -99,7 +99,6 @@ mesecon.queue:add_function("receptor_off", function (pos, rules)
 		local rulenames = mesecon.rules_link_rule_all(pos, rule)
 		for _, rulename in ipairs(rulenames) do
 			mesecon.vm_begin()
-			mesecon.changesignal(np, minetest.get_node(np), rulename, mesecon.state.off, 2)
 
 			-- Turnoff returns true if turnoff process was successful, no onstate receptor
 			-- was found along the way. Commit changes that were made in voxelmanip. If turnoff
--- a/mesecons/internal.lua
+++ b/mesecons/internal.lua
@@ -542,6 +542,7 @@ function mesecon.turnoff(pos, link)
 	end
 
 	for _, sig in ipairs(signals) do
+		-- If sig.depth is 1, it has not yet been checked that the power source is actually off.
 		if sig.depth > 1 or not mesecon.is_powered(sig.pos, sig.link) then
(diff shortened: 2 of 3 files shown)


SmallJoker:
@Desour Would you please be so nice to have a look at this PR to check whether it works as expected?

Desour:
TL;DR: :+1: 

Details:
* No timing issues as in the other PR (obviously).
* If I understand the change correctly, it makes sure that `.turnoff` is effect-less if the node at the pos is not actually off (or to-be-turned-off in case of conductors). And for conductors this is already done with the `is_receptor_on` above.
  This makes sense to do, as we depend on this behaviour in mvps (see below).
* The bug is fixed.
* The removed `.changesignal` also looks superfluous to me (and it would also require the power-on check if it was still there).

---------

Below:
We're only calling `.turnoff` in `receptor_off`. Here's a diff for printing cases when the receptor (or conductor) is still on:
```diff
diff --git a/mesecons/init.lua b/mesecons/init.lua
index c11e480..f158dac 100644
--- a/mesecons/init.lua
+++ b/mesecons/init.lua
@@ -93,6 +93,16 @@ end
 mesecon.queue:add_function("receptor_off", function (pos, rules)
        rules = rules or mesecon.rules.default
 
+       do
+               local node = mesecon.get_node_force(pos)
+               if mesecon.is_receptor_on(node.name) then
+                       minetest.log("recepter_off: receptor still on at "..vector.to_string(pos))
+               end
+               if mesecon.is_conductor_on(node, nil) then
+                       minetest.log("recepter_off: conductor still on at "..vector.to_string(pos))
+               end
+       end
+
        -- Call turnoff on all linking positions
        for _, rule in ipairs(mesecon.flattenrules(rules)) do
                local np = vector.add(pos, rule)
```
It is triggered by the construction in https://github.com/minetest-mods/mesecons/pull/594 (the thing that is fixed with the PR).
(Checking in `receptor_off` already if the thing is off could probably work for receptors. But for conductors, we still need this change (`turnoff` should always check for conductors if it's powered from anywhere; and also, conductors can have multiple states at once).)

PR #629, opened 2022-11-24
Straighten wire

fluxionary:
does just that


SmallJoker (approved):
Seems about right. Will merge in a few days unless there's objections.

PR #631, opened 2022-12-07
Compile FPGA configurations to Lua

TurkeyMcMac:
FPGA configurations are compiled and cached. Cache entries are removed if they are not used for at least a minute.

## Benchmark

Use this WorldEdit schematic: [fpgabenchmark.we.gz](https://github.com/minetest-mods/mesecons/files/10177181/fpgabenchmark.we.gz)

Apply this patch to Mesecons:

```patch
diff --git a/mesecons/actionqueue.lua b/mesecons/actionqueue.lua
index 72b464d..d84b6b2 100644
--- a/mesecons/actionqueue.lua
+++ b/mesecons/actionqueue.lua
@@ -64,6 +64,13 @@ function queue:add_action(pos, func, params, time, overwritecheck, priority)
 	table.insert(self.actions, action)
 end
 
+local sum = 0
+minetest.register_chatcommand("r", {
+	description = "reset",
+	params = "",
+	func = function() sum = 0 end,
+})
+
 -- execute the stored functions on a globalstep
 -- if however, the pos of a function is not loaded (get_node_or_nil == nil), do NOT execute the function
 -- this makes sure that resuming mesecons circuits when restarting minetest works fine (hm, where do we do this?)
@@ -77,6 +84,9 @@ local function globalstep_func(dtime)
 	-- queue.actions: actions to execute later
 	local actions_now = {}
 	queue.actions = {}
+	if #actions == 0 then return end
+	collectgarbage()
+	local before = minetest.get_us_time()
 
 	for _, ac in ipairs(actions) do
 		if ac.time > 0 then
@@ -110,6 +120,10 @@ local function globalstep_func(dtime)
 	for _, ac in ipairs(actions_now) do
 		queue:execute(ac)
 	end
+
+	local after = minetest.get_us_time()
+	sum = sum + (after - before)
+	minetest.log("error", sum / 1000000)
 end
 
 -- delay the time the globalsteps start to be execute by 4 seconds

```

Run the benchmark like so:

1. Turn the switch on and off 6 times to warm up.
2. Enter the command `/r` to reset the sum.
3. Turn the switch on and off 20 times.
4. Record the printed sum.

With the old Mesecons, I got a time of roughly 0.45. With the new Mesecons, the time was roughly 0.40.

## Testing

Besides the automated tests, I recommend enabling the new version of `mesecons_fpga` in Mesecons Lab and trying out some contraptions.

### DIFF
--- a/mesecons_fpga/init.lua
+++ b/mesecons_fpga/init.lua
@@ -312,6 +312,37 @@ plg.red_box_around = function(i)
 end
 
 
+local compile_cache = {}
+
+plg.compile_with_cache = function(instr)
+	local entry = compile_cache[instr]
+	if entry then
+		-- Prevent removal from the cache
+		entry.age = 0
+	else
+		entry = {
+			func = lcore.compile(lcore.deserialize(instr)),
+			age = 0,
+		}
+		compile_cache[instr] = entry
+	end
+	return entry.func
+end
+
+-- Entries are removed from the cache if they are unused for at least a minute.
+local function clean_compile_cache()
+	for instr, entry in pairs(compile_cache) do
+		if entry.age < 1 then
+			entry.age = entry.age + 1
+		else
+			compile_cache[instr] = nil
+		end
+	end
+	minetest.after(60, clean_compile_cache)
+end
+minetest.after(60, clean_compile_cache)
+
+
 plg.update = function(pos)
 	local meta = minetest.get_meta(pos)
 	if meta:get_int("valid") ~= 1 then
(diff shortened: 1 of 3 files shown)


Desour:
One could also try to store fpga configurations in cnf ( $A_{out} = \bigwedge{\bigvee{...}}$ ), and compute them with bit operations.
I.e. per port a list of (4 bit) bitmask pairs `(use, negate)`, and the port will be set to 1 iff `(input xor negate) and use ~= 0` holds for all of these pairs.

TurkeyMcMac:
That sounds worth trying, as it could speed up interpretation without needing a cache. It would increase the metadata size of FPGAs.

numberZero:
Compilation? CNF? What’s about simply *precomputing* FPGA output for all 16 possible inputs?

PR #633, opened 2023-02-19
add option to disable printing inside a luacontroller

### LINKED ISSUE #626: output of print() goes to the server log.
this partially intersects w/ #388, but i don't think it's entirely a dupe.

on public servers, having the output of `print()` go to the server log is worse than useless. players cannot see it, and it can clog up the log w/ a lot of spurious, unfilterable garbage. 

at the very least, there should be an option to disable the output that servers can enable. if others agree on that, i can create a PR.

fluxionary:
fixes #626.

default behavior is now to dump and log the argument to print similar to how it used to work. by setting `mesecon.luacontroller_print_behavior = noop` in minetest.conf, this behavior can be disabled.

### DIFF
--- a/mesecons_luacontroller/init.lua
+++ b/mesecons_luacontroller/init.lua
@@ -203,11 +203,13 @@ end
 -------------------------
 
 local function safe_print(param)
-	local string_meta = getmetatable("")
-	local sandbox = string_meta.__index
-	string_meta.__index = string -- Leave string sandbox temporarily
-	print(dump(param))
-	string_meta.__index = sandbox -- Restore string sandbox
+	if mesecon.setting("luacontroller_print_behavior", "log") == "log" then
+		local string_meta = getmetatable("")
+		local sandbox = string_meta.__index
+		string_meta.__index = string -- Leave string sandbox temporarily
+		minetest.log("action", string.format("[mesecons_luacontroller] print(%s)", dump(param)))
+		string_meta.__index = sandbox -- Restore string sandbox
+	end
 end
 
 local function safe_date()
--- a/settingtypes.txt
+++ b/settingtypes.txt
@@ -28,7 +28,7 @@ mesecon.luacontroller_memsize (Controller memory limit) int 100000 10000 1000000
 # IID is ignored and at most one interrupt may be queued if this setting is enabled.
 mesecon.luacontroller_lightweight_interrupts (Lightweight interrupts) bool false
 
-# Behavior of "print" inside a luacontroller. By default, this emits a message into debug.txt.
+# Behavior of print() inside a luacontroller. By default, this emits a message into actionstream.
 # Set it to noop if you wish to disable that behavior.
 mesecon.luacontroller_print_behavior (Behavior of print) enum log log,noop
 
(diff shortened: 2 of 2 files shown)


settingtypes.txt
```diff
@@ -28,6 +28,10 @@ mesecon.luacontroller_memsize (Controller memory limit) int 100000 10000 1000000
 # IID is ignored and at most one interrupt may be queued if this setting is enabled.
 mesecon.luacontroller_lightweight_interrupts (Lightweight interrupts) bool false
 
+# behavior of "print" inside a luacontroller. by default, this emits a message into debug.txt.
+# set it to noop if you wish to disable that behavior.
+mesecon.luacontroller_print_behavior (behavior of print) enum log log,noop
```
numberZero, line 33: Use caps please, like in other options.

numberZero (commented):
LGTM. Does it work well with nested objects (that are printed on several lines)?

fluxionary:
> LGTM. Does it work well with nested objects (that are printed on several lines)?

it still uses `dump()`, so a single message is often broken up over multiple lines. an alternate could be to split the dump by line, and log each separately, though that's not really a meaningfully better option IMO.

settingtypes.txt
```diff
@@ -28,6 +28,10 @@ mesecon.luacontroller_memsize (Controller memory limit) int 100000 10000 1000000
 # IID is ignored and at most one interrupt may be queued if this setting is enabled.
 mesecon.luacontroller_lightweight_interrupts (Lightweight interrupts) bool false
 
+# Behavior of "print" inside a luacontroller. By default, this emits a message into debug.txt.
```
Desour, line 31: It also prints into terminal.
```suggestion
# Behavior of print() inside a luacontroller. By default, this emits a message into actionstream.
```

mesecons_luacontroller/init.lua
```diff
@@ -203,11 +203,13 @@ end
...
-	local string_meta = getmetatable("")
-	local sandbox = string_meta.__index
-	string_meta.__index = string -- Leave string sandbox temporarily
-	print(dump(param))
-	string_meta.__index = sandbox -- Restore string sandbox
+	if mesecon.setting("luacontroller_print_behavior", "log") == "log" then
+		local string_meta = getmetatable("")
+		local sandbox = string_meta.__index
+		string_meta.__index = string -- Leave string sandbox temporarily
+		minetest.log("action", string.format("[mesecons_luacontroller] print(%s)", dump(param)))
```
Desour, line 210: (I'd personally prefer `"[mesecons_luacontroller] %s"`, but this is also fine I guess.)

Desour (commented):
:+1: Indeed LGTM.

mesecons_luacontroller/init.lua
numberZero: Fine as is. It should be clear the message origins in a (player-controlled) LuaC and not in the mod itself.

fluxionary: i prefer how i've written it to Desour's proposal, but only marginally - in practice, it doesn't probably matter either way. as i've got one backup, and i don't expect further comments, i request permission to resolve this.

PR #636, opened 2023-03-18
Translate Mesecons to German

Wuzzy2:
This translate the Mesecons modpack into German.

### DIFF
--- a/mesecons_delayer/locale/mesecons_delayer.de.tr
+++ b/mesecons_delayer/locale/mesecons_delayer.de.tr
@@ -1,0 +1,3 @@
+# textdomain: mesecons_delayer
+Delayer=Verzögerer
+You hacker you=Sie Hacker!
--- a/mesecons_delayer/locale/mesecons_delayer.de.tr
+++ b/mesecons_delayer/locale/mesecons_delayer.de.tr
@@ -1,3 +1,3 @@
 # textdomain: mesecons_delayer
 Delayer=Verzögerer
-You hacker you=Sie Hacker!
+You hacker you=Sie Hacker
--- a/mesecons_detector/locale/mesecons_detector.de.tr
+++ b/mesecons_detector/locale/mesecons_detector.de.tr
@@ -1,0 +1,3 @@
+# textdomain: mesecons_detector
+Player Detector=Spielerdetektor
+Node Detector=Node-Detektor
--- a/mesecons_detector/locale/mesecons_detector.de.tr
+++ b/mesecons_detector/locale/mesecons_detector.de.tr
@@ -1,3 +1,3 @@
 # textdomain: mesecons_detector
 Player Detector=Spielerdetektor
-Node Detector=Node-Detektor
+Node Detector=Blockdetektor
--- a/mesecons_insulated/locale/mesecons_insulated.de.tr
+++ b/mesecons_insulated/locale/mesecons_insulated.de.tr
@@ -1,0 +1,2 @@
+# textdomain: mesecons_insulated
+Straight Insulated Mesecon=Isolierte Mesecongerade
--- a/mesecons_lightstone/locale/mesecons_lightstone.de.tr
+++ b/mesecons_lightstone/locale/mesecons_lightstone.de.tr
@@ -1,0 +1,13 @@
+# textdomain: mesecons_lightstone
+Red Lightstone=Roter Leuchtstein
+Green Lightstone=Grüner Leuchtstein
+Blue Lightstone=Blauer Leuchtstein
+Grey Lightstone=Grauer Leuchtstein
+Dark Grey Lightstone=Dunkelgrauer Leuchtstein
+Yellow Lightstone=Gelber Leuchtstein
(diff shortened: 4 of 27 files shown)


mesecons_detector/locale/mesecons_detector.de.tr
```diff
@@ -0,0 +1,3 @@
+# textdomain: mesecons_detector
+Player Detector=Spielerdetektor
+Node Detector=Node-Detektor
```
numberZero, line 3: Shouldn’t that be “Block”? https://dev.minetest.net/Translating/de

Wuzzy2:
Yes, right. I will fix it tomorrow (hopefully). I wait for more comments first.

Wuzzy2:
Done.

mesecons_lightstone/locale/mesecons_lightstone.de.tr
```diff
@@ -0,0 +1,13 @@
+# textdomain: mesecons_lightstone
+Red Lightstone=Roter Leuchtstein
+Green Lightstone=Grüner Leuchtstein
+Blue Lightstone=Blauer Leuchtstein
+Grey Lightstone=Grauer Leuchtstein
+Dark Grey Lightstone=Dunkelgrauer Leuchtstein
+Yellow Lightstone=Gelber Leuchtstein
+Orange Lightstone=Orange Leuchtstein
```
Desour, line 8:
```suggestion
Orange Lightstone=Oranger Leuchtstein
```
oder
```suggestion
Orange Lightstone=Orangener Leuchtstein
```
idk

mesecons_materials/locale/mesecons_materials.de.tr
```diff
@@ -0,0 +1,4 @@
+# textdomain: mesecons_materials
+Glue=Klebstoff
+Fiber=Kabel
```
Desour, line 3: This is the insulator item.
Maybe "Gewebe"?

Desour (commented):
TBH, I kinda hate in-game translations. x)
But that's not a reason to be against this.

Wuzzy2:
“orange” is correct as the word “orange” has no flection in German. The same translation is used in MTG.

I’m not sure about the translation of "Fiber". What about "Faser”?

When I think about this, maybe the English word for this item is just bad because “Fiber” is an awfully generic word that could mean anything.

fluxionary:
i'm only here incidentally, and just want to add some comments that i don't think need to be acted on. 

looking at the icon for the item called "fiber", it is a woven sheathing intended to provide insulation to a cable. the uses for the item validate this. the name "fiber" would be a [synecdoche](https://en.wikipedia.org/wiki/Synecdoche), referring to individual strands of material used to weave such a sheath. the resulting item is "a thing made of fiber". in english, it makes sense to an engineer or electrician when they see the icon, but not everyone has that sort of experience. on the other hand, not everyone can *see* the icon, though i've yet to meet a blind minetest player, so perhaps this isn't so important. 

using [beolingus](https://dict.tu-chemnitz.de/dings.cgi?lang=en&query=Fiber), Faser seems to be the most literal translation for the relevant meaning, though it looks like it's as non-specific in german as it is in english, and may introduce unintended ambiguities. should you be able to build unified inventory bags, or ship sails, out of this fiber, or just string and wool?

Wuzzy2:
I changed the translation of "Fiber" from "Kabel" to "Faser".

Wuzzy2:
Bump.

mesecons_delayer/locale/mesecons_delayer.de.tr
```diff
@@ -0,0 +1,3 @@
+# textdomain: mesecons_delayer
+Delayer=Verzögerer
+You hacker you=Sie Hacker!
```
sfan5, line 3: missing `!`?

mesecons_insulated/locale/mesecons_insulated.de.tr
```diff
@@ -0,0 +1,2 @@
+# textdomain: mesecons_insulated
+Straight Insulated Mesecon=Isolierte Mesecongerade
```
sfan5, line 2: maybe it'd be better to change it to just "Isolated Mesecon"

mesecons_pistons/locale/mesecons_pistons.de.tr
```diff
@@ -0,0 +1,7 @@
+# textdomain: mesecons_pistons
+Piston=Kolben
+Activated Piston Base=Aktivierter Kolbenkörper
+Piston Pusher=Kolbenschieber
+Sticky Piston=Klebriger Kolben
```
sfan5, line 5: vllt. "Haftender Kolben"

sfan5 (approved):
LGTM otherwise

Wuzzy2:
* "You hacker you" had no "!" so I removed it from German now
* "Straight Insulated Mesecon": Sounds like you request a change to the source string which I'm obviously not going to do in a pure translation PR
* "Sticky Piston": done

Is this OK?

sfan5 (approved):

PR #639, opened 2023-04-18
MVPS: don't move non-physical objects

fluxionary:
problem: you can push item entities off the top of the pedestals from Tenplus1's itemframes mod, and then they become immortal ghosts. i see that there's a `mesecon.register_mvps_unmov`, and registering the entities w/ that would certainly be one solution, but i don't think non-physical things should be getting pushed around in the first place, hence this PR.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -274,27 +274,32 @@ function mesecon.mvps_move_objects(pos, dir, nodestack, movefactor)
 	dir = vector.multiply(dir, movefactor)
 	for id, obj in pairs(minetest.get_objects_inside_radius(pos, #nodestack + 1)) do
 		local obj_pos = obj:get_pos()
-		local cbox = obj:get_properties().collisionbox
-		local min_pos = vector.add(obj_pos, vector.new(cbox[1], cbox[2], cbox[3]))
-		local max_pos = vector.add(obj_pos, vector.new(cbox[4], cbox[5], cbox[6]))
+		local props = obj:get_properties()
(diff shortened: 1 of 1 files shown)


Desour:
It's likely this change will break existing functioning mods. Just take items in itemframes as an example, they should always be moved with the frame.
Also, the code in this PR doesn't even seem to allow to opt-in for non-physical objs.
It's better if individual mods fix this with the provided callback.

fluxionary:
> It's likely this change will break existing functioning mods

possibly, but it's also likely to fix "bugs" with other mods. 

> take items in itemframes as an example, they should always be moved with the frame.

itemframes and pedestals are already registered as MVPS stoppers - they don't want to be moved. 

> It's better if individual mods fix this with the provided callback.

IMO it's better if mods treat the world and its mechanics as something they need permission to modify, and not dictate that every other mod in the ecosystem has to conform to their intentions. mods that are part of a "game" are one thing - games define behavior - but mods that try to be agnostic about the game and other random mods on the server shouldn't make messes, or should be responsible if they do. 

> the code in this PR doesn't even seem to allow to opt-in for non-physical objs.

i like this suggestion, i'll implement such a feature if my arguments against the other objections are accepted.

mruncreative:
We need to agree on some form of general entity unmovability across games and mods. I made an engine issue for that. Whether it will be handled in the engine or by a community standard, we need to handle it the same.

**Entities that don't want to be moved by pistons typically don't want to be moved by any foreign mods.**

https://github.com/luanti-org/luanti/issues/15751

Issue #638, opened 2023-04-04
Effectors that are conductors


Desour:
Defining nodes that are a conductor and an effector at the same time doesn't work.
Example:
```lua
local rules_conductor = {
	vector.new( 1, 0, 0),
	vector.new(-1, 0, 0),
}

local rules_effector = {
	vector.new(0, 0, 1),
}

minetest.register_node("mesecons_conduct_effect:node_off", {
	description = "Mesecon Conductor and Effector",
	drawtype = "normal",
	tiles = {"default_stone.png^mesecons_silicon.png"},
	groups = {dig_immediate=2},
	mesecons = {
		conductor = {
			state = mesecon.state.off,
			offstate = "mesecons_conduct_effect:node_off",
			onstate = "mesecons_conduct_effect:node_on",
			rules = rules_conductor,
		},
		effector = {
			rules = rules_effector,
			action_on = function (pos, node)
				minetest.log("(c off) action_on at "..vector.to_string(pos))
			end,
			action_off = function (pos, node)
				minetest.log("(c off) action_off at "..vector.to_string(pos))
			end,
			action_change = function (pos, node)
				minetest.log("(c off) action_change at "..vector.to_string(pos))
			end,
		}
	},
})

minetest.register_node("mesecons_conduct_effect:node_on", {
	description = "You hacker you!",
	drawtype = "normal",
	tiles = {"default_stone.png^mesecons_silicon.png"},
	drop = "mesecons_conduct_effect:node_off",
	groups = {dig_immediate=2, not_in_creative_inventory=1},
	mesecons = {
		conductor = {
			state = mesecon.state.on,
			offstate = "mesecons_conduct_effect:node_off",
			onstate = "mesecons_conduct_effect:node_on",
			rules = rules_conductor,
		},
		effector = {
			rules = rules_effector,
			action_on = function (pos, node)
				minetest.log("(c on) action_on at "..vector.to_string(pos))
			end,
			action_off = function (pos, node)
				minetest.log("(c on) action_off at "..vector.to_string(pos))
			end,
			action_change = function (pos, node)
				minetest.log("(c on) action_change at "..vector.to_string(pos))
			end,
		}
	},
})
```
One could expect that the `action_*` callbacks get executed whenever some effector-rule input changes.

I'm unsure if this is a bug report or feature request.

numberZero:
See also: #406

PR #648, opened 2023-07-10
Feat:  effector with conductor

snowyu:
Issue: #638

Due to the immediate changes made by the conductor, while the effector is lagged and executed in a queue, the conductor must always `swap_node_force` to avoid an infinite loop. 

The best solution is to refactor the effector and conductor code to allow for power transmission to also be executed in a queue. In my opinion, the conductor is a special type of effector and should not be treated differently. The conductor should be derived from the abstract effector class, and its default action should be to check and `swap_node`.

### DIFF
--- a/.test_fixtures/mesecons.lua
+++ b/.test_fixtures/mesecons.lua
@@ -148,7 +148,7 @@ do
 		table.insert(mesecon._test_eff_conductor_events, {"off", pos})
 		minetest.swap_node(pos, {name = "mesecons:test_effect_conductor_off", param2 = node.param2})
 	end
-	local function action_change(pos, node, rule_name, new_state)
+	local function action_change(pos, _node, _rule_name, _new_state)
 		if mesecon.do_overheat(pos) then
 			table.insert(mesecon._test_eff_conductor_events, {"overheat", pos})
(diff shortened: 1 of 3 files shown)


Desour:
>The best solution is to refactor the effector and conductor code to allow for power transmission to also be executed in a queue. In my opinion, the conductor is a special type of effector and should not be treated differently. The conductor should be derived from the abstract effector class, and its default action should be to check and `swap_node`.

There are builds that require to have instantaneous conductors. And idk how you could avoid infinite loops.
Imo, the mechanics of conductors should not be changed in such a way.

---

Code has some style issues. I.e. mesecons doesn't usually use camelCase.

---

Doing things like checking if it's a conductor, and then doing things with the node as effector, doesn't make much sense to me.

---

It is important that the effector part of a node is only triggered if the signal comes via an effector rule, and that the conductor part is only used if the signal comes via a conductor rule. It doesn't look like the PR does this.

---

If someone wants to have a conductor, that also does something else when it changes state, they can copy the conductor's rules to their effector rules.

---

I should've given feedback earlier, sry.
If you don't want to work on this anymore, I might pick this up sooner or later (or much later).

PR #659, opened 2023-12-16
fix issue 645: prevent pistons from pushing blocks into their current position, deleting them.

lolbinarycat:
this isn't the most efficient fix, but as far as i can tell, anything better would require changing the public api, and i don't want to break any other mods that may depend on mesecons_mvps.

some things i noticed while fixing this bug:
* when pushing a sticky block, the piston pushing the block actually gets dragged along behind it.  the only reason this isn't visible is because a piston extension places two nodes. 
* an immovable node behind a sticky block will prevent that block from being moved forward.

### DIFF
--- a/mesecons_pistons/init.lua
+++ b/mesecons_pistons/init.lua
@@ -84,7 +84,16 @@ local piston_on = function(pos, node)
 	local pistonspec = get_pistonspec(node.name, "offname")
 	local dir = vector.multiply(minetest.facedir_to_dir(node.param2), -1)
 	local pusher_pos = vector.add(pos, dir)
+	local behind_pos = vector.subtract(pos, dir)
 	local meta = minetest.get_meta(pos)
+	-- NOTE: this gets calculated twice, but fixing this would mean changing the public api.
+	local stack = mesecon.mvps_get_stack(pusher_pos, dir, max_push, meta:get_string("owner"))
+	for _, n in ipairs(stack) do
+		-- fix issue 645
+		if vector.equals(n.pos, behind_pos) then
+			return
+		end
+	end
 	local success, stack, oldstack = mesecon.mvps_push(pusher_pos, dir, max_push, meta:get_string("owner"))
 	if not success then
 		if stack == "protected" then


mesecons_pistons/init.lua
```diff
@@ -84,7 +84,16 @@ local piston_on = function(pos, node)
...
 	local meta = minetest.get_meta(pos)
+	-- NOTE: this gets calculated twice, but fixing this would mean changing the public api.
+	local stack = mesecon.mvps_get_stack(pusher_pos, dir, max_push, meta:get_string("owner"))
+	for _, n in ipairs(stack or {}) do
+		-- fix issue 645
+		if vector.equals(n.pos, behind_pos) then
+			return
+		end
+	end
 	local success, stack, oldstack = mesecon.mvps_push(pusher_pos, dir, max_push, meta:get_string("owner"))
```
SmallJoker, line 97: Shouldn't `mvps_push` return `success = false` when the node cannot be pushed? That would seem to be the logical behaviour.

numberZero: But a piston *can* be pushed. It just shouldn’t be able to push *itself*. Unlike a movestone for which that shouldn’t be a problem.

lolbinarycat: my first thought was swapping the piston for the immoveable extended version *before* calling mvps_push (then swap it back if it was unsuccessful), which would return false and also wouldn't have any significant performance overhead.

unfortunately, a sticky block on the face of the piston will grab said piston and try to pull it along, even if it is immoveable.

this is different from the behavior or minecraft slime blocks, which have no problems pulling away from immoveable blocks.

however, the purpose of this commit is to fix a bug, not to change behavior, so i opted for this instead, which doesn't change any public apis, ensuring it doesn't break other mods.

lolbinarycat: > Shouldn't `mvps_push` return `success = false` when the node cannot be pushed? That would seem to be the logical behaviour.

the biggest reason i don't implement the new logic in mvps_push is because pistons and moveblocks have fundimentally different behavior:
* pistons get bigger when pushing things
* moveblocks stay the same size

one option would be to add an `is_piston` or `piston_pos`  parameter to `mvps_push`, and somehow handle the logic there, but once again, i didn't want to amend the public api on my first contribution.

the simplest change to fix the performance loss would be to add an optional nodes parameter to `mvps_push`, then add `nodes = nodes or mesecons.mvps_get_stack(<...>)`.

i can implement any of these changes if desired.

PR #662, opened 2023-12-17
optimize mvps_push_or_pull

lolbinarycat:
this function is a big bottleneck, since it contains several layers of nested loops, and has quite a few allocations in the innermost levels.

i did my best to optimize it, and i'm only confident in about half my changes, although some imprecice testing implied as much as a 10x speedup.

if someone could benchmark this properly and see what changes are worth it, that would be greatly appreciated.

### DIFF
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -70,8 +70,8 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
 	local nodes = {}
 	local pos_set = {}
 	local frontiers = mesecon.fifo_queue.new()
-	--local vector_equals = vector.equals
 	frontiers:add(vector.new(pos))
+	-- micro-optimization: lift local definitions out of loop
 	local nodedef
 	local np_hash
 	local nn
@@ -100,7 +100,7 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
 			-- If adjacent node is sticky block and connects add that
 			-- position
 			for _, r in ipairs(mesecon.rules.alldirs) do
-				local adjpos = vector.add(np, r)
+				adjpos = vector.add(np, r)
 				-- optimization: don't check blocks that are already part of the stack
 				if not pos_set[minetest.hash_node_position(adjpos)] then
 					adjnode = minetest.get_node(adjpos)
--- a/mesecons_mvps/init.lua
+++ b/mesecons_mvps/init.lua
@@ -95,16 +95,19 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
 			-- position
 			for _, r in ipairs(mesecon.rules.alldirs) do
 				local adjpos = vector.add(np, r)
-				local adjnode = minetest.get_node(adjpos)
-				if minetest.registered_nodes[adjnode.name]
-				and minetest.registered_nodes[adjnode.name].mvps_sticky then
-					local sticksto = minetest.registered_nodes[adjnode.name]
-						.mvps_sticky(adjpos, adjnode)
-
-					-- connects to this position?
-					for _, link in ipairs(sticksto) do
-						if vector.equals(link, np) then
-							frontiers:add(adjpos)
+				-- optimization: don't check blocks that are already part of the stack
+				if not pos_set[minetest.hash_node_position(adjpos)] then
+					local adjnode = minetest.get_node(adjpos)
+					if minetest.registered_nodes[adjnode.name]
+					and minetest.registered_nodes[adjnode.name].mvps_sticky then
+						local sticksto = minetest.registered_nodes[adjnode.name]
+							.mvps_sticky(adjpos, adjnode)
+	
+						-- connects to this position?
+						for _, link in ipairs(sticksto) do
+							if vector.equals(link, np) then
+								frontiers:add(adjpos)
+							end
 						end
 					end
 				end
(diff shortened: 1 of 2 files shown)


lolbinarycat:
apperently the CI uses a different lua version than minetest 5.7

since it lacks table.move.

not sure what to make of that

sfan5:
`table.move` was added in Lua 5.3 and appears to be supported as a non-standard extension in LuaJIT.
You can't use it.

Desour:
LuaJIT adds table.move. You can check at load if it exists, and implement `add_list` without it if it's not there.

mesecons_mvps/init.lua
```diff
@@ -71,40 +71,49 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
 	local pos_set = {}
 	local frontiers = mesecon.fifo_queue.new()
 	frontiers:add(vector.new(pos))
+	-- micro-optimization: lift local definitions out of loop
```
Desour, line 74: It would be new to me that lifting local variable declarations out of their scope is cheaper. May I ask where this is suggested?

lolbinarycat: this is based on personal testing, and is likely not worth it.

unless LuaJIT automatically does this, a local variable would have to have space for it allocated, which necessarily takes time.

mesecons_mvps/init.lua
```diff
@@ -95,16 +93,19 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
...
-				and minetest.registered_nodes[adjnode.name].mvps_sticky then
-					local sticksto = minetest.registered_nodes[adjnode.name]
-						.mvps_sticky(adjpos, adjnode)
-
-					-- connects to this position?
-					for _, link in ipairs(sticksto) do
-						if vector.equals(link, np) then
-							frontiers:add(adjpos)
+				-- optimization: don't check blocks that are already part of the stack
+				if not pos_set[minetest.hash_node_position(adjpos)] then
```
numberZero, line 97: What’s about r==dir?

mesecons_mvps/init.lua
```diff
@@ -81,12 +81,10 @@ function mesecon.mvps_get_stack(pos, dir, maximum, all_pull_sticky)
...
 			-- add connected nodes to frontiers
-			if minetest.registered_nodes[nn.name]
-			and minetest.registered_nodes[nn.name].mvps_sticky then
-				local connected = minetest.registered_nodes[nn.name].mvps_sticky(np, nn)
-				for _, cp in ipairs(connected) do
-					frontiers:add(cp)
-				end
+			local nodedef = minetest.registered_nodes[nn.name]
+			if nodedef and nodedef.mvps_sticky then
+				local connected = nodedef.mvps_sticky(np, nn)
```
numberZero, line 86: :+1:

mesecons_mvps/init.lua
```diff
@@ -259,9 +259,9 @@ function mesecon.mvps_push_or_pull(pos, stackdir, movedir, maximum, all_pull_sti
 		moved_nodes[i].meta = nodes[i].meta
 		moved_nodes[i].node_timer = nodes[i].node_timer
 	end
+	
 
-	on_mvps_move(moved_nodes)
-
+	on_mvps_move(moved_nodes)
```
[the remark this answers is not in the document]
lolbinarycat: yeah this must've happened when i was deleting and inserting lines.  are random formatting changes considered a major problem?

mesecons_mvps/init.lua
lolbinarycat: i don't fully understand, do you mean i should add an additional optimization to skip this code when r==dir, since that block will already have been added to the queue?

i guess that's a missed optimization then, yeah.  do you want me to add it?

lolbinarycat:
all this code review is nice, but what would really be helpful is for someone benchmark this to see if the speedup is worth me fixing everything up.

or if someone could point me in the direction of some lua/minetest benchmarking tools i could try to do it myself.

my limited testing was based off of realtime and framerate, both of which can be influenced by other processes.

SmallJoker:
Minetest has an internal profiler:

 * Settings to enable it: https://github.com/minetest/minetest/blob/335af393f09b3629587f14d41a90ded4a3cbddcd/builtin/settingtypes.txt#L1734-L1762
 * Chat command to export: https://github.com/minetest/minetest/blob/335af393f09b3629587f14d41a90ded4a3cbddcd/builtin/profiler/init.lua#L45-L46

According to `mesecons/actionqueue.lua`, the actions are executed each globalstep, thus you should see a difference in one of the mesecons -> `globalstep[??]` entries after running `/profiler save txt` (for comparison).

lolbinarycat:
> Minetest has an internal profiler:
> 
>     * Settings to enable it: https://github.com/minetest/minetest/blob/335af393f09b3629587f14d41a90ded4a3cbddcd/builtin/settingtypes.txt#L1734-L1762
> 
>     * Chat command to export: https://github.com/minetest/minetest/blob/335af393f09b3629587f14d41a90ded4a3cbddcd/builtin/profiler/init.lua#L45-L46
> 
> 
> According to `mesecons/actionqueue.lua`, the actions are executed each globalstep, thus you should see a difference in one of the mesecons -> `globalstep[??]` entries after running `/profiler save txt` (for comparison). I however do not know what the best way to test this optimization would be - 100 pistons with sand on top of them?

personally i used stacks of slimeblocks on top of sticky pistons connected to a 1-tick clock.

PR #663, opened 2023-12-19
fix 1-tick pulses causing delayers to get stuck in the on state

### LINKED ISSUE #656: Delays cannot handle 1 tick state changes
delays get stuck with 1 tick pulses, and will stick in the "on" position
I have noticed this behavior with a machine that has a detector tube from pipeworks to handle timing.

lolbinarycat:
Fixes #656

i believe the cause of this bug is a chain of events as such:
* delayer receives effector_on, and starts a timer
* delayer receives effector_off, however the node has not been updated somehow (perhaps due to caching somewhere), and thus this is dropped
* delayer swaps node into on state

 this may be caused by an underlying bug somewhere in the action queue, but simply changing the unpowered state so it also responds to power off events seems to do the trick just fine.

### DIFF
--- a/mesecons_delayer/init.lua
+++ b/mesecons_delayer/init.lua
@@ -62,6 +62,8 @@ local def = {
 	sounds = mesecon.node_sound.stone,
 	on_blast = mesecon.on_blastnode,
 	drop = "mesecons_delayer:delayer_off_1",
+	delayer_onstate = "mesecons_delayer:delayer_on_"..tostring(i),
+	delayer_offstate = "mesecons_delayer:delayer_off_"..tostring(i),
 }
 
 -- Deactivated delayer definition defaults
@@ -93,7 +95,6 @@ local off_state = {
 			param2 = node.param2
 		})
 	end,
-	delayer_onstate = "mesecons_delayer:delayer_on_"..tostring(i),
 	mesecons = {
 		receptor =
 		{
@@ -135,7 +136,6 @@ local on_state = {
 			param2 = node.param2
 		})
 	end,
-	delayer_offstate = "mesecons_delayer:delayer_off_"..tostring(i),
 	mesecons = {
 		receptor =
 		{
--- a/mesecons_delayer/init.lua
+++ b/mesecons_delayer/init.lua
@@ -103,6 +103,7 @@ local off_state = {
 		effector =
 		{
 			rules = delayer_get_input_rules,
+			action_off = delayer_deactivate,
 			action_on = delayer_activate
 		}
 	},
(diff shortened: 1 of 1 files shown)


SmallJoker:
How can I reproduce the issue to then test this PR's fix?

SmallJoker:
```
2024-01-06 12:43:08: ERROR[Main]: ServerError: AsyncErr: Lua: Runtime error from mod '??' in callback environment_Step(): MT/builtin/common/item_s.lua:229: bad argument #1 to '__index' (string expected, got nil)
2024-01-06 12:43:08: ERROR[Main]: stack traceback:
2024-01-06 12:43:08: ERROR[Main]: 	[C]: in function '__index'
2024-01-06 12:43:08: ERROR[Main]: 	MT/builtin/common/item_s.lua:229: in function <MT/builtin/common/item_s.lua:228>
2024-01-06 12:43:08: ERROR[Main]: 	[C]: in function 'swap_node'
2024-01-06 12:43:08: ERROR[Main]: 	MT/mods/mesecons/mesecons_delayer/init.lua:20: in function 'action_off'
2024-01-06 12:43:08: ERROR[Main]: 	MT/mods/mesecons/mesecons/internal.lua:207: in function <MT/mods/mesecons/mesecons/internal.lua:200>
2024-01-06 12:43:08: ERROR[Main]: 	MT/mods/mesecons/mesecons/actionqueue.lua:137: in function 'execute'
2024-01-06 12:43:08: ERROR[Main]: 	MT/mods/mesecons/mesecons/actionqueue.lua:111: in function <MT/mods/mesecons/mesecons/actionqueue.lua:73>
2024-01-06 12:43:08: ERROR[Main]: 	MT/builtin/common/register.lua:26: in function <MT/builtin/common/register.lua:12>
```

when placing an inverter into the indicated location:
error reproducer

lolbinarycat:
@SmallJoker thanks, should be fixed now!

SmallJoker:
@lolbinarycat Thank you. I can no longer reproduce that error. However, I'd still like to know which circuit bug this PR aims to fix. Where can I see the change in behaviour?

lolbinarycat:
@SmallJoker i didn't find a 100% reliable way to replicate it, but what i did was make a 1 tick piston clock (a sticky piston with a switch on its face, then use wire to connect the switch to the piston), then take a line from the clock to the back of a delayer.

if you destroy the wire directly behind the delayer with the correct timing, the delayer will get stuck in the "on" position despite there being no wire leading into it.  also when i was doing that i was also intentionally lagging the game, which may or may not be required.

after this change, i was unable to replicate the bug with this method.

SmallJoker:
The singleplayer server tick is a bit too fast for this type of oscillator. OpenAL is not too happy about it. However, with this setup I did reproduce the bug and the PR does indeed fix it.

grafik

SmallJoker (approved):
Will merge in a few days unless there are objections.

PR #667, opened 2024-03-12
Remove NaN values from digiline messages

OgelGames:
NaN can cause problems (and crashes) if the receiving node doesn't check for it, this PR makes it so NaN values are removed from digiline messages.


OgelGames:
@SmallJoker

Niklp09 (approved):
lgtm

SmallJoker:
How can I reproduce the bug and hence confirm the fix?

OgelGames:
> How can I reproduce the bug and hence confirm the fix?

See here: https://github.com/mt-mods/digistuff/issues/54

SmallJoker:
Thanks. Issue reproduced and the PR works as expected.

PR #670, opened 2024-04-19
Check for digiline message field type

Emojigit:
This PR fixes a fatal error: if a malformed digiline table is passed into a node detector, the server may crash because of a mismatched data type.

For example, this weird backtrace happened because technic is sending "malformed" tables:

```
ERROR[Main]: ServerError: AsyncErr: Lua: Runtime error from mod '??' in callback node_on_timer(): $HOME/.minetest/mods/mesecons/mesecons_detector/init.lua:63: bad argument #2 to 'set_string' (string expected, got table)
ERROR[Main]: stack traceback:
ERROR[Main]:       [C]: in function 'set_string'
ERROR[Main]:       $HOME/.minetest/mods/mesecons/mesecons_detector/init.lua:63: in function 'action'
ERROR[Main]:       $HOME/.minetest/mods/digicontrol/override.lua:45: in function 'transmit'
ERROR[Main]:       $HOME/.minetest/mods/digicontrol/override.lua:78: in function 'receptor_send'
ERROR[Main]:       ...test/mods/technic/technic/machines/switching_station.lua:112: in function <...test/mods/technic/technic/machines/switching_station.lua:87>
```

### DIFF
--- a/mesecons_detector/init.lua
+++ b/mesecons_detector/init.lua
@@ -223,10 +223,10 @@ local node_detector_digiline = {
 
 			if type(msg) == "table" then
 				if msg.distance or msg.scanname then
-					if msg.distance then
+					if msg.distance and type(msg.distance) == "string" then
 						meta:set_string("distance", msg.distance)
 					end
-					if msg.scanname then
+					if msg.scanname and type(msg.scanname) == "string" then
 						meta:set_string("scanname", msg.scanname)
 					end
 					node_detector_make_formspec(pos)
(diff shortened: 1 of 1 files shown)


mesecons_detector/init.lua
```diff
@@ -223,10 +223,10 @@ local node_detector_digiline = {
 
 			if type(msg) == "table" then
 				if msg.distance or msg.scanname then
-					if msg.distance then
+					if msg.distance and type(msg.distance) == "string" then
```
SmallJoker, line 226: Here and below: superfluous `msg.distance` check because `type` would return `nil` in that case.

Niklp09:
@Emojigit bump.

mesecons_detector/init.lua
Emojigit: Solved in 110b3c4

PR #671, opened 2024-05-15
[mesecons_doors] Improve code quality and fix Voxelgarden support

rudzik8:
The main code quality improvement here is the functionalization of code: now trapdoors are also overriden using their own function, and both `meseconify_door` and `meseconify_trapdoor` are now ran inside for loops with local lists of base node names. In the future, this could help refactor this mod entirely to not list doors manually but parse through the doors.registered_* variables, ensuring that every door works.

As a nice bonus, [Voxelgarden](https://codeberg.org/voxelgarden/voxelgarden) support is fixed, and potentially so is support for other games with exotic implementations of the doors mod. No longer are we assuming that the existence of `doors.get` means that all the trapdoors exist too; now, if the meseconify function doesn't find their definition, it just returns without crashing.

P.S. And yes, I'm the current Voxelgarden maintainer.

### DIFF
--- a/mesecons_doors/init.lua
+++ b/mesecons_doors/init.lua
@@ -83,18 +83,18 @@ local doors_list = {
 for i=1,#doors_list do meseconify_door(doors_list[i]) end
 
 -- Trapdoor
-local function trapdoor_switch(pos, node)
-	local state = minetest.get_meta(pos):get_int("state")
-
-	if state == 1 then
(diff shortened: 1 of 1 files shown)


SmallJoker (approved):
Works. Tested xpanes and doors. Will merge in a few days unless there are objections in the meantime.

PR #672, opened 2024-05-18
Add node detector blacklist api

Niklp09:
Useful for mods which register "airlike" nodes, e.g. wielded_light for invisible light nodes.

### DIFF
--- a/mesecons_detector/init.lua
+++ b/mesecons_detector/init.lua
@@ -157,6 +157,23 @@ minetest.register_abm({
 -- Node detector
 -- Detects the node in front of it
 
+local blacklist = {
+	air = true,
+	ignore = true,
+}
+
+function mesecon.node_detector_blacklist_add(node)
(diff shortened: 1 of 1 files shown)


SmallJoker:
1. The name is misleading. It *only* backlists nodes in case no node name is specified by the player. I'd expect blacklists to apply in all cases. I'd suggest to rename it to "airlike" or "empty" nodes.
2. Might it make sense to generally ignore all nodes with `def.drawtype == "airlike" and not def.drowning` for simplicity? That would automatically cover `wielded_light`.

Niklp09:
> It only backlists nodes in case no node name is specified by the player.

Do you think it's better to ignore (blacklisted) nodes when they're specified? I left this since it's maybe helpful for some players :thinking: .

> Might it make sense to generally ignore all nodes with def.drawtype == "airlike" and not def.drowning for simplicity?

I'll add this.

PR #677, opened 2024-07-20
MineClonia compat

nonfreegithub:


nonfreegithub:
globals added

SmallJoker:
Thanks. Will merge in a few days unless there are objections.

PR #679, opened 2024-08-13
Fix mesecons displayed as unknown(??) mod in the profiler

zmv7:
Currently there is the replacement of existing globalstep function, but it doesn't consider callback origins, therefore, the engine doesn't know what mod is the owner of the new function. The result - there is unknown entry in the profiler output, displayed as `??` modname with `*` instrumentation.

This patch fixes it by setting callback origin of the replaced globalstep function

### DIFF
--- a/mesecons/actionqueue.lua
+++ b/mesecons/actionqueue.lua
@@ -70,7 +70,17 @@ end
 -- However, even that does not work in some cases, that's why we delay the time the globalsteps
 -- start to be execute by 4 seconds
 
+local m_time = 0
+local resumetime = mesecon.setting("resumetime", 4)
+
 local function globalstep_func(dtime)
+	-- don't even try if server has not been running for XY seconds; resumetime = time to wait
+	-- after starting the server before processing the ActionQueue, don't set this too low
+	if m_time < resumetime then
+		m_time = m_time + dtime
+		return
+	end
+
 	local actions = queue.actions
 	-- split into two categories:
 	-- actions_now: actions to execute now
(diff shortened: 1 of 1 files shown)


mesecons/actionqueue.lua
```diff
@@ -127,6 +127,10 @@ do
 		end
 		-- replace this globalstep function
 		minetest.registered_globalsteps[globalstep_func_index] = globalstep_func
+		minetest.callback_origins[globalstep_func] = {
+			mod = minetest.get_current_modname() or "mesecons",
```
SmallJoker, line 131: I don't think you need to assign this more than once. This value should never be removed.

zmv7: But what does *"should not be removed by Minetest"* actually mean? I just adding the new pair into the `minetest.callback_origins` table. Or you're about `or "mesecons"` section?
Well, as an alternative, I can move 4-seconds interval into the main globalstep and so replacement will not be needed

sfan5: > I can move 4-seconds interval into the main globalstep and so replacement will not be needed

That would be cleaner and I prefer it.

sfan5 (approved):

SmallJoker (approved):
Yes. LGTM.
If you wanted to optimize it even further, `minetest.after` might be an option to delay the globalstep registration. But then again, you'd also have to assign the function origin once.

zmv7:
> Yes. LGTM. If you wanted to optimize it even further, `minetest.after` might be an option to delay the globalstep registration. But then again, you'd also have to assign the function origin once.

Callbacks registered using `minetest.after` also displayed as `??` unfortunately

SmallJoker:
@zmv7 Hence the 2nd sentence, by which I was referring to `minetest.callback_origins`.

PR #685, opened 2024-10-17
Fix crash when `doors.registered_doors` isn't available

Niklp09:
Fixes https://github.com/minetest-mods/mesecons/issues/684


SmallJoker:
I do not recommend merging this before the root cause is known. Otherwise this is cargo-cult code from the beginning.

Issue #689, opened 2024-11-19
Allow use of sha256 by code in the lua controller


Darin755:
I was working on making some fun things with digilines and lua controllers when I discovered that there doesn't seem to be a way to do a sha256 hash from the lua controller.

What I am trying to build is a secure entrance that allows a player to either authencate with there username (player detector) and/or a password. I want this to be secure so I was hoping to hash the credentials. The idea is that even if a "bad guy" gets access to the lua controller they still will not have the password. This is overkill but I think that is part of the fun.

Darin755:
Thanks!

PR #690, opened 2024-11-21
added sha256 function to lua controller

Darin755:
This is a simple solution for #689


Darin755:
sha1 is not secure as it has known hash collisions (where to different inputs get the same hash)

I could probably add this to the docs. However, the lua controller docs are dated and not great so if get around to working on docs I will likely make some big changes.

sfan5:
It looks like the documentation isn't even in this repo, so should be fine as is.

PR #694, opened 2024-12-18
MineClonia redstone update support

nonfreegithub:
Recently MineClonia has [reworked the redstone mods](https://codeberg.org/mineclonia/mineclonia/pulls/2034)

It adds support to


mesecons_gamecompat/init.lua
```diff
@@ -14,7 +14,7 @@ if minetest.get_modpath("default") then
 	dofile(minetest.get_modpath("mesecons_gamecompat").."/compat_mtg.lua")
 end
 
-if minetest.get_modpath("mcl_core") then
+if minetest.get_modpath("mcl_core") and minetest.get_modpath("mcl_trees") then
```
SmallJoker, line 17: Why `mcl_trees` ? I don't see any such dependency.

SmallJoker: Are the used API functions of Voxelibre and MineClonia incompatible? Why can't we support both?

sfan5: Also has either Voxelibre or MCL considered adding a way to detect the fork?

nonfreegithub: changed, please merge

j-r: > Are the used API functions of Voxelibre and MineClonia incompatible? Why can't we support both?

I think the bigger problem are the different levels of pollution of the mesecons namespace. VL still has a heavily modified mesecons built in as its redstone implementation. Running mesecons on top of VL seems to require more than a few tweaks to the gamecompat code. MCLA recently gained a separate redstone implementation and dropped the mesecons based one previously shared with all mcl2 based games. With the few fixes in this PR mesecons will run on top of MCLA and is usable with some caveats:
- for world migration mcla still defines a lot of aliases in the mesecons namespace to the new redstone items; most are harmless because they get overwritten by actual mesecons items, but at least some recipes are broken because of this (e.g. the ones using mesecons:mesecon which is also just an alias in mesecons, or the mesecons torch where the craft is registered before the item causing the craft to output mcla redstone torches)
- mcla also dropped mvps in favor of a new system
- mcla redstone cannot power mesecons machines and vice versa (but its reasonably easy to build optical converters using observers)

TL;DR it looks like mcla is the only mcl* that can be supported reasonably by mesecons currently. The fact that until now nobody cared that the mcl gamecompat code has obvious typos seems to support that.

j-r: > Also has either Voxelibre or MCL considered adding a way to detect the fork?

For most purposes they are still compatible. In the context of mesecons compatibility the existence of the `mcl_redstone` mod seems like an adequate discriminator.

SmallJoker (approved):
Looks about correct. Does not break Minetest Game, thus :+1:. Will merge in a few days unless there are objections.

mruncreative:
SmallJoker, I already integrated those changes into my merge request. So this one is now superfluous.

SmallJoker:
Superseded by a0580f9f20fd63eba87

PR #695, opened 2024-12-20
Limit chatcommand parameter length in command blocks

luk3yx:
This PR adds a configurable maximum length for chatcommand parameters. Long parameters can possibly be used to cause crashes and other problems since mods don't expect parameters to be very long.

Hopefully the default of 10,000 bytes won't break any existing command blocks.

This issue this PR aims to fix somewhat mitigated by the formspec input length limit and https://github.com/minetest/minetest/commit/9c9309cdbb053598aaf08506928a4824e78b4622 (introduced in 5.7.0), however other mods (like [irc](https://github.com/minetest-mods/irc/pull/53)) don't have a message size limit, though a 10 kB message will probably do just as much harm, server owners would have to configure the limit if they can live with some command blocks no longer working.

### DIFF
--- a/mesecons_commandblock/init.lua
+++ b/mesecons_commandblock/init.lua
@@ -158,8 +158,8 @@ local function commandblock_action_on(pos, node)
 			return
 		end
 		if #param > param_maxlen then
-			minetest.chat_send_player(owner, "Command parameters can only be " ..
-				param_maxlen .. " bytes long")
+			minetest.chat_send_player(owner, "Command parameters are limited to max. " ..
+				param_maxlen .. " bytes.")
 			return
 		end
 		local has_privs, missing_privs = minetest.check_player_privs(owner, cmddef.privs)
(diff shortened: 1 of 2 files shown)


sfan5 (approved):
looks good

mesecons_commandblock/init.lua
```diff
@@ -156,6 +157,11 @@ local function commandblock_action_on(pos, node)
 			minetest.chat_send_player(owner, "The command "..cmd.." does not exist")
 			return
 		end
+		if #param > param_maxlen then
+			minetest.chat_send_player(owner, "Command parameters can only be " ..
+				param_maxlen .. " bytes long")
```
SmallJoker, line 162: Common sense sould tell you that `n < limit` is acceptable. I don't trust common sense, thus this suggestion:
```suggestion
			minetest.chat_send_player(owner, "Command parameters are limited to max. " ..
				param_maxlen .. " bytes.")
```

PR #697, opened 2024-12-26
Add missing `use_texture_alpha`

MoNTE48:
Seems like the warnings should be more explicit. They can be overlooked for years...

### DIFF
--- a/mesecons_receiver/init.lua
+++ b/mesecons_receiver/init.lua
@@ -30,6 +30,8 @@ local receiver_get_rules = mesecon.horiz_rules_getter({
 	{x = 0, y = 0, z = -2},
 })
 
+local use_texture_alpha = minetest.features.use_texture_alpha_string_modes and "clip" or nil
+
 mesecon.register_node("mesecons_receiver:receiver", {
 	drawtype = "nodebox",
 	paramtype = "light",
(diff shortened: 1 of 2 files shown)


mesecons_receiver/init.lua
```diff
@@ -143,6 +144,7 @@ mesecon.register_node("mesecons_receiver:receiver_down", {
 	groups = {dig_immediate = 3, not_in_creative_inventory = 1},
 	drop = "mesecons:wire_00000000_off",
 	sounds = mesecon.node_sound.default,
+	use_texture_alpha = "clip",
```
sfan5, line 147: all the nodeboxes in this PR are opaque, so no need for `clip`
also compatibility code is missing, like e.g. here:
```
mods/mesecons-mp/mesecons_lamp/init.lua
14:local use_texture_alpha = minetest.features.use_texture_alpha_string_modes and "opaque" or nil
19:	use_texture_alpha = use_texture_alpha,
44:	use_texture_alpha = use_texture_alpha,
```

MoNTE48:
> **all** the nodeboxes in this PR are opaque, so no need for clip

at the same time someone with a texturepack...

sfan5:
I suppose it doesn't hurt to allow that. `clip` it is then.

PR #700, opened 2025-01-29
Fix mesecons_gamecompat mvps hardcoding issues

### LINKED ISSUE #699: mesecons_gamecompat mvps hardcoding issues
I ran into more issues with nodes being hardcoded.
My beds I registered to the beds mod are not registered as mesecons mvps stoppers.
For both MTG and MCL we should check for beds and door mods inside a core.register_on_mods_loaded and register all their beds and doors from their apis instead of making crazy lists ourselves.

mruncreative:
Fixes #699
Registers mvps stoppers for all door and bed nodes registered using beds, mcl_beds, doors, mcl_doors mods.
In mtg trapdoors have the group door and need to be obtained through the api while beds don't have a registered_beds and need to be detected by their group.
In mcl_doors, doors and trapdoors have separate groups and no mcl_doors.registered_doors table.
The wooden trapdoor was not an mvps stopper, so I removed the other trapdoors from being mvps stoppers for no reason.

### DIFF
--- a/mesecons_gamecompat/compat_mcl.lua
+++ b/mesecons_gamecompat/compat_mcl.lua
@@ -41,3 +41,16 @@ end
 -- Textures
 
 mesecon.texture.steel_block = "default_steel_block.png"
+
+if minetest.get_modpath("mesecons_mvps") then
+	core.register_on_mods_loaded(function()
+		for _,v in pairs(core.registered_nodes) do
+			if v.groups and v.groups.bed then
+				mesecon.register_mvps_stopper(v.name)
+			end
+			if v.groups and v.groups.door then
+				mesecon.register_mvps_stopper(v.name)
+			end
+		end
+	end)
+end
--- a/mesecons_gamecompat/compat_mtg.lua
+++ b/mesecons_gamecompat/compat_mtg.lua
@@ -48,28 +48,22 @@ if minetest.get_modpath("mesecons_mvps") then
 	for _, name in ipairs({
 		"default:chest_locked",
 		"default:chest_locked_open",
-		"doors:door_steel_b_1", -- old style doors
-		"doors:door_steel_b_2", --
-		"doors:door_steel_t_1", --
-		"doors:door_steel_t_2", --
-		"doors:door_steel_a",   -- new style doors
-		"doors:door_steel_b",   --
-		"doors:door_steel_c",   --
-		"doors:door_steel_d",   --
 		"doors:hidden",
-		"doors:trapdoor_steel",
-		"doors:trapdoor_steel_open",
-		"beds:bed_bottom",
-		"beds:bed_top",
-		"beds:fancy_bed_bottom",
-		"beds:fancy_bed_top",
-		"xpanes:door_steel_bar_a",
-		"xpanes:door_steel_bar_b",
-		"xpanes:door_steel_bar_c",
-		"xpanes:door_steel_bar_d",
-		"xpanes:trapdoor_steel_bar",
-		"xpanes:trapdoor_steel_bar_open",
 	}) do
 		mesecon.register_mvps_stopper(name)
 	end
+	core.register_on_mods_loaded(function()
+		if minetest.get_modpath("doors") then
+			for k,_ in pairs(doors.registered_doors) do
+				mesecon.register_mvps_stopper(k)
+			end
+		end
+		if minetest.get_modpath("beds") then
+			for _,v in pairs(core.registered_nodes) do
+				if v.groups and v.groups.bed then
+					mesecon.register_mvps_stopper(v.name)
+				end
+			end
+		end
+	end)
 end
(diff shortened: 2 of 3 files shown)


mruncreative:
The checks fail because I used core instead of minetest somewhere and it's not specified as global to luacheck.

mruncreative:
I have moved mesecons_doors to gamecompat.
Reasoning:
1. Gamecompat already does compatibility with doors.
2. Mesecons Doors provides no API
3. Mesecons Doors is a subpackage of Mesecons while having a hard dependency on doors
4. Point 3 makes Mesecons harder to enable outside of mtg
5. Point 3 makes Mesecons be listed as incompatible with non-MTG games on ContentDB

mesecons_extrawires/init.lua
```diff
@@ -3,4 +3,6 @@ dofile(minetest.get_modpath("mesecons_extrawires").."/tjunction.lua");
 dofile(minetest.get_modpath("mesecons_extrawires").."/corner.lua");
 dofile(minetest.get_modpath("mesecons_extrawires").."/doublecorner.lua");
 dofile(minetest.get_modpath("mesecons_extrawires").."/vertical.lua");
-dofile(minetest.get_modpath("mesecons_extrawires").."/mesewire.lua");
+if core.registered_nodes[mese_nodename] then
```
SmallJoker, line 6: `mese_nodename` is not available in this scope. Luanti should log a warning to `debug.txt` similar to this: `Undeclared global variable "mese_nodename" accessed at (path)`.

mesecons_doors/mod.conf
```diff
@@ -1,3 +0,0 @@
-name = mesecons_doors
-depends = mesecons, doors
-optional_depends = xpanes
```
SmallJoker, line 3: I think it might be helpful for server owners to keep `mesecons_doors` present for a few months, as a transition phase. Preferably with an `init.lua` file that warns about the deprecation.

mesecons_extrawires/init.lua
mruncreative: I just noticed this myself trying to build with mesecons but good that you caught that.

mesecons_doors/mod.conf
mruncreative: It has no api, so it makes no sense for anything to require it. Also we absolutely want to get rid of that hard dependency of doors or CDB will list it as requirement for the whole modpack and not list Mineclonia as compatible.

SmallJoker: Oh actually never mind. I thought that missing mods is a blocking error that would prevent the world from loading. That's actually only true if another mod were to depend on the missing mod.
Hence my concern is invalid. This change is OK as-is.

SmallJoker:
`Check / Luacheck (pull_request) ` is currently failing because `.luacheckrc` lacks the `core` table, which is identical to `minetest`.

mruncreative:
The bit module is missing from the testing container. I don't think I can fix that on my end.

SmallJoker (commented):
Logged the `mesecon.register_mvps_stopper` calls; alphabetically sorted.
```
register	beds:bed_bottom
register	beds:bed_top
register	beds:fancy_bed_bottom
register	beds:fancy_bed_top
register	doors:door_glass_a
register	doors:door_glass_b
register	doors:door_glass_c
register	doors:door_glass_d
register	doors:door_obsidian_glass_a
register	doors:door_obsidian_glass_b
register	doors:door_obsidian_glass_c
register	doors:door_obsidian_glass_d
register	doors:door_steel_a
register	doors:door_steel_b
register	doors:door_steel_c
register	doors:door_steel_d
register	doors:door_wood_a
register	doors:door_wood_b
register	doors:door_wood_c
register	doors:door_wood_d
register	xpanes:door_steel_bar_a
register	xpanes:door_steel_bar_b
register	xpanes:door_steel_bar_c
register	xpanes:door_steel_bar_d
```

The trapdoors are missing. Is this intentional? If not --> loop through `doors.registered_trapdoors` to add those nodes.

mruncreative:
> The trapdoors are missing. Is this intentional? If not --> loop through doors.registered_trapdoors to add those nodes.

The wooden trapdoor was not an mvps stopper, so trapdoors in general do not need to be mvps blockers and sfan5 said the metal trapdoors were mvps stoppers because they can be locked. I think they can be dug though if they aren't protected. What is your take on that?
If you don't like it, I will add the metal trapdoors from MTG back to the hardcoded list.

mruncreative:
How about additional soft functions instead that can fail? Like soft_set_pos soft_set_velocity soft_set_accuracy.
Mods that apply stuff to unknown entities can use those insteads. Entities could define functions to override them and do other stuff with the information instead.

SmallJoker:
> I thought you'd have seen it.

Are you referring to the following message? https://irc.minetest.net/luanti/2025-01-31#i_6239224  Unfortunately I haven't seen that one.
Doors may glitch out, thus those should be stopper nodes. As for trapdoors - I am not sure, hence my question above whether this change was intentional. I suppose either way would be okay.

SmallJoker (approved):
Works for me. Tested with Minetest Game. Looks good. Will merge in a few days unless there are objections.

SmallJoker:
Squashed manually before merge. It was quite messy. Anyway. Thanks for the contributions.

PR #701, opened 2025-02-01
Fix typos in mcl compatibility

j-r:
This fixes the names of the gamecompat aliases for mese and the dye iteration. No alias for the mese crystal fragment is defined, because there is no suitable item in mcl and it isn't needed.


j-r:
I misseed the other PR. Sorry:-)

PR #702, opened 2025-02-12
Better Russian language

SkyBuilder1717:
Current Russian was too offset from original strings, but I am here to fix it!

Fixed several Russian translations files, also deleted several useless strings.

Commits is bit of a mess, but trust me!

### DIFF
--- a/mesecons_delayer/init.lua
+++ b/mesecons_delayer/init.lua
@@ -106,8 +106,7 @@ minetest.register_node("mesecons_delayer:delayer_off_"..tostring(i), off_state)
 
 -- Activated delayer definition defaults
 local on_state = {
-	--## It makes no sense to leave such an Easter egg here. It's literally useless! ##--
-	--description = "You hacker you",
+	description = "You hacker you",
 	inventory_image = "jeija_gate_on.png^jeija_delayer.png",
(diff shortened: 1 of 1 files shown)


SmallJoker:
Reviews are welcome. I skimmed through the strings and they seem OK.
@randomei Might you be interested in having a look at it? (ping due to #692)

SmallJoker:
LGTM. Will merge in a few days unless there are objections.

SmallJoker:
"mesecons" has no direct meaning in English. It contains "cons", a short form for "connections", but that's about it.
"mesecons" is a name. For example, I won't think you would translate "Amazon" into "Gigantic River" - or "Greenland" into "Northern Ice Plains". I would like to have a translation that resembles the English variant in pronunciation in this case. Hence, "Mezecons".

randomei:
@SmallJoker also this pr is removing easter egg at mesecons_delayer/init.lua 109, what do you think about this?

SmallJoker:
Apparently I did not pay attention at all while browsing through the diff of this PR.
I personally like that easter egg. It's only one string and doesn't need a translation if that's a concern. Please revert the Lua code changes for now. Let's limit this PR to translations only.

randomei:
@SkyBuilder1717 But still, the mod name and its derivatives were changed.

PR #703, opened 2025-03-02
Add Ukrainian localization

FromKaniv:


modpack.conf
```diff
@@ -1,3 +1,6 @@
 name = mesecons
 description = Mod that implements a ton of items related to digital circuitry.
 min_minetest_version = 5.0
+release = 29687
```
SmallJoker, line 4: Quoting lua_api.md:

>     * `release`: Ignore this: Should only ever be set by ContentDB, as it is
>                 an internal ID used to track versions.

Please undo these changes to `modpack.conf`.

FromKaniv: OK. If I understood correctly, I was supposed to remove the "release" line in the mod.conf file, which I just did

FromKaniv: When will my PR be reviewed?

SmallJoker:
> When will my PR be reviewed?

Never.

PR #705, opened 2025-05-02
replace `mesecon.on_rotate_horiz` with `mesecon.on_rotate` and `paramtype2 = "4dir"`

The4codeblocks:
function left for compatibility


SmallJoker (approved):
Works. Tested by creating a circle using `mesecons_extrawires:corner_off` + screwdriver. After applying the PR (and thus changing the paramtype2), the node appearance is still the same.

Will merge in a few days unless there are objections.

sfan5:
Compatibility could be a concern as Mesecons technically still supports MT 5.0 but :shrug:

SmallJoker:
> Compatibility could be a concern

That is indeed a good argument. In this case, the PR also has to require 5.7.0 or newer to prevent users on ancient builds from running into issues.

The4codeblocks:
> 1. https://github.com/minetest-mods/mesecons/blob/master/modpack.conf#L3

done, also where do you get version spec?

SmallJoker:
> done, also where do you get version spec?

What do you mean by that? To find out the minimal version? I do that by checking versions of https://github.com/minetest/minetest/blob/5.7.0/doc/lua_api.txt?plain=1 - or `lua_api.md` - by replacing the `5.7.0` part and checking whether there's any mention of `4dir` (in this case). Unfortunately the web version refuses to show a `git blame` and my shallow copy of the repository does not reach that far back.

SmallJoker:
Accidentally used Rebase & Merge. Force-pushed as one with improved commit description as 2707254.

PR #708, opened 2025-06-08
Switch to mineunit-actions

### LINKED ISSUE #706: Mineunit CI issues/breakage
### Regarding this:

https://github.com/minetest-mods/mesecons/blob/b79962402b829e5ea939572cdc06ba149e9588a9/.github/workflows/check.yml#L26-L28

As Luarcoks package is still more experimental and not very well tested thing I'd strongly recommend using mineunit docker images instead: https://hub.docker.com/r/mineunit/mineunit.
These images are tested a bit better, regularly updated and can be easily locked to revision if needed.

Additionally, for local use, docker offers relatively easy and "standard" way to run mineunit locally on your own machine while isolating both mineunit itself and mod code (which both can be a security risk without proper isolation).

Since last release I've added note about "recommended way" to readme: https://github.com/S-S-X/mineunit/blob/master/README.md#how-to-use-mineunit

### Luarocks package has not been main focus mostly because of shortcomings of luarocks itself with:

* compatibility (lack of updates, lack of versioning).
* security (no isolation whatsoever).
* interoperability (truckload of issues with different operating systems and linux distributions).
* relatively complicated package releases (language/application specific).

### Optionally for reference, if going with prebuilt docker images, there's:

* GitHub action available at https://github.com/mt-mods/mineunit-actions
* Example action for command line arguments / modpack testing / different working directory at https://github.com/mt-mods/technic/blob/379bedc20d7ab11c758afa52d5916b23dced5354/.github/workflows/mineunit.yml#L13-L18
    ```yaml
    - id: mineunit
      uses: mt-mods/mineunit-actions@master
      with:
        working-directory: ./technic
        mineunit-args: --engine-version 5.4.1
    ```
* Somewhere there was other examples like examples to load other mods as dependencies, if needed I could probably dig up something.

Ping @sfan5 as you uncovered this mineunit luarocks package issue at https://github.com/S-S-X/mineunit/issues/104

S-S-X:
Use GitHub action repo for mineunit tests.

* Should be enough to close #706 
* Tested on fork: https://github.com/S-S-X/mesecons/actions/runs/15518379561
* Provides a bit cleaner workflow definition.
* Should also improves test performance a bit by utilizing pre built containers from https://hub.docker.com/r/mineunit/mineunit

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,22 +19,27 @@ jobs:
     name: "Mineunit tests"
     steps:
     - uses: actions/checkout@main
-    - name: apt
-      run: sudo apt-get install -y luarocks lua-bitop
-    - name: install lua deps
-      run: for x in busted luacov; do luarocks install --local $x; done
-    - name: install mineunit
-      # FIXME: this seems to be abandoned?
-      run: luarocks install --server=https://luarocks.org/dev --local mineunit
     - name: run mesecons tests
-      working-directory: ./mesecons/
-      run: $HOME/.luarocks/bin/mineunit -q
+      uses: mt-mods/mineunit-actions@master
+      with:
+        coverage: false
+        mineunit-args: -q
+        working-directory: ./mesecons
     - name: run mesecons_mvps tests
-      working-directory: ./mesecons_mvps/
-      run: $HOME/.luarocks/bin/mineunit -q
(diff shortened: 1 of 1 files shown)


SmallJoker (approved):
LGTM. Will merge in a few days unless there are objections.

PR #710, opened 2025-06-13
Support for Hades Revisited.

sfence:
Integrate support for Hades Revisited into the Mesecons modpack, allowing it to tag https://content.luanti.org/packages/SFENCE/hades_mesecons/ as obsolete.

### DIFF
--- a/mesecons_gamecompat/mod.conf
+++ b/mesecons_gamecompat/mod.conf
@@ -1,3 +1,3 @@
 name = mesecons_gamecompat
 depends = mesecons
-optional_depends = fire, default, dye, mesecons_mvps, tnt, doors, beds, mcl_fire, mcl_core, mcl_dye, mcl_tnt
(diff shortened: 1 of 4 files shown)


mesecons_gamecompat/mod.conf
```diff
@@ -1,3 +1,3 @@
 name = mesecons_gamecompat
 depends = mesecons
-optional_depends = fire, default, dye, mesecons_mvps, tnt, doors, beds, mcl_fire, mcl_core, mcl_dye, mcl_tnt
+optional_depends = fire, default, dye, mesecons_mvps, tnt, doors, beds, mcl_fire, mcl_core, mcl_dye, mcl_tnt, hades_core, hades_sounds, hades_torches, hades_doors, hades_beds, hades_fire, hades_tnt
```
SmallJoker, line 3: Technically you only need to depend on `hades_sounds`. Everything else is only accessed after all mods have finished loading.
The same holds true for the `mcl_ *` mods, but touching that is out of scope of this PR.

PR #713, opened 2025-10-29
siliconless gate recipes

The4codeblocks:
surprisingly enough, the 3² grid is just big enough to fit an xor gate

"Add alternate gate recipes without silicon

Added alternate crafting recipes for AND, NAND, and XOR gates without silicon." - copilot


SmallJoker:
Why would you need this? 4x silicon can already be crafted using 3x sand + 1x steel (`mesecons_materials/init.lua`).

The4codeblocks:
> Why would you need this? 4x silicon can already be crafted using 3x sand + 1x steel (`mesecons_materials/init.lua`).

some of these recipes are too simple to warrant silicon (and, nand), and I doubt an xor gate needs as much/more silicon than an *fpga*

the only reason I kept the old recipes was to avoid breaking stuff, who knows, someone may have an autocrafter for these somewhere

The4codeblocks:
?

PR #715, opened 2025-11-09
Luacontroller: Add overheat factor for deferred tasks

SmallJoker:
Fixes #714

@Opons1 Would you please be so nice to check whether this reduces the server lag?


a-tour-ist:
Since there was the question for review on IRC, I'll add my two cents...

I could not reproduce the lag issue in a new plain world... I suspect there is a large pipeworks Nodebreaker/Deployer machine connected to channel 1? (And @SmallJoker changed the "luacontroller_maxevents" setting to time out earlier?)

So lag doesn't come from the luacontroller itself, but rather from the callbacks run by the connected digiline devices?
I don't know a good way to limit digiline messages, so burning the controller might be the best/easiest idea...

Interrupt has indirect punishment already (once the controller is reactivated), I would only punish digiline_send

PR #716, opened 2025-11-10
mesecons: Unify duplicated code

SmallJoker:
The new comments should help understanding what is going on to ease maintenance in the long-run.

This is a partial adoption of #614

**How to test:**

* Create mesecons circuit
* The conductors must update correctly
* No errors during operation (these are very basic functions, thus issues would be visible very quickly).


sfan5 (approved):
seems to work

PR #717, opened 2026-01-12
heatsinkable luac

The4codeblocks:
prereq? for mt-mods/digistuff#88


SmallJoker:
If `digistuff` depends on `mesecons_luacontroller`, then `mesecons_luacontroller` should provide some *documented* anchor such that mods can use `core.override_item` on all relevant nodes to extend its features.
It is surely possible to add this group as you proposed, but wouldn't it be easier if mods could easily do whatever they want without having to implement features directly in `mesecons`?

PR #724, opened 2026-06-01
Run `optipng` on all `.png` files

Sheriff-Unit-3:
4 files were optimized with the `optipng` command.


SmallJoker:
Fine by me. Will merge in a few days unless there are objections.

PR #728, opened 2026-08-25
Add 'disable_door_override' setting

tenplus1:
This adds a 'mesecon.disable_door_override' setting to stop mesecon_gamecompat from overriding all registered doors which can break some 3rd party mods.

Note: This use to be a 'mesecon_doors' mod that could be easily disabled.

### DIFF
--- a/mesecons_gamecompat/init.lua
+++ b/mesecons_gamecompat/init.lua
@@ -24,6 +24,7 @@ if minetest.get_modpath("hades_core") then
 	dofile(minetest.get_modpath("mesecons_gamecompat").."/compat_hades.lua")
 end
 
-if minetest.get_modpath("doors") then
+if minetest.get_modpath("doors") and
+		not core.settings:get_bool("mesecon.disable_door_override") then
 	dofile(minetest.get_modpath("mesecons_gamecompat").."/doors.lua")
 end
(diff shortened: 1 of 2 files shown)


sfan5:
Uh, maybe we should fix the door override instead? What's the problem?

tenplus1:
@sfan5 - Doors Redo had it's own mesecon support baked in when 'mesecons_doors' could be simply switched off and all was well, but now that mesecon_gamecompat automatically registers all doors and overrides the already added mesecons table in the node definition it breaks compatibility.

SmallJoker:
> the already added mesecons table in the node definition it breaks compatibility.

How so? Are these tables different from what `mesecons_gamecompat` adds? -->
https://github.com/minetest-mods/mesecons/blob/3a2d6916ee090353b69bd019f9954971ddc8f4a9/mesecons_gamecompat/doors.lua#L59-L75

sfan5:
it should not be overriding doors that already have custom mesecons definition:
https://github.com/minetest-mods/mesecons/blob/3a2d6916ee090353b69bd019f9954971ddc8f4a9/mesecons_gamecompat/doors.lua#L55-L58

tenplus1:
@sfan5 - True, although having a simple switch to disable the door overiddes would have the backwards compatibility we had before.

PR #730, opened 2026-09-01
fix on_blast for pistons

a-tour-ist:
Currently it is possible to create unconnected piston pushers / the corresponding piston_*_on by creating an explosion which only blasts half of the double-node away...

### DIFF
--- a/mesecons_pistons/init.lua
+++ b/mesecons_pistons/init.lua
@@ -229,6 +229,11 @@ local function piston_punch(pos, _, player)
 	end
 end
 
+local function piston_blast_on(pos, intensity)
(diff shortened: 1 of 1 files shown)


mesecons_pistons/init.lua
```diff
@@ -229,6 +229,11 @@ local function piston_punch(pos, _, player)
 	end
 end
 
+local function piston_blast_on(pos, intensity)
+	local node = core.get_node(pos)
+	piston_remove_pusher(pos, node, true)
```
SmallJoker, line 234: This plays the sound effect `piston_retract`, which seems *wrong*, though low priority.

PR #732, opened 2026-09-14
rotatable movestones

The4codeblocks:
gives movestones `facedir` rotation and top and bottom rules
the facedir rotation modifies behavior such that the side graphics always match the travel direction

as a side feature rules merging doesn't result in replicates


The4codeblocks:
so..... they didn't have vector.rotate back then

the checks for newer versions shouldn't be cancelled by older versions failing, because the failures could be caused by relying on recent features

